PR #586, opened 2015-01-31
issue#570 add upload ssl certificate

kamal-gade:
Enable user to upload own ssl certificate.
This will replace the existing ssl certificate.


kamal-gade:
This is not a complete and tested version. Backend calls have to be created.
Just have a look at upload logic and let me know if that works.

kamal-gade:
Added backend python code to support front end

PR #626, opened 2015-03-18
Upgrading Django Rest Framework from 2.x to 3.1.

gkadillak:
The big change has been the uprooting of the login in the rfc.GenericView. The
only logic needed in the class is the _handle_exception method since everything
else is incorporated into DRF itself.

Work left to do is the flat view of the json response for the shares.


schakrava:
Thanks Garret. For the Share serializer, how about doing the following?

1. Add serializer for snapshots field
class ShareSerializer(serializers.ModelSerializer):
    snapshots = SnapshotSerializer()

2. Add snapshots to fields
class Meta:
         model = Share
         fields = ('pool', 'qgroup', 'name', 'uuid', 'size', 'owner', 'group', 'perms', 'toc', 'subvol_name',
                      'replica', 'compression_algo', 'cur_rusage', 'cur_eusage', 'snapshots')

3. Add snapshots as a property to the Share model class
@property
def snapshots(self, *args, **kwargs):
    Snapshot.objects.filter(share=self)

PR #662, opened 2015-05-30
Issue # 619 Modify task frequency input field

kamal-gade:
Add more options for providing task frequency input


schakrava:
Thanks for starting this off @kamal-gade . @gkadillak used jquery-cron per your pull request and finished it in #728

PR #672, opened 2015-06-08
Issue#670 bootstrap upgrade

gkadillak:


gkadillak:
Please pull this branch down and run the UI in order to find as many bugs as possible before this pull request is merged.

phillxnet:
@gkadillak This looks like a great move, however, it would help at least me if you could specify the exact procedure / set of commands that are required to pull your branch onto presumably an existing Rockstor system and have it build so that the resulting tests / observations are valid.  Sorry, I know there is the "Contirbute to Rockstor" doc but I am unsure if this relates exactly to your request and it doesn't specify the build directory.  Also last time I tried the referenced procedure I ended up with nothing of Rockstor running at all.  Probably me fumbling something.  Also I see that Bootstrap have recently moved to v3.3.5, might that be the one to go for if we are all change anyway.

I am more than willing to have a go if you could specify in commands a sane procedure as I'd like to help out.  Thanks.

schakrava:
@phillxnet Perhaps we can take this opportunity to improve the contribution document. 

You should be able to pull this particular branch from garret with these commands
git checkout -b gkadillak-issue#670_bootstrap-upgrade master
git pull https://github.com/gkadillak/rockstor-core.git issue#670_bootstrap-upgrade

Then follow the steps in "Testing changes" section.

phillxnet:
@schakrava Thanks; yes I was hoping once I've got more familiar / successful with the process to do a pull request on the contribution doc to address what wasn't clear to me initially.
Two questions:-
1:- Should the "deploy_dir" on our VM in "Testing changes" be /opt/rockstor or not?
That might have been what I did wrong last time,
2:- Do we still only need to start the three services noted or should we also include ztask-daemon?
Cheers

phillxnet:
@gkadillak I think I've found a format anomaly, in Storage Pools table Usage has percentage free in brackets but this seems to be formatted with a <big> tag and I suspect <b> was intended.  Can't seem to find where this is defined though.

gkadillak:
Just made that change right now, @phillxnet. Testing right now.

Issue #673, opened 2015-06-09
KVM VirtIO drive serial number not found


phillxnet:
Originally though to be related to issue #666 but that one turned out to be a esxi client bug.

If one creates drives in kvm that use the VirtIO disk bus the serial number is not visible to Rockstor.

This I believe is due to Rockstor using the following command in btrfs.py to gain drive info:-
'''
/usr/bin/lsblk -P -o NAME,MODEL,SERIAL,SIZE,TRAN,VENDOR,HCTL,TYPE,FSTYPE,LABEL,UUID
'''
But this returns the following for the vda block device:-

NAME="vda" MODEL="" SERIAL="" SIZE="2G" TRAN="" VENDOR="0x1af4" HCTL="" TYPE="disk" FSTYPE="" LABEL="" UUID=""

So no serial number returned by lsblk but /etc/libvirt/qemu/rockstor3.7.xml contains the VMM configured serial number:-
```
<disk type='file' device='disk'>
      <driver name='qemu' type='qcow2'/>
      <source file='/var/lib/libvirt/images/original-test-volume.qcow2'/>
      <target dev='vda' bus='virtio'/>
      <serial>11111111111111111111</serial>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x09' function='0x0'/>
    </disk>
```
But:-
```
udevadm info /sys/class/block/vda
```
retrieves the serial number successfully:-
```
P: /devices/pci0000:00/0000:00:09.0/virtio3/block/vda
N: vda
S: disk/by-id/virtio-11111111111111111111
E: DEVLINKS=/dev/disk/by-id/virtio-11111111111111111111
E: DEVNAME=/dev/vda
E: DEVPATH=/devices/pci0000:00/0000:00:09.0/virtio3/block/vda
E: DEVTYPE=disk
E: ID_SERIAL=11111111111111111111
E: MAJOR=253
E: MINOR=0
E: SUBSYSTEM=block
E: TAGS=:systemd:
E: USEC_INITIALIZED=16682
```
I am working on a patch for this issue but it may take me a while.

phillxnet:
Pull request submitted and ready for review; hopefully relatively uninvasive.

phillxnet:
Closing as patch now merged and released in version 3.8-1

PR #675, opened 2015-06-10
Issue#673 kvm virtio drive serial

phillxnet:
Patch to provide an additional mechanism to retrieve drive serial numbers.  Currently only seems necessary for virtio drives as lsblk fails to retrieve their serial which we need.  So when existing mechanisms fail to get a drive serial this patch tries again with for example a cat /sys/block/vda/serial.
This entry seems not to exist with other drive types and we have a fail through to return an empty string if no serial entry is found so (hopefully) no worse off anyway if accidentally called with another device type.
Should only be called when existing mechanism fails to get a serial for a block device listed in lsblk.

N.B. does not address the issue of being able to install on a vda (rather than an sda) device.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -32,7 +32,6 @@ from system.exceptions import (CommandException, NonBTRFSRootException)
 from pool_scrub import PoolScrub
 from pool_balance import PoolBalance
 
-
 MKFS_BTRFS = '/sbin/mkfs.btrfs'
 BTRFS = '/sbin/btrfs'
 MOUNT = '/bin/mount'
@@ -530,7 +529,7 @@ def balance_status(pool, pool_device):
         if (re.match('Balance', out[0]) is not None):
             stats['status'] = 'running'
             if ((len(out) > 1 and
-                         re.search('chunks balanced', out[1]) is not None)):
+                    re.search('chunks balanced', out[1]) is not None)):
                 percent_left = out[1].split()[-2][:-1]
                 try:
                     percent_left = int(percent_left)
@@ -607,7 +606,7 @@ def scan_disks(min_size):
                 val_iter = True
                 i = i + 2
             elif (val_iter and sl[i] == '"' and
-                      (i == (len(sl) - 1) or sl[i + 1] == ' ')):
+                    (i == (len(sl) - 1) or sl[i + 1] == ' ')):
                 val_iter = False
                 name_iter = True
                 i = i + 2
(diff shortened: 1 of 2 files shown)


phillxnet:
Fixes issue #673

schakrava:
@phillxnet Big thank you for opening this pull request and sorry it took a bit longer than I wanted to provide feedback. A couple of comments

1. Indentation seems a bit off, perhaps your tab is mapped to 8 spaces? You can see this on lines 48-49 in btrfs.py and a few other places. Can you change it to 4 spaces and reformat?

2. in get_virtio_disk_serial function, is there any particular reason for using the cat command instead of programmatically checking the existence of the file and reading it like so:

import os
dev_path = ('/sys/block/%s/serial' % device_name)
if (os.path.isfile(dev_path)):
    with open(dev_path) as dfo:
        for line in dfo.readlines():
            # parse the contents and extract serial number

3. Can you attach a sample output of cat /sys/block/vda/serial ?

phillxnet:
@schakrava Thanks for the review.

Ok, yes sorry about the indentation.
1:
I am set to 4 spaces here +4 more for continuation lines. Seems there was an interplay between existing indents and (in one case re-occurring) additional auto indents.
The additional patches should address this, your noted example was a result of the auto-indenter grouping like with like.  I have re-formatted slightly there and that seems to have sorted things.
Should at least be more consistent now than it was before.
2:
re: reasoning of cat
I was originally going to use "udevadm info /sys/class/block/vda"
but then found that eg cat /sys/block/vda/serial gave us directly what we were after so seemed like the most direct path with the added bonus of an error code / dealing with the non existence.
ie a sample output of the cat command in this context via a ssh:-
answers your 2.1
```
[root@rockstor ~]# cat /sys/block/vdc/serial 
12345678901234567890[root@rockstor ~]#
```
So as can be seen its offered up on a plate so I went with it.
2: (some more reasoning)
Seemed like the simplest solution as then no file handler and iterator and loop were needed to get at something that we could just get directly with a single native (and hence quick) command.  I know there may be the overhead of firing up a whole shell to get this command but there is going to be similar 'weight' behind the scenes in the cross platform file system access; only this way we can more easily see what happens as less code.

If you would prefer that we avoid this cat method then I can re-write with your suggestion but it seems heavier to me.  I was also aware that in the shell the cat command has always returned special files just fine but other file opening programs have on occasions failed with special files; leading me superstitiously I admit, to the simple and I though clean solution. Especially given the pre-existence of your run_command. Sorry no examples to hand of special file access failure.

Would you like me to re-write it?

schakrava:
@phillxnet your reasoning for cat sounds fair enough, thanks for explaining it. My tendency is to use python if I can, but that's just me. no need to re-write your code. one other thing popped out with regards to return str(out)[2:-2] . since out is a list where the first element is our serial number, you can just return out[0] . It's a trivial change and because I am closing out for the release update, I made that change myself and merged. See you in the next pull request and thanks again!

PR #691, opened 2015-06-25
650 raidlevel changes

schakrava:


phillxnet:
@schakrava haven't tested this pull request re issue but just referenced in case it's related.

PR #724, opened 2015-07-09
Issue#706 websockets

gkadillak:


src/rockstor/storageadmin/templates/storageadmin/base.html
```diff
@@ -100,6 +100,7 @@
     <script src="/static/js/lib/moment.min.js"></script>
     <script src="/static/js/lib/simple-slider.min.js"></script>
     <script src="/static/js/lib/chosen.jquery.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/socket.io/0.9.6/socket.io.min.js"></script>
```
phillxnet: @gkadillak Sorry if I've got the wrong end of the stick but shouldn't we copy this library "socket.io.min.js" to our local stash (/static/js/lib/) as what if someone is on a network with limited / or no internet access?

phillxnet:
@gkadillak from what I can tell you are setting up some kind of system service and using nginx as a proxy but if we use 127.0.0.1:8080 do we not run the risk of clashing with our default ownCloud Rock-on WebUI port which I'm assuming is also listening on localhost as well as the ethernet port.

ie on 3.8-2 release with ownCloud Rock-on on:-
```
nmap -sT localhost | grep 8080
8080/tcp  open  http-proxy
```
and with ownCloud off it's not there so it looks like the ownCloud Rock-on is setting up it's own proxy on this port and localhost already.

Sorry if I have this all wrong.

schakrava:
That's a very good point @phillxnet. I am voting for 8001, since gunicorn runs next to it on 8000.

gkadillak:
8001 works with me. I'll make that change right now.

src/rockstor/storageadmin/templates/storageadmin/base.html
gkadillak: This is merely me being lazy, @phillxnet. Once the code gets pushed to production, the hard copy of the file will be referenced.

schakrava:
Fixes #724

PR #759, opened 2015-07-25
Kernel Error Not Showing Up fix

### LINKED ISSUE #746: Kernel error not showing up
An global alert should be shown telling the user to reboot in order to run the correct kernel.

gkadillak:
Fixes #746


schakrava:
Thanks @gkadillak

PR #786, opened 2015-08-12
Issue#757 mmcblk no serial

phillxnet:
This patch set is to address the no serial returned by lsblk for virtio and mmcblk* (some sdcard devices) block devices.  It deprecates get_virtio_disk_serial with a more robust and generic method that uses udevadm.

Currently it is only called when no serial is returned by the existing lsblk mechanism however I have tested this mechanism using the provided hard wired over ride flag so that it is used by default on all devices. This flag is false by default.  The results are promising and in all cases (multiple mmcblk devices and multiple with and without serial virtio devices and multiple hdd's including a USB bus driven one and the mSATA adapter device sold in Rockstor shop with a KINGSTON SMS200S330G installed and it currently serves as a drop in replacement (returning the exact same serial numbers) for lsblk where lsblk did return serials numbers that is.

I have provided the flag mechanism so that we might trial this mechanism on for example the 16 disk system with Areca ARC1680IX ver 2.3 raid controller that in JOBD (preferred) config lsblk returned non unique serials (they were all the same) where as udevadm returned "disk label looking" serials; as lsblk usually does. The mechanism to prioritize ID_SCSI_SERIAL has not been tested as I don't have such a controller here but the code looks OK for this.

see forum post:- http://forum.rockstor.com/t/s-m-a-r-t-statistics-not-working-with-areca-raid-controller/266/3

Currently non invasive but promising as a replacement for lsblk for getting device serial numbers and can initially serve as a test for this cascade rational.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -781,9 +781,9 @@ def scan_disks(min_size):
             if (dmap['SIZE'] < min_size):
                 continue
             if (dmap['SERIAL'] == ''):
-                # lsblk fails to retrieve SERIAL from KVM VirtIO drives
-                # so try specialized function.
-                #dmap['SERIAL'] = get_virtio_disk_serial(dmap['NAME'])
+                # lsblk fails to retrieve SERIAL from VirtIO drives and some
+                # sdcard devices so try specialized function.
+                # dmap['SERIAL'] = get_virtio_disk_serial(dmap['NAME'])
                 dmap['SERIAL'] = get_disk_serial(dmap['NAME'], None)
             if (dmap['SERIAL'] == '' or (dmap['SERIAL'] in serials)):
                 # No serial number still or its a repeat.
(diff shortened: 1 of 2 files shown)


schakrava:
@phillxnet Thanks for adding a lot of detail. I am inclined to (1) remove lsblk entirely and (2) just use ID_SERIAL from udevadm. In my testing so far, it seems like if ID_SERIAL is empty, then there is hardly any scenario where lsblk gives us something useful. I'll test more, but do this is what I think so far.

phillxnet:
My concern was to provide a drop in replacement and I like that the cascade provides the "drive sticker" serial number. If we whole sale move to ID_SERIAL we will break pretty much all installs as that serial is almost always not what lsblk returned.  But the code would be simpler and neater and maybe I am wrong with the breakage.

schakrava:
Thanks @phillxnet This is a certain improvement. and +1 for helpful comments. 

I have plans to use udev/pyudev for grabbing udev notifications, at which time, i'll revisit this logic and see if we can eliminate lsblk entirely.

phillxnet:
@schakrava Cheers. Yes it does look like udev is the way to go and the udev rules workaround for no serial is a nice find; sorry missed that comment edit some how but did catch your forum post on the same:-
http://forum.rockstor.com/t/proxmox-serial-number-passthrough/319/2
Linking here for context.

PR #790, opened 2015-08-14
Issue#741 smartd.conf update

phillxnet:
This corrects the smartd.conf update mechanism and repairs the mouseover tooltip on the associated custom smartd config form. The form now correctly displays previous custom entries as well.
I intended to based my branch on 3.8-4 to help at my end with dev but looks like enough has changed that it can't auto merge.  Let me know if you need me to re-submit, the changes are actually only small; I just made a meal of it.
N.B. I am unsure if I've gone about this correctly but the result is things working as intended and I have left additional logging in place in case I've missed / caused a scope issue.


schakrava:
Thank you @phillxnet, this is just the kind of fix we need the most at this time.

phillxnet:
Cheers.

PR #794, opened 2015-08-16
create postfix/sasl_passwd with email not just user

### LINKED ISSUE #793: sender not username in sasl_password
I believe it is far more common and required to use the full email address in sasl than just the username (ie the before "@" bit)
gmail I think will work with either though as only one domain.
ie we want something like:-
[server_name]:587 user@domain.com:password
but what we are getting is:-
[server_name]:587 user:password
which will not work with many smtp servers as they deal with many domains unlike gmail.
The full email is required with for example tsohosts's email server and fastmail's email server.

I am preparing a pull request on this now.

phillxnet:
Simple patch to use the full email (sender) rather than the sender.split('@')[0] in sasl_passwd file creation.
This will work with gmail as well as many other providers who's smtp servers deal with more than one domain.
fixes #793

### DIFF
--- a/src/rockstor/storageadmin/views/email.py
+++ b/src/rockstor/storageadmin/views/email.py
@@ -56,11 +56,11 @@ def update_forward(email, revert=False):
         if (not revert):
             fo.write('%s, root\n' % email)
 
-def update_sasl(smtp_server, username, password, revert=False):
+def update_sasl(smtp_server, sender, password, revert=False):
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet. I've tested with gmail account. Looks good to me.

phillxnet:
Great, worked a treat with fastmail also.

PR #801, opened 2015-08-18
validate pool name length and correct js code indentation.

Mchakravartula:


phillxnet:
@Mchakravartula Hello, sorry if this is inappropriate but I think that the minimum number of drives for btrfs raid levels might be (listing the ones I think are incorrect in these files):-

RAID5 is 3 drives, not 2
RAID6 is 4 drives, not 3.

I see that you didn't change these but though it might be an opportunity to double check them.
I am not certain of these corrections but pretty sure. Sorry if I have this wrong, maybe my numbers are what makes sense not the absolute minimum.

Mchakravartula:
@phillxnet Thank you for your inputs. I will double check before I write functional tests for RAID5/6..

schakrava:
@phillxnet minimum is 2 devices for RAID5 in BTRFS. It's a deviation from conventional RAID5 concepts, but I got confirmation from the devs too. In case of a 2 device RAID5 pool, data is on one and parity data on the other. So there is really no data striping in case of 2 devices. But technically, it's possible to have a 2 device RAID5 pool. Same logic applies to RAID6 with a minimum of 3 devices.

phillxnet:
@schakrava OK great thanks, sorry for the noise; just thought I better chime in just in case.  I saw a recent post on linux-btrfs that stated min 4 for RAID6 so tried looking it up and came up blank except for:- https://lwn.net/Articles/577961/ Non authoritative and old I know but all I could find quickly late last night.

linux-btrfs post:-
Austin S Hemmelgarn to Duncan
"...  It may also be worth pointing out that BTRFS raid6 lets you use 4 disks minimum, as opposed to most other raid6 implementations that (unnecessarily, as a 4 disk RAID6 is not a
degenerate form) require 5."
And I haven't yet seen anyone counter the claim so I went with it here.
http://news.gmane.org/gmane.comp.file-systems.btrfs
17 August 13.38

schakrava:
no worries @phillxnet, we are not alone in this confusion.

Issue #832, opened 2015-08-26
Improve top menu bar & breadcrumbs menu


snamstorm:
- Improve styling of top menu bar
- Improve styling of breadcrumbs
- Fix hamburger menu
- Fix responsiveness size

PR #839, opened 2015-08-29
Improve top menu bar & breadcrumbs menu #832

snamstorm:


schakrava:
great work @snamstorm

PR #844, opened 2015-08-30
Improve top menu bar

snamstorm:
- added functionality to collapse hamburger menu after click


schakrava:
works great @snamstorm and thanks for testing @phillxnet

Issue #851, opened 2015-09-02
some built and run time issues with paths


phillxnet:
I think I've found some path issues that only affect dev work but wanted to pop them in anyway as think it might help in the future.
Relates to buildout and runtime path issues I found while trying to find my feet.
Sorry for not following correct procedure but got caught up in this as a side line.
Have pull request ready to address what I understand so far of these issues.
Please take with great caution as I'm rather new to buildout.

phillxnet:
Closing as now merged.

PR #852, opened 2015-09-02
issue#851 Build time rockstor.service path issue and runtime initrock.py path issue

phillxnet:
I found some build time anomalies with regard to paths but am not sure of the repercussion especially with regard to the creation of production packages.

Please pay particular attention to the now buildout in-line edited rockstor.servie. I have maintained the existing contents in case it is required by a script elsewhere in rockstor-prod.service.in but I haven't created any buildout directive for processing this file.


phillxnet:
fixes #851

schakrava:
@phillxnet Thanks for this pull request. Everything looks fine, but I have one concern. During the development phase, we are not setup to use systemd as you know. Is your goal with this pull request to use systemd even during development so we don't have to execute supervisor commands separately?

phillxnet:
@schakrava my goal was simply to remove the hard wired paths and have the rockstor.service file generated appropriately for the directory. I don't think I get why we are not using systemd to start in dev when that is how it is run in production? I think I am missing something here. Do these changes then interfere with aspects of dev I haven't appreciated. I thought that the build process should produce a fully working system with systemd integration and all and be dir independent. Are we not inches away from this anyway. As it was the init script would fail unless there was an existing independent /opt/rocksor located install, this just didn't seem right. This turned up when I tried to build from scratch without anything but the source in a different dir than /opt/rockstor which is now deleted anyway as part of the build. Sorry I think there is an element that I haven't grasped yet. Why are we not using systemd in dev? I had just assumed that we did I'm afraid. Looks like that was part of my misunderstanding. Should the code not be able to build what is shipped and apart from these (tiny) changes what else is missing for that to happen. Excuse my ignorance if I have this all about face.

phillxnet:
So short of it is that yes I think we should run in dev in exactly the way it is run in production, especially as we are getting more complicated / involved with other services ie the docker starting thing etc. Sorry beginner here so may be unhelpful contribution / perspective.

schakrava:
I think I was just used to the same build process since day 1 and systemd came along later and integrated for the prod but I never really thought of making dev the same. Now that you brought it up, it seems obvious that we should make the process the same. And yes, we are just inches away anyway. I'll happily test this pr, make any minor changes I could think of and merge it.

Issue #863, opened 2015-09-06
Improve tables across system


snamstorm:

FroggyFlox:
Unfortunately, this seems to have run out of developer time or interest so I will be closing it as a tidy up. If someone is still interested in picking this up, feel free to open a new issue.

PR #866, opened 2015-09-06
Improve tables across system #863

snamstorm:


schakrava:
It's a bummer we couldn't get this fully done @snamstorm . The code has deviated a bit too much for this pull request to make sense. A new pr would be better.

PR #892, opened 2015-09-18
Issue#595 add support for ups

phillxnet:
Initial pull request to ease collaboration.
Not ready to merge yet as core function not yet complete.

Outstanding issue currently is nut_service.py need to only start nut-monitor when mode in config is netclient, otherwise all is good as is.
Stuck on how to act appropriately on config in this area of the code. Thanks.

### DIFF
--- a/base-buildout.cfg
+++ b/base-buildout.cfg
@@ -61,7 +61,17 @@ command =
 	postgresql-server postgresql-devel kernel-ml btrfs-progs rsync \
 	nfs-utils kernel-ml avahi netatalk smartmontools net-tools sos hdparm \
 	postfix cyrus-sasl-plain yum-cron nano usbutils pciutils \
-	nut nut-xml epel-release
+	epel-release
+
+[rpm-deps-nut]
+recipe = plone.recipe.command
+stop-on-error = true
+install_cmd = sudo /usr/bin/yum install --setopt=timeout=600 -y
+# Install Network UPS Tools from epel: note nut brings with it nut-client
+# N.B. We are not in [rpm-deps] as we first need the install of epel-release
+# which takes place in [rpm-deps] section.
+command =
+        ${rpm-deps-nut:install_cmd} nut nut-xml
 
 [python]
 recipe = zc.recipe.cmmi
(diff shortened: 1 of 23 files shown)


phillxnet:
Almost ready now; any pointers on the outstanding issue outlined above?
Once sorted I can test netclient mode. Currently standalone and netserver modes have been tested and email alerts on ups events are working. Auto system shutdown in these modes appears to work when system is On Battery and Battery Low. Have made provision for auto shutdown on timer when On Battery for a period of time as this is a popular config. Have now fully enabled built in ups event scheduler and provided working templates for all known events.
N.B. Currently only tested on a serial attached ups (still common would you believe).
Will soon test usb connected ups and when above issue resolved I can test in netclient mode where the ups is attached to another nut server on the network.

phillxnet:
Have now tested using a USB to Serial adapter with the mge-utalk driver.
Have also tested my provision for early-shutdown timer config made in upssched.conf.

phillxnet:
Now works in netclient mode. More testing required and a little more tidy up. Plus testing on direct usb connection as serial-serial and usb-serial now tested OK. Should be able to trial this soon.
N.B. suspect not multi network card happy, will look to this.

Also interface doesn't initially configure itself, ie configuration screen should hide some of the options but this is only effective once the "Nut Mode" has been selected, even if that selection is what was already selected. Need this to happen on the first showing of the config screen otherwise it's misleading and confusing.

phillxnet:
todo testing on some more ups units and remove logging.
Maybe address potential empty config problem in corner cases.

Tested OK on power up and works in all three modes with usb-serial adapter where appropriate and nutserver on lan otherwise. On slow real hw with mSATA root.

phillxnet:
All issues in code review have been address.
Outstanding other issue with this pr is whenever the config page is viewed it fails to re-size / hide appropriate elements until the mode is manually selected, ie even if the manual selection is no different than what was already displayed.

It is intended for the form to have already adapted to its mode config either from prior config or the widget default of standalone prior to being displayed to the user.
@priyaganti Can you point me in the right direction on this one. I will have another look at this tomorrow anyway so no worries I'm just a bit rubbish with all this bootstrap / js stuff.

Testing ups wise is now possible without a real UPS by selecting the dummy-ups driver and in UPS port field enter the following path/filename <build-root>/conf/dummy-ups.dev
ie:-
nut_dummy_ups_config

@schakrava N.B. this builds dandy from scratch ie no epel-release or any nut packages installed before hand.
But I now always get the "system does not have an ip address" message even though it does. No amount of CR's help it seems.
This happens even when nut is disabled. However I do have postfix configured.
I don't think it's my fault however :)
Do you want me to re-base on master to pick up the recent network changes?

Issue #734, opened 2015-07-13
bogus noip message in console


phillxnet:
In a recent forum post involving a machine with twin Ethernet adapters the reporter received this message:-
"The system does not have an ip address.
Rockstor cannot be configured using the web-ui without an ip address."
http://forum.rockstor.com/t/install-on-hp-dl360-g5-fails-to-detect-hardware-raid-solved/197/16

This caused "an opting out" from an important update that was though to cause it; however:-

I have since duplicated this message but it is very intermittent on my hardware and when it did occur the machine did in fact have a valid and working ip that was usable to access the web-ui so message was at least in my case bogus.

initrock.py writes either the above quoted message or a success with ip message to /etc/issue so as to be shown on the terminal, ready to inform the user.

I suspect this is a race condition and am attempting to find a repeatable hardware condition so that I can test a potential fix. @schakrava do you have any pointers for me on this as I am looking into it now.

phillxnet:
If users get this message a Ctrl+Alt+F2 can reveal the ip address (multiple times actually) required to connect, assuming we are in the bogus noip message scenario.  I have just proved this on real hardware here so until a fix is found this is a real life workaround where a keyboard is available .

phillxnet:
In another instance of bogus no ip message Carriage Return on console prior to login shows ip, but also contains the bogus no ip massage.

phillxnet:
I am currently working on a pull request for this issue.

phillxnet:
Due to the low priority I have not spend much time and made little progress as of yet on this issue. It doesn't seem to affect many installations / systems but there is definitely improvements to be made here. ie the repeating messages that build up /etc/issue. I will try and have another look at this soon.

phillxnet:
https://github.com/rockstor/rockstor-core/commit/87c9b1105621ddca205dfc4d47854a56adaa53d8
has now resolved the build up of old messages in /etc/issue

phillxnet:
Closing as #940 should address this.

PR #940, opened 2015-10-11
924 no ip address in console partially fixes #924

### LINKED ISSUE #924: no ip address given in console
As of 3.8-8 the console displays "The system does not have an ip address"
Yet it does, login and ifconfig displays this.
Reproduced by fresh 3.8-7 install, applied 3.8-8 update and then rebooted.
All subsequent reboots display this same misleading message.

Rockstor machine can be found in list of all machines via "fing" app on android or "arp -a" in terminal on linux system or via router webUI dhcp / connected clients section.

phillxnet:
Recent changes have caused a regression where rockstor.py only tires once to discover the IP address, which almost always resulted in a fail as the interfaced hasn't yet got it's IP. This results in the no ip message. Patch set returns multiple tries in background function and additionally suggests on fail issue message that pressing return may yield an updated network status. Bit hacky but didn't want to put any pre-delays before /etc/issue update in until this was merged; though that may be the way to go.
Works when applied to 3.8-8.05 testing on two real machines, both using pcie mSATA as root. One single core (but multi threaded) and one dual core.
Also note that a logging related variable has been re-instated.

### DIFF
--- a/src/rockstor/scripts/initrock.py
+++ b/src/rockstor/scripts/initrock.py
@@ -27,6 +27,7 @@ import time
 from tempfile import mkstemp
 from django.conf import settings
 
+logger = logging.getLogger(__name__)
 
 SYSCTL = '/usr/bin/systemctl'
 BASE_DIR = settings.ROOT_DIR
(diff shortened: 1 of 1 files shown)


phillxnet:
fixes #734 
As well, or what remains of it as the message with pr applied will indicate an option, ie press enter, so no longer bogus.

schakrava:
Thanks @phillxnet . Your changes, combined with the fact that initrock doesn't run as part of rc.local anymore, should fix this problem pretty good.

phillxnet:
@schakrava Cheers.

PR #941, opened 2015-10-11
update to ssl in menu items #trivial. Fixes #938

### LINKED ISSUE #938: "Update certificate" is a confusing menu item
Does it mean "update the certificate", or "something to do with an update certificate, is my stable key a certificate for an update, is it for an ssl certificate?...

phillxnet:
To avoid confusion with the newly added update channels, overtly state certificate user facing text as pertaining to SSL.
Should remove concerns over invalidating a system for it's update channel.
"Update" next to certificate is not longer displayed in top level menus re ssl but button has "Update SSL Certificate" on so we should be good. Also shorter text in nav elements so neater.

Formatting tested in Chrome and Firefox.


schakrava:
Thanks @phillxnet !

phillxnet:
@schakrava Cheers.

PR #944, opened 2015-10-11
remove 1m reboot and shutdown delay. Fixes #943

### LINKED ISSUE #943: Remove 1m reboot and shutdown delay
As it can be quite frustrating to have an imposed delay prior to reboot or shutdown I am preparing a pr to address this.

phillxnet:
Removes the default of 1m delay before reboot or shutdown and uses "now" instead, also addresses associated aesthetics. 

I have also lowered the js timer / dialog monitor feature that confirms the requested action.
I found there was a problem in reducing shutdown confirmation below 180 seconds so have left at this.
A closer look would be needed to reduce it further.


roweryan:
+1

PR #962, opened 2015-10-16
961 no ip in console regression

### LINKED ISSUE #961: no ip in console regression
Fix minor regression introduced in #940 caused it to not actually fix #924 
Correct to actually fix #924.
I am currently working on a pull request for this.

phillxnet:
Corrected regression and added comments on intended flow.
Fixes #961 
Tested on fresh build from master on real hardware.


phillxnet:
@schakrava Thanks.

PR #972, opened 2015-10-20
postfix smtp sender address. Fixes #970

### LINKED ISSUE #970: Postfix SMTP sender address
**Problem**
When mail is sent by Rockstor (or manually using `echo "test" | mail root`), my ISP complains:

> invalid DNS MX or A/AAAA resource record (in reply to MAIL FROM command)

...because Rockstor's Postfix sends the following header with a local domain:

> MAIL FROM:<root@Rockstor.localdomain>

**Solution**
This can be prevented by (see http://www.postfix.org/ADDRESS_REWRITING_README.html#generic):

1. adding this to /etc/postfix/generic:
> root mysender@myisp.com

2. running `postmap generic`
3. adding this to /etc/postfix/main.cf:
> smtp_generic_maps = hash:/etc/postfix/generic

4. running `postfix reload`

I guess it's not too hard to include these steps in the mail configuration script.

phillxnet:
Rewrite / remap local email to be from the sender email address so as to avoid some isp's blocking the notification  / root emails due to them not having a "real" domain.

previously for nut notifications we had:-
From nut@rockstor.localdomain To root@localhost.localdomain
now:-
From given-sender-email To root@localhost.localdomain

and for notifications re SMART error logs we have
previously
From notifications@hostname.localdomain To root@hostname.localdomain
now:-
From   given-sender-email  To given-sender-email

So not perfect but at least we now have legitimate "From" labelling so should no longer be susceptible to blocking on account of non public domains.

### DIFF
--- a/src/rockstor/storageadmin/views/email_client.py
+++ b/src/rockstor/storageadmin/views/email_client.py
@@ -75,6 +75,7 @@ def update_generic(sender, revert=False):
         if (not revert):
             fo.write('@%s %s\n' % (hostname, sender))
             fo.write('@%s.localdomain %s\n' %(hostname, sender))
+            # todo need an entry here to add @<hostname>.<domain>
     os.chmod(generic_file, 0400)
     run_command([POSTMAP, generic_file])
     os.chmod('%s.db' % generic_file, 0600)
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet

PR #973, opened 2015-10-21
Sort shares by name in UI

maxhq:
The share list in *Storage - Shares* or in the dropdown on *Storage - Samba* is now sorted by name per default - IMHO this gives a nicer overview. This is done by making the API (class `ShareListView`) accept any object attribute in URL parameter `sortby`.

To discuss:
If sorting is not considered a relevant overhead even for a large number of shares, we could set a default server-side sorting by attribute `name`. Currently, if `sortby` URL parameter is NOT given, no sorting takes place.


maxhq:
By the way, the idea of the `sortby` parameter was that it could be used later for other reporting, e.g. on a page that shows the biggest shares etc.

schakrava:
@maxhq My apologies for taking a long time to merge this. For some reason, I mistakenly assumed you had closed this pr.

Thanks for your contribution!

PR #975, opened 2015-10-21
Fix submit button in Plex storage configuration

maxhq:
Old (?) code tried to process none existing `event` object to disable submit button. This is now done via `getElementById()`.


schakrava:
Nice find @maxhq. Thanks for the pull request.

PR #1003, opened 2015-11-06
A bit of CSS cleanup

grogi:
The CSS was slightly cleaned up and the menu items got aligned. The hover now always takes full width of the container. Removed some duplicate code etc.

The color of the hover was updated to grey as well - but that is a matter of taste...

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/css/style.css
+++ b/src/rockstor/storageadmin/static/storageadmin/css/style.css
@@ -186,7 +186,7 @@ select, textarea, input[type="text"], input[type="password"], input[type="dateti
 .dropdown-menu>li>a:hover,
 .dropdown-menu>ul>li>a:hover {
   color: #FFFFFF;
-  background-color: #444444;
+  background-color: #E76545;
   text-decoration: none;
 }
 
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @grogi I'll test your pr today. If you haven't already, please join our form: http://forum.rockstor.com

schakrava:
@grogi I really like alignment changes and extra spacing in the side and nav bars. But I do like the orange  hover background we currently have. I am wondering what your reasoning behind changing it.

btw, what do you think about showing dropdown menu upon hovering on the navbar versus clicking on an item like we currently have?

grogi:
Over the years I've developed taste for simplicity. But I agree - orange builds branding here, is consistent with the website etc.

When it comes to menus on hover, I am not so sure. I guess once you have one open, going to another one could open it automatically on hover. But I don't really entertain the idea of menus automatically showing up if you just move the mouse.

schakrava:
Thanks for your design feedback @grogi , makes good sense. Also, your point about branding is well articulated.

Sorry for the delay, but I am happy to have merged this pull request! Please contribute more if you can. I'd be happy to give free activation codes for your Rockstor appliances.

PR #1028, opened 2015-11-29
toggle sftp service. Fixes #1027

### LINKED ISSUE #1027: toggle_sftp_service always sent false
master branch build from 24th Nov

toggle_sftp_service in system/ssh.py always receives false even when sftp switch on services page is switched on.
The above is called by:- smart_manager/views/sftp_service.py

Suspect recent changes to switch subsystem but haven't go that far with diagnosis.

Sorry that's as far as I got. Will have a look again when I'm next able.

sftp button in services page will temporarily show OFF then revert to ON.

Consequence of this is sftp becomes permanently disabled due to no sftp subsystem line in:-
/etc/ssh/sshd_config
Usual line would be:-
Subsystem       sftp    internal-sftp
normal placed directly after "# override default of no subsystems"
This line is removed and never replaced, I suspect as a result of the above.

phillxnet:
On install sftp was indicated as ON but on switching OFF the switch would return almost immediately to indicate sftp was ON however the off function was executed and so no sftp service was then active.  But as the switch was then in an ON position it could only send OFF signals, there by blocking activation of Rockstor initiated sftp service (as opposed to that enabled by the default config).

Problem was interpretation of the return code of -1 by the new switch subsystem as ON, fixed by returning 1 in the sftp disabled state ie where no line in sshd_config matches "Subsystem\s+sftp".

Also brought service_status check in system/services.py in-line with previously improved ssh.py check.for sftp substring. This avoids false negatives on initial install.

sftp checked via:-

-disabled behaviour-
sftp root@rockstor.local
subsystem request failed on channel 0
Couldn't read packet: Connection reset by peer

-enabled behaviour-
sftp root@rockstor.local
Connected to rockstor.local.
sftp> 

Tested with clean build and default sshd_config and then again with Rockstor generated config as slightly different sftp substrings.

### DIFF
--- a/src/rockstor/smart_manager/views/sftp_service.py
+++ b/src/rockstor/smart_manager/views/sftp_service.py
@@ -51,7 +51,7 @@ class SFTPServiceView(BaseServiceDetailView):
                 if (command == 'start'):
                     toggle_sftp_service()
                 else:
-                    e_msg = 'non config and non start so assuming stop in sftpserviceview'
+                    e_msg = 'non config and non start so assuming stop in sftpserviceview, received %s' % command
                     logger.error(e_msg)
                     toggle_sftp_service(switch=False)
             except Exception, e:
--- a/src/rockstor/smart_manager/views/sftp_service.py
+++ b/src/rockstor/smart_manager/views/sftp_service.py
@@ -51,6 +51,8 @@ class SFTPServiceView(BaseServiceDetailView):
                 if (command == 'start'):
                     toggle_sftp_service()
                 else:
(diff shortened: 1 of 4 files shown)


phillxnet:
Fixes #1027

schakrava:
I've just tested it and your changes look great @phillxnet . I really appreciate your code comments and github documentation as well. You made it easy to see the bug and test your fixes.

phillxnet:
@schakrava Thanks, I was a little worried I was spamming the issue but though that this way someone else could have taken over where I left off in case I didn't make it to a solution.

PR #1030, opened 2015-11-29
AD_switch_on_in_error. Fixes #1029

### LINKED ISSUE #1029: AD switch on in error
In recent testing builds the AD switch has indicated an ON state incorrectly and attempting to turn it OFF results in:-

Error!
Missing configuration. Please configure the service and try again. Exception: expected string or buffer

Thanks also to @roweryan for reporting this on the forum:- http://forum.rockstor.com/t/why-are-sftp-and-active-directory-enabled-now/658

I suspect this is due to a recent switch subsystem change and a pending fix for a similar issue with sftp looks related. These two miss representing switches were also linked / reported by @roweryan .

phillxnet:
I'm afraid I have yet to setup my FreeIPA system to test Rocktor's sssd registration system but with the supplied patch the new bootstrap switch is now OFF by default and responds appropriately and as far as I can tell as it did prior to this issue arising. 

Note this may well be a work around rather than a fix, please see #1027 for a related issue.

Awaiting more informed review.

Patch notes:-
The new bootstrap switch subsystem reads the -1 rc as
ON state which results in inability to switch AD on as the
subsystem believes it is already ON, leading to a catch 22.

### DIFF
--- a/src/rockstor/system/services.py
+++ b/src/rockstor/system/services.py
@@ -158,7 +158,8 @@ def service_status(service_name, config=None):
             for l in o:
                 if (l == config['domain']):
                     return '', '', 0
-        return '', '', -1
+        # bootstrap switch subsystem interprets -1 as ON so returning 1 instead
+        return '', '', 1
 
     return init_service_op(service_name, 'status', throw=False)
 


schakrava:
@phillxnet I took the opportunity to also fix up a small frontend bug as part of merging this pull request. Thanks for bringing light to this issue!

phillxnet:
@schakrava That's great; Cheers.

PR #1045, opened 2015-12-07
removed disk wrong details. Fixes #937

### LINKED ISSUE #937: removed a defective disk, storage disk showing wrong disk details

phillxnet:
Fixes #937 by enhancing the disk/DiskMixin _update_disk_state by:-

~~1) (New) First saving the information of the drives that were known to the db but are no longer found attached; judged purely on comparing the db device serial numbers and those of the attached devices.
2) (no change) Doing what we have been doing all along; ie overwriting the existing db entries with the information of the attached devices, and adding db entries for those not seen before.
3) (New) populating the remaining db slots with the information gained in 1 above.~~

The update_disk_state() method has been rewritten to take no heed of device names and simply overwrite all dev names with unique but nonsense devices and re-populate these dev names based purely on serial numbers received from a slightly modified scan_disks (see below). This allowed for a simplification of the method which helped to remove a bug that lingered from my first attempt (see the scrikethrough text above).

A significant departure from the previous mechanism is to re-name the name (unique) field in the db of all those drives that are not found to be attached but that were know to the db before the last scan (missing / offline) as a unique 32 char string. This string is updated every time an update_disk_state is performed as it is essentially arbitrary, as are the dev names. However the kernel will ensure dev names are unique at any one point in time but storing offline dev names is redundant and potentially harmful. This way all removed / offline / detached / missing devices will have a dev name of 32 chars long. This should make debugging calling code much easier as it is clear that a device has been retrieved from an offline db entry as it will has a dev name (db name field) 32 random chars long.

Note that a further enhancement to the scan_disks function that _update_disk_state uses to get the information about the connected disks.
Previously if a disk was found to have either an empty serial number or one that duplicates one found earlier in the same scan then scan_disks will store this drives serial number as it's dev name eg "vda" this is problematic in the new scenario as it could easily lead to duplicate serial numbers stored for removed devices, ie if vda was added with no serial and was then removed it would leave in the db an entry with a serial number of vda. The same procedure repeated would then leave duplicate serial entries in the removed section of the db. This complicates the ability of any mechanism to uniquely identify removed devices. To address this the stored (returned from scan_disks) serial numbers for empty or repeat serial numbers found on attached devices is now made unique yet still obviously fake, ie: fake-serial-37691025-308a-41de-b766-72bff06f9a93 = 48 chars long, again this is derived from a uuid4 generator. This way we don't harm our own store of serial numbers. Note that the Web UI previously looked for empty serial numbers or those that were equivalent to the dev name. This is still the case (for legacy db entries) except that we also look for 48 chars long serial numbers and assume they are also bogus; with the associated WebUI warning. ~~This has the nice effect that once deleted these devices maintain their bogus serial status.~~ Devices that have generated fake serial numbers are no longer retained in the db when detatched as given the serials are fake and transient (they are regenerated on each scan) their is no sense in retaining these entries when they are found to be missing. This also prevents an untidy side effect of repeat / fake serial number devices causing a build up of fake removed devices associated with their prior serial numbers.

I have tested the behaviour of current master with this patch set applied in multi data disk arrangements and believe it to be more correct than current behaviour at it's level of responsibility. There is a chance of course that this new behaviour will trigger as yet unseen side effects but I believe them to be less risky than the previous behaviour as we will no longer keep potentially valid device names in the db against removed disks and are much less likely to encounter duplicate serial and device entries in the db which can easily lead to red herrings in logs.

I will attach a series of screen shots indicating the nature of my fix verification test.

I believe this pr is ready for review.


phillxnet:
Verification of desired behaviour, initial state with 3 data drives:-
before-middle-drive-removal-proof-1

The VM is shutdown and the middle 4GB drive is removed (serial number manual-serial-444):-
after-middle-drive-removal-proof-2
Note that this missing / offline drive is now recorded accurately with regard to it's size and serial number.

A view of the pools page while the 4GB drive is offline / detached from the VM.
pools-after-middle-drive-removal-3

The VM was then powered off and the same 4GB drive was re-attached and the VM powered up again.
middle-drive-reattached-4
Note that the offline entry has disappeared and our drive is back as attached.

Also not that our Pools page is displaying what would be expected now that our same drive has returned, become re-attached:
pools-after-middle-drive-reattached-5

"btrfs fi show" during the drives absence and after it's reattachment displayed what would be expected.

I have also tested several other scenarios including those involving multiple new and multiple missing drives and all appear to behave as expected.

N.B. Tests were confined to the level of responsibility of these units, ie no multi volume btrfs tests. However this is not within the scope of the altered code and I am of the opinion that these changes will assist in future debugging and hopefully reduce the chance of using an incorrect device in the first place.

phillxnet:
Also fixed was that removed / offline devices previously had their SMART status and capability stored, this has been corrected to Not Supported / disabled. Upon re-attachment devices are re-evaluated.

phillxnet:
@schakrava I have just discovered that there is still work to be done on this, it fails to address an issue that I was aiming to sort, ie incidentally adding duplicate serial entries under specific drive rearrangements. I can have a go at this again next week to try and improve further as I have a machine and config which provokes this behaviour.

phillxnet:
@schakrava I now think it's going to be safer to not merge this until it's had some more work, I will resume this again on Friday.

schakrava:
Thanks for the comment @phillxnet. I'll wait for you.

phillxnet:
OK, I've done a reworking of the existing method which has resulted in a simplified and more robust mechanism that no longer exhibits the fantom incidental creation of duplicate db entries that share the same information as existing attached devices. I have tested dozens of boot cycles on a system that reliably provoked this occurrence every few boots with no repeat "removed" db entries having occurred.

Please note that I have edited the intro / header text to this pr to reflect the changes after my initial (premature) submission. All claims for the previous submission of this pr have been re-verified and all fail manual test cases detailed in the associated issue #937 have also been re-verified.

More ready for review than last time.

Note that I have added a blanket disable of SMART on all vd* (virtio) devices to reduce log clutter. This may need to be undone later, but should hopefully only affect very niche hw scenarios.

phillxnet:
@schakrava Thanks for your patience on this one.

schakrava:
@phillxnet I really like your new approach. I've started testing, but not done just yet, but I haven't seen any issues so far. Thank you!

phillxnet:
@schakrava Thanks, if it works out I though we could later display the long missing device names in the WebUI as just "missing / offline" or something but for now, early testing channel, it might be good to see them. This UI improvement might go well along with / under #1051.

schakrava:
I agree, there are further changes to improve the whole mechanism so user is aware of what's going on. But these changes solve a important problem with swapping device names and that is plenty for this pull request.

Issue #1063, opened 2015-12-22
enhance scan_disks to handle md block devices


phillxnet:
Although the build in fail over to use udev rather than lsblk to find an md block device's serial number succeeds the fact that it appears twice in lsblk's output (in the case of a 2 device md) then triggers the duplication removal logic and the real udev rule applied serial number successfully retrieved on the first pass is overwritten with an auto generated fake-serial on the second. This in turn causes the "Disk Serial number not legitimate" warning even when a valid serial is available (applied via udev rule).
N.B. this has only been proved on non root md devices but is strongly suspected to apply to the system device also which is the only likely supported use for md devices.

Please update the following forum post on progress with this issue:-
http://forum.rockstor.com/t/issues-cant-wipe-some-drives-no-smart-diagnostics-disk-serial-numbers-not-legitamate/820

phillxnet:
the attached logging patch indicates the issue.
```
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:841] ########### process next block device ############
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:842] dmap['serial'] = WD-WMAET2305538 
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:843] always_use_udev_serial = False 
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:841] ########### process next block device ############
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:842] dmap['serial'] =  
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:843] always_use_udev_serial = False 
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:845] USING the UDEV based get_disk_serial SYSTEM
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:850] get_disk_serial returned = 235711
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:841] ########### process next block device ############
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:842] dmap['serial'] = 6RW17C9A 
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:843] always_use_udev_serial = False 
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:841] ########### process next block device ############
[22/Dec/2015 20:31:49] DEBUG [fs.btrfs:842] dmap['serial'] =  
[...]

PR #1071, opened 2015-12-29
enhance scan disks for md #1063

phillxnet:
In the case of md devices such as software raid ie mdadm and some hardware raid, the md devices and their respective partitions have repeat listings in lsblk, this accidentally triggers the duplicate device by serial number mechanism as the entries are identical; the consequence of this is a fail-safe that overrides the serial on the second occurrence (which has the same dev name) with a unique but fake serial. Avoid this udev assigned serial accidental override by only processing the first occurrence of each device by name in the lsblk output.

@schakrava I will have another go at testing this on a software raid system disk tomorrow (UK).

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -809,17 +809,14 @@ def scan_disks(min_size):
                 i = i + 1
             else:
                 raise Exception('Failed to parse lsblk output: %s' % sl)
-        logger.debug('NAME found from lsblk parse was = %s ' % dmap['NAME'])
         # md devices, such as mdadmin software raid and some hardware raid block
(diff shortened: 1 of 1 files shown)


phillxnet:
Haven't managed to install system disk on software raid as installer doesn't offer btrfs with raid. This would be a nice one to work out though. Can be done via installer if one selects eg XFS as root but then we arn't testing Rockstor as is. Also doesn't seem to present identically as original poster's system as their intel motherboard hardware md is presented as a partitioned md where as the installer goes with multiple md devices in software raid. So not equivalent. May revisit this attempt but taking too long. Would be nice to document a working solution when found though.

schakrava:
Thanks @phillxnet!

phillxnet:
Cheers @schakrava

PR #1076, opened 2016-01-02
no smart info on msa70. Fixes #997

### LINKED ISSUE #997: no SMART info on MSA70
Thanks to @titus_36 on the forum for reporting this.
It seems there is no WebUI SMART info displayed yet it is available on the command line.
Reason seems to be that the info is differently formatted to what the current parser expects and there is much less of it. Original reporter is hopefully to send in no-smart-msa70-report.tar.gz file resulting form commands requested on the following forum thread:- http://forum.rockstor.com/t/no-smart-support-for-msa70/641

Please update the forum post on progress with this issue.

phillxnet:
By adding a mode that reads smartctl outputs directly from file dumps of the same I was able to use @priel71 's smartctl outputs from their lsi9207-8i setup and correct some parsing issues.
The most noteworthy of which caused the existing parser to report no smart availability.
Since some smart info is not presented in a standard way I have also modified the UI presentation to dump raw output of for example Error Logs when no parsing for a summary was successful but when error info exists. Also the patch set adds info to user of no Attributes and Capabilities returned from some drives rather than just showing an empty table.
I was unable to verify Error Logs & Self-Test Logs tab function as the supplied outputs had no info on either of these. I am assuming if present this info is formatted similarly to what has worked to date.
Where logs were previously saved verbatim but not displayed as such I have via "pre" tags reproduced the original formatting as this is important for interpreting tabulated / spaced output; and in the case of error logs is the only means of viewing them. There is apparently no standard formatting on non ATA/SATA drives.

All changes have been tested with several existing drives that have error logs and test logs and drives that have neither. 

Note that main presented in the Identity / Info tab is paired with equivalents and no db / model changes were required.

Fixes #997 Although the modifications were based on a later submitter they are assumed to be equivalent. I have re-contacted the original reporter on the forum but have not heard back.

Possible simple improvements are to remove the redundant double spacing in the display of:-
Raw S.M.A.R.T error log
non summary part of Self-Test Logs
This is currently better readability wise but aesthetically challenged.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
@@ -281,10 +281,12 @@
 	  <% }); %>
 	</tbody>
       </table>
+			<pre>
       <% testlogdetail.forEach(function(detail) { %>
-      <%= detail.line %><br>
+      <%= detail.line %>
       <% }); %>
       <% } %>
+			</pre>
     </div>
   </div>
 
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
@@ -167,39 +167,43 @@
       </div>
   </div>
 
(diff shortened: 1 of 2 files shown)


schakrava:
Thanks @phillxnet. Comments are super helpful!

PR #1090, opened 2016-01-11
Make access key delete based on id instead of name. Fixes #1080

### LINKED ISSUE #1080: Error while deleting Access key with space in the name
We should probably change the API to be called with ID instead of name.

Steps to reproduce:
1. Create an access key with space in it's name.
2. Delete it. It fails with client error.

Mchakravartula:


schakrava:
nice job @Mchakravartula !

PR #1091, opened 2016-01-11
raw smart error logs missing. Fixes #1084

### LINKED ISSUE #1084: raw smart error logs missing
Minor regression in 3.8-10.15. In 3.8-10.12 the raw smart error log was displayed as a section below the summary of errors in the "Error Logs" tab. This is important as it is the only place where some smart errors are currently visible ie some scsi style reporting such as that from the lsi controller output used to address issue #997 which was previously fixed by pr #1076.

phillxnet:
Corrects minor regression in display of raw smart error logs.
Created a custom handlebarsjs comparison block helper to restore simple view logic on wether to show raw logs or not.
Please see #1084 for output after patches.
Tested on a sata drive with errors and some without and on submitted smart dumps (using internal testmode) for lsi9207-8i , the same dumps used when this enhancement was originally added in #1076 .

N.B. The raw log outputs are now no longer double spaced which is nice.

Ready for review as works for me in both Chrome and Firefox.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
@@ -194,48 +194,52 @@
       </table>
     </div>
   </div>
-  <div id="smarterrorlogs">
-    <div class="row">
-      {{#if errorLogSummaryNull}}
-        <h3>There are no errors.</h3>
-        {{else}}
-        <p>{{errorlogZero}}</p>
-	<p>{{errorlogOne}}</p>
-      <table id="smarterrorlogsummary-table" class="table table-condensed table-bordered table-hover table-striped tablesorter" summary="Table of S.M.A.R.T error log summary">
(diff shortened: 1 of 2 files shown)


phillxnet:
@schakrava I left the file indented with tabs and not spaces to help the readability of the changes.

schakrava:
Thanks @phillxnet ! Changes look good. But you don't need to tab indent changes.

phillxnet:
@schakrava Cheers. I thought the file already used tabs so continued with tabs. The existing indents were inconsistent so corrected but with tabs not spaces, that was probably it. I take it we should always use spaced like in Python, even if we find the file already has tab intends? I have PyCharm setup to use 2 spaces for js, jst files and 4 spaces for Python. But I think it will prioritize what it already finds and in this case it was tab indented or at least that is what it said. I can in future change files to spaces as I come across them if that suites?

phillxnet:
OK, I get it now I think. Some of this white space is active. Sorry wasn't thinking straight.

PR #1103, opened 2016-01-16
mend share permissions edit in testing channel. Fixes #1100

### LINKED ISSUE #1100: Share Access control issue found on latest unstable/test updates
I believe I found some bug with "Access control" in a share a user makes and then edits.  I can repeat the issue as I wasn't sure what broke until after some testing.
 
Symptoms:  When editing a created share's permissions,(in the webgui) for example: enabling read/write/execute for all 3, owner/group/other (for example 777). Upon saving, it then goes to read only for owner/group/other (444), despite selecting other options.

Default install of 3.8-10(Linux: 4.2.5-1.el7.elrepo.x86_64) ISO does not exhibit this behaviour.  Upon activating Testing updates, and installing/rebooting (Rockstor 3.8-10.17) (Linux: 4.3.3-1.el7.elrepo.x86_64), the above symptoms now arrive.

Installing updates after clean install are only things done.  New or old shares created and then edited now present the "read only" option despite selecting other options for access control.

phillxnet:
As reported in #1100 by @parker1c share permission edits would result in 444 access writes irrespective of selection. Turned out to be minor omission in recent massive UI overall.

I have tested this commit when applied to 3.8-10.17 and the share permissions both edit and view behaved as expected. Owner and Group fields were also tested for edit and display and all 9 tick boxes responded as expected. ie Owner Group Other for Read Write and Execute permission. Verification was carried out via:-
```
watch -d ls -la /mnt2
```
while saving edits and viewing edits on a number of shares.

Fixes #1100 .

@schakrava Ready for review and submitting as tested via the above method.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
@@ -492,21 +492,21 @@ ShareDetailsLayoutView = RockstorLayoutView.extend({
 			if (this.permStr[0] == "1") { 
 				html += '<input type="checkbox" name="perms[]" value="owner-r" checked="true">';
 			} else { 
-				html += '<input type="checkbox">'; 
+				html += '<input type="checkbox" name="perms[]" value="owner-r">';
 			} 
 			html += '</td>';
 			html += '<td>';
 			if (this.permStr[3] == "1") { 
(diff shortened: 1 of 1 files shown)


schakrava:
Really nice catch and fix @phillxnet! Thanks a bunch!

phillxnet:
@schakrava Cheers.

PR #957, opened 2015-10-14
smart data collection broken Fixes #657

phillxnet:
In some instances it seems that smartctl -l error returned a command line error of 64. This was correctly caught by Rockstor but interpreted as a failed command execution. In tests on several drives that displayed this wall of red text error that looked like a failure this pr produces valid smartctl error logs.
As this is an error condition I have added an email root function that is executed when this error code is encountered. N.B. if one refreshes smart data read via the web ui refresh button on a drive with S.M.A.R.T errors logged this email will be triggered. Note that I don't thing that all drives are capable of self logging.

Submitting as is for code review and further testing / analysis of my silencing an exception.

email part tested with and without email notifications enabled.

PR #1108, opened 2016-01-17
enhance smart return code handling. Fixes #1107

### LINKED ISSUE #1107: enhance SMART return code handling
I have observed another return code from smartctl execution that if not caught leads to a failure to populate the smart tabs when the data is otherwise available and the command also executes as expected. This is akin to the circumstance addressed in pull request #957 only this time the command is:-
```
smartctl -l selftest -l selective /dev/sdb
```
and the exit code is 128
```
$?
-bash: 127: command not found
```

phillxnet:
Simple enhancement to deal with return code of 128 from the smart command used to retrieve self test logs that otherwise produced sane output. This avoids all smart data being inaccessible for drives causing this return code on this smart probe.

I have tested on a single drive that exibits this behaviour and a number of other drives that behave more normally.

Note this pull request abstracts the mechanism that was previously used to override specific return codes that was introduced in pull request #957 so future codes that we need to handle differently can be more easily incorporated.

### DIFF
--- a/src/rockstor/system/smart.py
+++ b/src/rockstor/system/smart.py
@@ -239,9 +239,9 @@ def test_logs(device, test_mode=TESTMODE):
     :return: test_d as a dictionary of summarized test
     """
     if not test_mode:
-        o, e, rc = run_command(
-            [SMART, '-l', 'selftest', '-l', 'selective', '/dev/%s' % device],
-            throw=False)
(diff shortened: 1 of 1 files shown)


phillxnet:
Fixes #1107 
Pull request ready for review.

schakrava:
Looks great @phillxnet, thanks!

phillxnet:
@schakrava Cheers.

Issue #1110, opened 2016-01-19
mouse over hints too narrow


phillxnet:
The yellow mouse over hints that appear on mouse over events are barely 2 word wide, I think they should be wider so that they don't require as much height on the screen and gain readability.

schakrava:
Fixed by #1126

PR #1126, opened 2016-01-25
Increase the width of tooltip hint. #1110

Mchakravartula:


schakrava:
Thanks @Mchakravartula. that did it!

PR #1140, opened 2016-01-31
populate disk info for system disk. Fixes #1116

### LINKED ISSUE #1116: populate disk info for system disk
The Model, Transport, Vendor entries on the Storage - Disks page for the system disk are always null; it would be nice to have these show the correct information.

phillxnet:
Relatively non invasive additions that extends the existing mechanism that populates the root partition's serial entry to also populate its model, transport, vendor, and hctl entries.

Tested on a couple of real machines and all looks fine. These values would have previously been attributed the None value.

Corrected and improved some comments in the same area.

Fixes #1116

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -780,7 +780,7 @@ def scan_disks(min_size):
     dnames = {}
     disks = []
     serials = []
-    root_serial = None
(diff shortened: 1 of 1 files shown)


phillxnet:
@schakrava Ready for review.

phillxnet:
@schakrava Cheers for the merge.

PR #1158, opened 2016-02-09
improve bios raid handling. Fixes #1151

### LINKED ISSUE #1151: improve bios raid handling on system disk
When using bios raid, such as Intel Raid Storage Technology on the system disk there is no system disk identification due to differences in the device naming scheme, ie /dev/md126p3 rather than the expected /dev/sda3. This causes the base md126 as well as the swap partition to be listed in the disks page all with inappropriate flags.

This has been seen in the forum in the following post:-
http://forum.rockstor.com/t/issues-cant-wipe-some-drives-no-smart-diagnostics-disk-serial-numbers-not-legitamate/820/11

I have reproduced the above system disk layout locally:-
3 8-11-default-install-irst

phillxnet:
I have tested this patch set on a variety of machines both real and virtual with regular installs and a single Intel bios raid install. Intel bios raid is essentially a bios level instantiation of software raid that in linux is managed as part of the standard mdadm subsystem. This allows for the bios to define a single bootable drive from say 2 and this single raid drive shows up as a bootable bios raid during the installer but isn't actually built until after the install.
Due to differences in naming the existing scan_disks and root_drive functions in btrfs were unable to identify what we had, from manual testing this looks to be resolved with no regression identified.
N.B. there is still the issue of the raid members showing up with the cog icon but this can go into another issue as we are big enough here already. Also may require a disk role that ties into other pending issues.

Minor logic changes were applied to scan_disks but only after fairly extensive readability changes and a generous about of comments to aid in this and hopefully future changes.
Of note is the exclusion of swap in a far simpler and earlier fashion than was used previously. Also the inclusion of md as a type that identifies a partition, this was confirmed in a regular md setup and the bios raid setup which were the same with regard to naming.

Please see the #1151 issue for testing prior to pr submission and more details.

Note the model field of the md root device is used to store the member disk serial numbers and their raid index as well as the total disk count.
I have also extended the get_disk_serial() function to return the md uuid as device serial to avoid having to create udev rules for md devices.

Submitting as works for me and I don't think I've done anything risky here but this is a prime candidate for careful review.

Hope it works and if there are problems then I'm happy to chip away some more.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -769,8 +769,7 @@ def root_disk():
             fields = line.split()
             # todo does this not allow ext4 root, ie silencing no btrfs root
             # todo exception raised here otherwise?
-            if (fields[1] == '/' and
-                    (fields[2] == 'ext4' or fields[2] == 'btrfs')):
+            if (fields[1] == '/' and fields[2] == 'btrfs'):
                 disk = os.path.realpath(fields[0])
                 if (re.match('/dev/md', disk) is not None):
                     # We have an Multi Device naming scheme which is a little
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -767,8 +767,6 @@ def root_disk():
     with open('/proc/mounts') as fo:
         for line in fo.readlines():
             fields = line.split()
-            # todo does this not allow ext4 root, ie silencing no btrfs root
-            # todo exception raised here otherwise?
             if (fields[1] == '/' and fields[2] == 'btrfs'):
                 disk = os.path.realpath(fields[0])
                 if (re.match('/dev/md', disk) is not None):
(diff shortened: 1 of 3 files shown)


src/rockstor/fs/btrfs.py
```diff
@@ -871,26 +945,38 @@ def scan_disks(min_size):
             elif (size_str[-1] == 'T'):
                 dmap['SIZE'] = int(float(size_str[:-1]) * 1024 * 1024 * 1024)
             else:
+                # Move to next line if we don't understand the size as GB or TB
+                # Note that this may cause an entry to be ignored if formatting
+                # changes.
+                # Previous to the explicit ignore swap clause this often caught
+                # swap but if swap was in GB and above min_size then it could
+                # show up when not in a partition (the previous caveat clause).
```
schakrava: Nice catch.

phillxnet:
@schakrava Thanks.

PR #1165, opened 2016-02-16
Change font and color of banner elements

gkadillak:
Instead of the green elements, I propose that instead of the green elements, they should be white and the font should match that of the rest of the banner. Example shown below.

image


schakrava:
Good to see you back @gkadillak

PR #1166, opened 2016-02-16
improve root not in partition behaviour and mdraid member labeling. Fixes #1164

### LINKED ISSUE #1164: improve root not in partition behaviour
With an mdraid install (system disk) where root (btrfs) is on for example md126, ie not in a partition, the Disks page fails to display the system disk and the rockstor_rockstor pool is also not found.
I have made a start on modifying scan_disks to account for this arrangement to improve reporting of the system disk and the associated rockstor_rockstor pool.

phillxnet:
@schakrava  Please see #1164 for development details and testing.

Appears to maintain existing functionality whilst allowing for / on an md device directly rather than in a partition. Mdraid disk members, both intel bios raid and pure software mdraid, are now labelled as such and no longer have the "wipe me" invite flag.

Changes outside the remit of this issue:-
I have added a smart_options to the Disk model for the proposed custom smart options per disk. This I thought would cut down on the number of db migrations and ease the associated development on that issue.

As this patch set alters the categorization logic of scan_disks() it requires careful review, also this is the first model change I've made so please advise if I need to do more than I have anticipated.

Submitting as works for me.

N.B. I see the disk role field as also serving future flagging of such things as temp externally connected usb drives used for backup or import, where their once labelled 'role' will be maintained in the db against their serial number as long as their entry is not deleted from disk.

### DIFF
--- a/src/rockstor/storageadmin/models/disk.py
+++ b/src/rockstor/storageadmin/models/disk.py
@@ -40,6 +40,13 @@ class Disk(models.Model):
     vendor = models.CharField(max_length=1024, null=True)
     smart_available = models.BooleanField(default=False)
     smart_enabled = models.BooleanField(default=False)
+    """custom smart options for drive, ie for USB bridges / enclosures"""
+    """eg "-d usbjmicron,p" or "-s on -d 3ware,0"."""
+    smart_options = models.CharField(max_length=64, null=True)
+    """drive role ie "isw_raid_member" or "linux_raid_member"."""
+    """could also be "import" or "backup" for temp external drive connection."""
+    """role is aux info to flag special use disks"""
+    role = models.CharField(max_length=256, null=True)
 
     @property
     def pool_name(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


src/rockstor/storageadmin/models/disk.py
```diff
@@ -40,6 +40,13 @@ class Disk(models.Model):
     vendor = models.CharField(max_length=1024, null=True)
     smart_available = models.BooleanField(default=False)
     smart_enabled = models.BooleanField(default=False)
+    """custom smart options for drive, ie for USB bridges / enclosures"""
+    """eg "-d usbjmicron,p" or "-s on -d 3ware,0"."""
+    smart_options = models.CharField(max_length=64, null=True)
```
schakrava: Is 64 character limit enough for this option? Since it becomes a VARCHAR in postgres, I think it's ok to be a bit generous.

phillxnet: As discussed in our private chat but for the record I'll pop in here as well. The longest I could find was "-T verypermissive -a -d usbjmicron,x,1" so 38 chars from:- https://www.smartmontools.org/wiki/Supported_USB-Devices

schakrava:
Just got done with testing. Really clean patch set. Thanks @phillxnet

phillxnet:
@schakrava Cheers.

PR #1175, opened 2016-02-20
Remove console.log preventing rockon restart

nicolaslt:
A console.log trying to log an undefined var was preventing the user to be able to restart a stopped rock-on from the web ui.


schakrava:
Nice catch. Thanks @nicolaslt !

PR #1203, opened 2016-03-05
Change ipv4 address key

todelod:
Changed the key from which the ipv4 address is read (old: IP4.ADRESS[1], new: ipv4.addresses)

With the old key one could not use ip-address aliases on interfaces.

It would be a nice feature to be able to handle them (the aliases in rockstor gui, or at least se them) but as the first step I thought I could prevent rockstor from using the wrong ip-address.

Output from nmcli -t c show enp2s0 (in my case) 

ip a add 192.168.0.14/24 dev enp2s0
(creates a non-persistent alias on the interface)

[nmcli_single_ip_output.txt](https://github.com/rockstor/rockstor-core/files/160031/nmcli_single_ip_output.txt)

[nmcli_with_ip_alias_output.txt](https://github.com/rockstor/rockstor-core/files/160032/nmcli_with_ip_alias_output.txt)

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -279,7 +279,7 @@ def net_config_helper(name):
 
             elif (config['method'] == 'manual'):
                 #manual
-                if (re.match('IP4.ADDRESS', l) is not None):
+                if (re.match('ipv4.addresses', l) is not None):
                     kv_split = l.split(':')
                     if (len(kv_split) > 1):
                         vsplit = kv_split[1].split('/')


schakrava:
Hello @todelod . Thanks for opening this pull request + documenting it very well. And sorry for the delay in my response.

I see your point, but "ipv4.addresses" key is not populated for dhcp configurations. Have you tested that?

I am a little confused what you mean by wrong ip-address. Do you mean Rockstor should pick up the first address in case there are multiple? If so, isn't it better to grab the "IP4.ADDRESS[1]"? That is populated both for manual and dhcp configurations and also works with your output.

Let me know what you think. I'd love to merge your pull request soon!

todelod:
Hi!
I'm not that familiar with centos and I hadn't used nmcli before. (Hadn't tested DHCP for instance)

The problem seems, at least for now, to be that the main address don't become  no. [1] but rather 2 och 3. I'm not certain that it has to be the last either.

If I use ip command and run 'ip a'

`2: enp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP qlen 1000
    link/ether 00:9c:02:a7:ef:9b brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.12/24 brd 192.168.0.255 scope global enp2s0 valid_lft forever preferred_lft forever
    inet 192.168.0.14/24 scope global secondary enp2s0 valid_lft forever preferred_lft forever
    inet6 fe80::29c:2ff:fea7:ef9b/64 scope link valid_lft forever preferred_lft forever
`

It is obvious that the second address is an alias, but with nmcli I havn't found the same information.

I will try to find som more time to look in to the issue.

Thank you for your time and effort.

schakrava:
Thanks. I look forward to your new findings and code changes. Please document as elaborately as you can.

phillxnet:
@todelod Closing this as now very old and doing general house keeping. Please re-open, or better still re-submit any revised code if you fancy having another go at this or anything else any time. Thanks again for your efforts.

Issue #1206, opened 2016-03-06
smart commands act on root partition not root device


phillxnet:
When gathering smart info the partition device name for root is used rather than the base device name.
This is due to the exception made for root, when installed on a partition (default), of storing the partition name rather than the base device name ie sda3 rather than sda. Rather than loose this information we could make the smart routines robust to these calls. Note that mostly these partition rather than base device calls function as expected. However in some instances when calling smartctl on a device's partition we loose information eg:
```
/usr/sbin/smartctl -l selftest -l selective /dev/sdc3
smartctl 6.2 2013-07-26 r3841 [x86_64-linux-4.3.3-1.el7.elrepo.x86_64] (local build)
Copyright (C) 2002-13, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF READ SMART DATA SECTION ===
Device does not support Self Test logging
```
and we have a return code of 4 (via $?)

Where as if we use the base device name of sdc:
```
/usr/sbin/smartctl -l selftest -l selective /dev/sdc
smartctl 6.2 2013-07-26 r3841 [x86_64-linux-4.3.3-1.el7.elrepo.x86_64] (local build)
Copyright (C) 2002-13, Bruce Allen, Christian Franke, www.smartmontools.org

/dev/sdc: Unknown USB bridge [0x0781:0x5580 (0x010)]
Please specify device type with the -d option.

Use smartctl -h to get a usage summary
```
we have a generic return code of 1 (via $?) and in the second instance we have helpful info on what is wrong and a suggestion on what to try which would be nice not to loose.

These examples were as a result of an install to a SanDisk Extreme USB 3.0 which as the second example suggests, requires custom smartctl options (-d sat as it goes); but with the current call to partition this is not surfaced. Hence the suggestion we move to calling only on base devices.

I am currently looking into this issue and hope to make some more progress tomorrow (UK).

I am considering this issue as a precursor to #1079 as if sorted then the suggestion that additional custom smart options may be required will be surfaced.

phillxnet:
I am currently working on this issue.

phillxnet:
Turns our this issue was also providing false positives of smart availability when no useful info was actually available without custom smart options.

phillxnet:
A parsing issue I'm sorting while here is that of inadvertent smart value trimming when values contain a ": " in the Identity tab (from smartctrl -H --info devname):
The following two values show the result:
```
In smartctl database [for details use
[...]

Issue #1079, opened 2016-01-05
add custom SMART parameter option


phillxnet:
Thanks to @sudowoodo and @kcomer in the following forum threads for identifying the specific need for custom smart options, ie "-d sat" and "-d 3ware,0" respectively.
http://forum.rockstor.com/t/s-m-a-r-t-for-usb-drives/834
http://forum.rockstor.com/t/s-m-a-r-t-support-is-not-available-for-this-disk/830/7
Please update the above forum threads on progress with this issue.

N.B. a complexity of this issue is that this custom setting needs to be stored / applied per drive to account for mixed drive / controller environments.

phillxnet:
Noting smartmontools page for usb devices / bridges and their respective custom options:-
https://www.smartmontools.org/wiki/USB

phillxnet:
Awaiting possible extension of Disk model to accommodate custom smart options per disk.

phillxnet:
I have an external USB to SATA enclosure that requires "-d usbjmicron,p" to have smart function so I can have a look at this issue once the approach to take with regard to how / where we store the custom smart options is finalized.

phillxnet:
We now have a "smart_options" db field in our disk model:
src/rockstor/storageadmin/models/disk.py which was added in:
https://github.com/phillxnet/rockstor-core/commit/cc7cfc1750cafa43b3a7c8192faa1bdce5c22f10

phillxnet:
Having only had a quickly discussed with @schakrava on this issue I propose that we have an icon next to the cell  for each drive on the Disks page under the S.M.A.R.T column, akin to the icons in the drive name column and serial column, This icon's mouse over will indicate the possibility of adding custom smart parameters, clicking will enable that option by presenting either a dialogue or a page to allow for entry of these parameters. This icon will always be presented irrespective of the smart available status as in some cases (SanDisk Extreme USB 3.0 for example) the custom parameters will change the reading of smart availability (when calling the base device which is proper). A disk re-scan via _update_disk_state() will be required to update this state but this is called anyway when refreshing the Disks page.

The UI logic of the Disks table will show existing previously entered custom parameters by the aforementioned icon, if any such options exist. As these options are linked to the disk model they should travel with the serial number of the device, given the function of _update_disk_state() which will have to be modified to preserve these settings over a disk refresh.

I have started on adding this feature as indicated and will be working against this issue with any relevant commits.

phillxnet:
[...]

PR #1212, opened 2016-03-08
smart commands act on root partition not root device #1206 

phillxnet:
Introduces a smart device name pre-processor so that we can translate for example sda3 to sda.
This gives more consistent SMART behaviour and more informative results.
See Issue #1206 for examples.
Should only have an effect on partition names ie the default install of root in sda3 will now no longer call smart commands with sda3 but this sda.

Note however that in some circumstances this leads to a system device being labelled as having smart available and enabled but after this pr is applied it will be recorded as Not Supported.

This to me is correct as in the first instance although some smart Identity info and status is retrieved, nothing else is and the error message displayed (due to non zero return codes) when attempting to do a smart info refresh is not helpful. As calling on the base device is more informative we get helpful info in rockstor.log and all other calls are blocked due to no smart available returns with the wrapper in place.

An example device used in proving this fix was the SanDisk Extreme USB 3.0 which requires custom smart options. Via the manual test method of reading from smart output file dumps I was able to retrieve and present full smart status info by calling on the base device (along with the custom -d sat). Where as using the partition device name led to misinformation as to status and availability as no other smart calls would succeed without custom options.

This pr is a pre-cursor to that required for #1079 and may also help to accommodate the custom smart device names used by some 3ware controllers.

Also tested on a bios raid install and a regular mdraid install.

Additionally there is a minor Identity tab values parsing improvement.

### DIFF
--- a/src/rockstor/system/smart.py
+++ b/src/rockstor/system/smart.py
@@ -74,8 +74,6 @@ def info(device, test_mode=TESTMODE):
                 # Assume all characters after colon are the result / value and
                 # strip off begin and end spaces. Limit to 64 chars for db.
                 res[i] = line[first_colon + 1:].strip()[:64]
-                logger.debug('extracted result / value of res[i] = %s', res[i])
-                logger.debug('char count of this line = %s', len(res[i]))
     # smartctl version is expected at index 14 (15th item)
     res.insert(14, version)
     return res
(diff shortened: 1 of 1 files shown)


phillxnet:
@schakrava Ready for review.
Fixes #1206

schakrava:
Thanks @phillxnet! Everything looks good.

phillxnet:
@schakrava Cheers.

PR #1213, opened 2016-03-08
smart selftest log parsing error. Fixes #1207

### LINKED ISSUE #1207: smart selftest log parsing error during self tests
Some hdd models eg Seagate ST3000VN000 update their Self-Test logs during a test rather than only after the test is complete. But this live update is not elegantly dealt with by the current parsing system and causes the following error to be thrown directly after initiating a self test:
```
Error!

invalid literal for int() with base 10: "
```
this blocks smart refreshes for the drive in question during the period of the self test (ie up to 6.6 hours) and is due to a parser failing to split the appropriate fields.

A normal line output from **/usr/sbin/smartctl -l selftest -l selective /dev/sda**:
```
# 7  Conveyance offline  Completed without error       00%      1032         -
```
A problematic line is of the form:
```
# 1  Conveyance offline  Self-test routine in progress 10%         6         -
```

Given we currently split by multiple space the problematic "Self-test routine in progress 10%" is read as a single entry as all other columns can be split by multiple spaces but in this case we have only one between the "progress" and "10%", this in turn leads to an error as the following field is wrongly assumed to be a percentage.

I have started to approach this issue and hope to make more progress soon.

phillxnet:
Introduce a specialized parser to account for in-progress Self-Test log entries.
Tested normal operation on a number of both well and ill drives and specialized parsing on trigger drives of Seagate ST3000VN000 both in first 9 hours of life (by editing smart out file dumps and using test mode) and in post 10 hour life where initial error was no longer generated but parsing was still faulty.

All parsing and behaviour, error wise, is now as expected with existing and trigger drives.

@schakrava Ready for review.

Fixes #1207

### DIFF
--- a/src/rockstor/system/smart.py
+++ b/src/rockstor/system/smart.py
@@ -255,6 +255,8 @@ def test_logs(device, test_mode=TESTMODE):
                     for j in range(i + 2, len(o)):
                         if (re.match('# ', o[j]) is not None):
                             fields = re.split(r'\s\s+', o[j].strip()[2:])
+                            logger.debug('fields extracted are as follows %s', fields)
+                            logger.debug('problematic field 3 = %s', fields[3])
                             fields[3] = 100 - int(fields[3][:-1])
                             test_d[fields[0]] = fields[1:]
                         else:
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet! looks good.

phillxnet:
@schakrava Cheers.

PR #1221, opened 2016-03-20
add custom smart parameter option. Fixes #1079

### LINKED ISSUE #1079: add custom SMART parameter option

phillxnet:
Adds per disk custom S.M.A.R.T capabilities via simple specialized input page which includes links to upstream docs for smartmontools. All entered config is displayed on the Disks page and can be removed by submitting an empty entry in the same config page. The new config area is indicated as for advanced use only and a Disks page Rescan is required to apply the entered configuration; this allows for a before and after feedback; most often an update from "Not Supported" to the smart Switch appearing in the On state.

Of note this that this pr depends on the disk.smart_options db field added in pr #1166 in commit:
https://github.com/rockstor/rockstor-core/pull/1166/commits/cc7cfc1750cafa43b3a7c8192faa1bdce5c22f10

Custom per disk smartctl switches are limited (via input validation) to -d (device) and -T (tolerance) and an explanation is offered to inform the user that these are in addition to the usual internally used already -a --info etc switches.

Non rudimentary Input validation is included but not complete (especially in raid controller -d options and associated targets). The validation mechanism used is easily extensible to allow for future acceptance of as yet unknown options for these switches or if need be additional switches.

In the case of a raid -d option I have left the option to specify a specialized raid device target as optional. With disks behind some raid controllers the format of the smartctl command changes so that one changes the end device from for example the disk device (which of course may be a compound / raid device now) to the controller raid device. The raid -d option is then used to stipulate the port / wire that the disk of interest is attached to.
non raid syntax:
smartctl -a /dev/sda
raid syntax:
smartctl -a -d 3ware,0 /dev/twl0
This means the first raid controller /dev/twl0 and the first port/drive ie the ",0" bit.
This currently is a manual mapping but at least we now have the facility to accommodate this. The optional nature of adding this raid controller dev target is to allow for alternative controllers (some controllers don't use the alternative syntax) and to allow for diagnosing issues as the output from the two formats on controllers that support both is very different. Usually a great deal more info is available using the specialized syntax and very little using the non raid syntax.

Thanks to @kcomer on the forum for his assistance and testing of raid specific smartctl command syntax. 
https://forum.rockstor.com/t/s-m-a-r-t-support-is-not-available-for-this-disk/830/24?u=phillxnet

I have tested this patch set on several real and virtual machines and no regressions were found to other S.M.A.R.T function. (Rockstor versions from 3.8-11.20 to 3.8-12.03)

Included is a couple of minor fixes found while adding enhanced user feedback in the case of disabled or unavailable S.M.A.R.T info.

@schakrava Apologies in advance for:
1: my js code. :) (a little inconsistent but it does work and is commented and readable).
2: duplication of many commits in this pr (it's smaller than it looks on commit count). Bit of a git hiccup on my part I'm afraid.

This patch set builds on pr #1212 as it modifies and augments it's abstraction layer for device name preprocessing. But given #1212 has now been merged I am assuming this should not cause any problems.

Images to follow of UI changes.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/router.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/router.js
@@ -36,6 +36,7 @@ var AppRouter = Backbone.Router.extend({
 	"home": "showHome",
 	"disks": "showDisks",
 	"disks/blink/:diskName": "blinkDrive",
+	"disks/smartcustom/:diskName": "smartcustomDrive",
 	"disks/:diskName": "showDisk",
 	"pools": "showPools",
 	"pools/:poolName": "showPool",
@@ -166,6 +167,14 @@ var AppRouter = Backbone.Router.extend({
 	this.renderSidebar('storage', 'disks');
 	this.cleanup();
 	this.currentLayout = new BlinkDiskView({diskName: diskName});
+	$('#maincontent').empty();
+	$('#maincontent').append(this.currentLayout.render().el);
+    },
+
+	smartcustomDrive: function(diskName) {
+	this.renderSidebar('storage', 'disks');
+	this.cleanup();
+	this.currentLayout = new SmartcustomDiskView({diskName: diskName});
 	$('#maincontent').empty();
 	$('#maincontent').append(this.currentLayout.render().el);
     },
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disk_details_layout.jst
@@ -22,13 +22,17 @@
 <div class="messages"></div>
 <!-- page heading -->
 
-{{#if diskSmartNotAvaialble}}
+{{#if diskSmartNotAvailable}}
 <div class="alert alert-danger">
-  <h4>S.M.A.R.T support is not available for this disk ({{diskName}})</h4>
+  <h4>S.M.A.R.T support is not available for this disk ({{diskName}}).<br>
+	Please see the S.M.A.R.T column entry for this disk on the Storage - Disks
+	page; custom options may be required.</h4>
 </div>
 {{else if diskSmartNotEnabled}}
 <div class="alert alert-warning">
-  <h4>Warning! S.M.A.R.T support is not enabled for this disk ({{diskName}}).<br>If available, You can enable it in the Storage screen.</h4>
+  <h4>Warning! S.M.A.R.T support is not enabled for this disk ({{diskName}}).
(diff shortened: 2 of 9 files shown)


phillxnet:
2 SanDisk Extreme USB stick devices, one now has smart capability available via a custom option addition of "-d sat":
custom_smart_options_sandisk_extreme_usb

Config page presented on edit (pen) icon click:-
custom_smart_options_form

An external USB caddy after having it's S.M.A.R.T capability enabled via in this case "-d usbjmicron,p", previously the "Not Supported" message would be displayed, also note the tooltip.
custom_smart_options_enabled

phillxnet:
@schakrava pr ready for review.

schakrava:
Really thorough comments both in the code and here, very helpful. Thank you very much. Your js is just fine, don't worry about it. Take a look at the two commits I just pushed. Indentation was not consistent. Secondly, you don't need an _else_ following an _if_, if there's a return statement in the _if_, just a minor thing.

Please do give another round of testing and let me know.

phillxnet:
@schakrava Thanks. I see what's happened indentation wise (in part) now and have taken steps to avoid this in the future, I have enabled show-whitespaces (copying here to help other contributors):
show-whitespaces

So I now see a mixture of spaces and tabs, the tabs I inherited from the files I cut and paste some of the code in here from ie blink_disks.js but I (via PyCharm) used spaces (in sets of 4 in js files) there after. However I see from your commits in this branch that you used tabs. I had assumed from the last time I made a mess of this we were to use spaces but that was in Python not JS.

Would you like me to change the files you modified all to spaces or all to tabs; being wise to my previously flaky indentation as I go? Cheers.

phillxnet:
@schakrava Cheers re the else after if return clean up, silly me.
I have reformatted the js files touched in this pr (except router.js), all of which had mixed tab and space indent. No additional code changes were made.
I've tested this pr with the above patches applied and all still looks OK here.

disk_details_layout.jst also remains mixed on the tabs and spaces indent front however.

Let me know if you want me to also re-format the remaining files now you've seen the code changes.

schakrava:
Thanks @phillxnet looks good. I use spaces in my setup, but I'll double check.

phillxnet:
@schakrava Cheers.

PR #1226, opened 2016-03-22
Update add_task.jst

MFlyer:
Added 5, 10, 15 and 20 minutes for snapshots/scrubs tasks.
I think it's better than just 1 minute / 30 minute

Flyer/Mirko


schakrava:
Thanks @MFlyer. This is nice!

PR #1227, opened 2016-03-22
snapshots.py max_count bad list count

MFlyer:
Example: max_count set to 25

Got 25 snaps so len(snapshots) > max_count false
Add another snap so we go to 26 (25+1)

Next time IF statement is true so we delete [max_count:] -> [25:] -> element 25 and so on
another snap, back to 26 snaps (25+1) cos list index start from 0!

### DIFF
--- a/src/rockstor/scripts/scheduled_tasks/snapshot.py
+++ b/src/rockstor/scripts/scheduled_tasks/snapshot.py
@@ -68,8 +68,8 @@ def main():
     prefix = ('%s_' % meta['prefix'])
     snapshots = Snapshot.objects.filter(share=share, snap_type=stype,
                                         name__startswith=prefix).order_by('-id')
-    if (len(snapshots) > max_count):
-        for snap in snapshots[max_count:]:
+    if (len(snapshots) >= max_count):
+        for snap in snapshots[max_count-1:]:
             url = ('shares/%s/snapshots/%s' % (meta['share'], snap.name))
             try:
                 aw.api_call(url, data=None, calltype='delete', save_error=False)
--- a/src/rockstor/scripts/scheduled_tasks/snapshot.py
+++ b/src/rockstor/scripts/scheduled_tasks/snapshot.py
@@ -66,19 +66,7 @@ def main():
     max_count = int(float(meta['max_count']))
     share = Share.objects.get(name=meta['share'])
     prefix = ('%s_' % meta['prefix'])
-    snapshots = Snapshot.objects.filter(share=share, snap_type=stype,
-                                        name__startswith=prefix).order_by('-id')
-    if (len(snapshots) >= max_count):
-        for snap in snapshots[max_count-1:]:
-            url = ('shares/%s/snapshots/%s' % (meta['share'], snap.name))
-            try:
-                aw.api_call(url, data=None, calltype='delete', save_error=False)
-                logger.debug('deleted old snapshot at %s' % url)
-            except Exception, e:
-                logger.error('Failed to delete old snapshot at %s' % url)
-                logger.exception(e)
-                return
-
+    
     now = datetime.utcnow().replace(second=0, microsecond=0, tzinfo=utc)
     t = Task(task_def=tdo, state='started', start=now)
     try:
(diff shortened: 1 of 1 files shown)


MFlyer:
manually deleted snapshot to get under my max_count, but recorsively it should aid itself to the correct number

schakrava:
I like your new/different proposal @MFlyer . In case snapshot creation fails, it will ensure that max_count of snapshots are present. But if there's a bug in delete, this will result in ever increasing number. So we could put a check before creating the snapshot to some upper bound. For example, if the current number of snapshots is max_count + 1, then delete first and only if it succeeds, create new one.

This is a nice little logic rework. Would you be interested in finishing it up?

MFlyer:
Hi @schakrava I partially agree. Thinking about colleagues messing on files i would prefer always to get a max_count+1 or more (debug on console or mail alert) rather than max_count-1, so we could choose an easier but safe solution: snapshot creation - on failure mail alert; snapshot deletion with an if statement (only if current snapshots number is greater than max_count - last created/supposed to be created snapshot substitutes the last one - this way, also assuming a failing delete task, on the next first successful delete it will reduce snapshots to desired max_count).

Recap: switch snapshot delete with snapshot create and before deleting check if snapshots collection meets our max_count ( so doing acts also as a min_count check ).
Obviously interested in finishing up, probably will ask you to edit for mail part / explain how to do it :)

M.

MFlyer:
Hi @schakrava think we've got some issues with snapshots creation / snapshots list - check last commit, it seems like snapshot creation is too slow to get caught by snapshots listing

possible solutions
go back to delete / create way, so you always get the actual (and real) snapshots list length
add a check var during snap creation and add in delete if statement to act max_count-2

MFlyer:
@schakrava 
Finally got solved, Snapshot.object collect data correctly, error was on share (issue #1230)
Ready to merge / your suggestions

schakrava:
Nicely done @MFlyer. I refactored and documented all the new changes. Do take a look the final changes and let me know if you see anything to improve further. It should be helpful reading for your future pull requests as well. 

Really glad to have you as a contributor!

MFlyer:
Hi @schakrava just seen your refactored version, but don't agree at all for deletion before creation (line 98 if(delete(aw, share, stype, prefix, max_count)): )

So doing if you get ok on delete but fail on create you are on a max_count-1 snapshots: avoid extra snapshots over max_count and go safe for disk space, but with a possible missing snapshots.

Talking about refactoring: ok, got it for future commits, I'll be more explicit in code :)

MFlyer:
@schakrava  Generally talking about refactoring and expanding scheduling snapshots, what do you think about let users decide?

Your way - Space conservative - exclude exceding data but lets snapshots incosistency (if you get error on delete you don't create snaps so users could have latest snasphot refering to n hours ago, while new data was added or modified on share)

My way - Conservative against current data state but with risk of 'over snapshots' - get always your snapshot first also if delete fails

Other thing i'd like to add: scheduling snapshots on an anacron basis ( want my snapshots every 5 mins but only during office time 9to6 )

Waiting for your opinion cos i'd like to work on these things :)

schakrava:
I am assuming you've read my refactored code and comments. I allow at most 1 extra snapshot in cases where deleting oldest snapshot fails. If deletion of snapshot fails, chances are, creation also fails. Even if it doesn't there's a high likelihood that it only creates bigger problems if ignored for longer. So, the best is to not create any new snapshots and then alert the user immediately. The alert part is missing currently and something we should get to sooner than later.

Regarding your second question about anacron style snapshots, I really like the idea. Feel free to look into it if you are interested. The backend part is easier than the UI part, I think. If you do work on it, create a separate issue just for this.

MFlyer:
> Even if it doesn't there's a high likelihood that it only creates bigger problems if ignored for longer. So, the best is to not create any new snapshots and then alert the user immediately. The alert part is missing currently and something we should get to sooner than later.

Ok, assuming an alert system fully agree on current code (that was my only real worry) ;)

> Regarding your second question about anacron style snapshots, I really like the idea. Feel free to look into it if you are interested. The backend part is easier than the UI part, I think. If you do work on it, create a separate issue just for this.

Got it #1233 Think it will be a mix on backend and some UI adds

Issue #1230, opened 2016-03-24
Snapshots deleted via shadow copies - share corrupted


MFlyer:
@schakrava @phillxnet 
As tested while working on scheduled tasks (was still using this share [Forum thread](https://forum.rockstor.com/t/shadow-copies-on-samba-over-btrfs-snapshots-2-simple-reasons-to-avoid-it/1239/4?u=flyer)), deleting also a single snapshot via shadow copies (possible security issue related to human error and ransomware too) kill all the others

I got snapshots nested inside snapshots, qgroup assign without rescan and total size / exclusive size 0 0

MFlyer:
@schakrava - Pls close this one, will open it again in the future just for docs integration and "not panicing" alert :wink:

PR #1231, opened 2016-03-24
Shadowcopies warning

MFlyer:
@schakrava 

On my own i would disable shadow copies, so at least decided to add a warning for admins enabling them #1230


schakrava:
This is good, but we need to furnish more information. Would you be interested in creating a doc in rockstor-docs repo clearly demonstrating the problem with screenshots/video and explanation? Then we can put a link to it in the warning. What we don't want is bunch of people panicing without knowing exactly what the risk is.

MFlyer:
So you asking to destroy a share (again xD) and create docu for that? Ok, I'll do it

PR #1235, opened 2016-03-30
Parse snapshot path correctly when it contains spaces

sprintman:


MFlyer:
Hi @schakrava is this one an orphan PR?!?

schakrava:
Unfortunately so. We need to change the API endpoints to use resource id's instead of names. I plan to get to it soon.

phillxnet:
@sprintman This one is now super old so closing as a general tidy up and as per prior comments.

PR #1237, opened 2016-03-31
1190 django upgrade

schakrava:


phillxnet:
@schakrava I found the comment that I thought I'd left as a todo relating to this pr:
```
storageadmin/views/disk.py:                do.delete()  # django >=1.9 returns a dict of deleted items.
```
in line 83 that I added as part of the rework in that area re serial numbers and removed drives.

schakrava:
This was re-done in #1545 and #1546

Issue #1245, opened 2016-04-06
Unsupported Compression Algorithm Error when no option is selected


priyaganti:
In the shares detail view, under the Edit compression Algorithm section, if Submit button is clicked with the default text option saying "Don't enable compression", the following error is shown. 
  
![image](https://cloud.githubusercontent.com/assets/1999710/14301905/53062576-fb50-11e5-831c-8e12df7609de.png)

MFlyer:
@priyaganti Got solved

priyaganti:
Thanks @MFlyer. I have noticed the same thing today and fixed. Should have checked your comment earlier.

schakrava:
fixed by #1248 and #1247

PR #1247, opened 2016-04-06
#1245 Added missing quotes - no more errors

MFlyer:
#1245 Just added missing double quotes, now working

Flyer

P.S. sorry for mess on PR etc etc, first time created branch under other branch :sweat:


schakrava:
Thanks @MFlyer . It's totally understandable, once you open a few of these it will become a lot more straight forward.

PR #1251, opened 2016-04-08
Issue#1198 tooltip chop

priyaganti:
This PR solves the tooltip issue at 2 places:
a) Edit Compression Algorithm option under pool details page. 
b) Nut mode under NUT-UPS service Config

Explanation:
1) Add container = 'body' to the tooltip plugin. 
2) It appends the tooltip to the body element. 
3) Since Navbar is also under the body element, tooltip is displayed over it instead of getting chopped.


schakrava:
Nicely done @priyaganti. Thanks!

Issue #560, opened 2015-01-03
Support bonded networking


schakrava:
I tried to set up bonded networking. There were several issues I ran into.

    NetworkManager conflicts with Rockstor and they both try to use different interface names and overwrite each others setting. I just uninstalled NetworkManager with yum.
    Bonded networking adds a file /sys/class/net/bonding_masters which Rockstor tries to enumurate as a network interface and errors out. In system/osi.py network_devices I changed
        devices = os.listdir('/sys/class/net')
    to
        ifpath = '/sys/class/net'
        devices = [name for name in os.listdir(ifpath) if os.path.isdir(os.path.join(ifpath, name))]
    so it only enumerates directories.
    If viewing the interface properties in Rockstor, even if no changes are made and cancel is clicked, Rockstor rewrites the ifcfg file and clobbers the bonding settings

It would be great if Rockstor could fully support bonded networking for either failover or aggregation, but at minimum it should play nicely with a manual setup.

schakrava:
More user feedback to consider

I noticed that the default for the network interfaces is DHCP. I have 2 interfaces on the system, one for internal network testing (no DHCP) and one external with DHCP. I noticed that the Network Manager spams the logs trying to configure this second interface. 
Include an option to manually disable network interfaces, if it is the only active interface have a warning before proceeding.
If the interface cannot be configured after 10 attempts, automatically disable it. This can always be reconfigured later.

roweryan:
If you can't support bonded networking yet, could you give us a way to configure the networking ourselves, so that rockstor doesn't overwrite it?

schakrava:
@roweryan You might want to look at this wiki post on the forum: http://forum.rockstor.com/t/network-management-implementation-details/441

In summary, Rockstor doesn't (and shouldn't) override your custom configuration. It would be great if you get bonding working and share the steps by way of replying to that wiki post on the forum.

schakrava:
Fixed by #1253

PR #1253, opened 2016-04-10
wip stub code for teaming support. #560

schakrava:


MFlyer:
Wow @schakrava , that's the full skeleton for teaming :+1:

schakrava:
@MFlyer thanks! I haven't pushed db migration files yet, so not really testable at this point in case you are curious. But I'll update here soon as it is. Would love to roll this one out on testing channel soon and get feedback from others.

PR #1254, opened 2016-04-10
Disable Add storage button when Rock-on is ON. Fixes #1178

### LINKED ISSUE #1178: Add Storage only possible when Rock-On not running
Hello,

currently, RockStor does allow editing and thus adding storage while a rock-on is running.
But then, when submitting the changes, an error message pops up telling that this action is only possible when the rock-on is not running.

This is annoying, as this only comes up, after the user has specified all the settings in good hope that it will run.

Proposals:
Editing config should only be possible if rockon is off
OR
Editing config should be possible if rockon is running, but the user should be informed that after the changes a restart is needed. This restart should be offered

Greetings,
Hendrik

priyaganti:
Add a alert message that storage can be added to Rock-on only when it's
turned off and then disable the add storage button to prevent any other
events from triggering.


schakrava:
Thanks @priyaganti

Issue #885, opened 2015-09-14
drive power down interface


phillxnet:
Currently it seems the only way to set / enable drive powerdown is via cli hdparm commands.

Now that Rockstor has hdparm by default we can add a facility to set this figure on each device.
Issues:-
Drive names change but that is dealt with at some level elsewhere I am just not sure how we translate that to the device parameter that hdparm takes, presumably via the same function that shows the drive name in Disks table.
Reasons:-
Power / noise save.
Avoid users having to enter eg /sbin/hdparm -S120 /dev/sdb above initrock -x entries in /etc/rc.local to enable 10 min power down. (a value of 239 sets a 20 mins power down)

Obvious place to add this interface might be an additional column or icon in Disks. This might then also equate to the current power state. I believe smart can relay the drives current power state so we could poll from that database but I think it better to have a dedicated entry / mechanism for power setting / state read as then lighter and can be polled without drawing in all of the smart data.

Options in this drive setting - set how long before drive enters sleep (no disk motor activity) state and an instant sleep mode option on each / all drives.

maxhq:
Two thoughts about that:

**1. Per-pool timeout instead of per-disk**

I wonder if setting this for each drive might lead to confusion, as the user has to check which drives belong to the same pool. Maybe it should be better set per pool (BTRFS filesystem)?
Then the corresponding drives could be fetched via `btrfs fi show -d` and timeout can be set.

**2. Non-rotational devices**
There might be problems when spindown is configured on non-rotational devices (USD stick, SD card, SSD). So that might have to be checked first.
I tried it on a USB stick and it became completely unaccessible (very old Linux though).

Currently I do the following in a systemd startup script:

```Shell
#!/bin/sh
# list all disks
lsblk -p -i -o KNAME,TYPE,SIZE,MODEL | while read dev type size model; do
	# only process disks (skip partitions etc.)
	test $type != "disk" && continue
	
	# skip non-rotational disks (indicators: rotation rate or acoustic management)
	spin_rate=`udevadm info --query=all --name=$dev | grep 'ID_ATA_ROTATION_RATE_RPM=' | cut -d= -f2`
	test -z $spin_rate && spin_rate=0
	aam=`udevadm info --query=all --name=$dev | grep 'ID_ATA_FEATURE_SET_AAM=' | cut -d= -f2`
	test "$spin_rate" == "0" -a "x$aam" != "x1" && continue

[...]

PR #1260, opened 2016-04-12
drive power down interface. Issue #885

phillxnet:
Opening this pull request to aid in collaboration.

The last issue I'm having problems with in this pr is the passing of an additional value 'spindown_message' from:

storageadmin/static/storageadmin/js/templates/disk/spindown_disks.jst
and it's associated js in
storageadmin/static/storageadmin/js/views/spindown_disks.js

The value is the index in the js object "spindownTimes" in spindown_disks.js that the 'display_spindown_time' handlebar helper in the same file uses to populate the form drop down named spindown_time.

So I want to pass as an additional value the text (key in js object) used to populate that drop down.
The selected value is already passed successfully I'm just stuck on adding the text of the selected key as an additional element.

The receiving  method in:
storageadmin/views/disk.py
def _spindown_drive(cls, dname, request):
has in place a request.data.get('spindown_message', 'message issue!'))
and passes this on to:
set_disk_spindown(disk.name, spindown_time, spindown_message)

So stuck on presenting the js object key as spindown_message.

Everything else is currently working as intended.

Sorry if this is a silly one it's just the last hurdle and I'm drawing a blank.

### DIFF
--- a/conf/rockstor-hdparm.service
+++ b/conf/rockstor-hdparm.service
@@ -1,0 +1,9 @@
+Description=Rockstor hdparm settings
+After=suspend.target
+
+[Service]
+Type=oneshot
+ExecStart=
+
+[Install]
+WantedBy=suspend.target basic.target
--- a/conf/rockstor-hdparm.service
+++ b/conf/rockstor-hdparm.service
@@ -1,9 +1,13 @@
+[Unit]
 Description=Rockstor hdparm settings
 After=suspend.target
+# Rockstor-hdparm - the automated maintenance of this file requires a clear line
+# at the end of the Unit and Service sections, but no other clear lines.
+# Also requires at least one ExecStart entry in Service section with a comment
+# directly after each ExecStart entry.
 
 [Service]
 Type=oneshot
-ExecStart=
 
 [Install]
 WantedBy=suspend.target basic.target
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -46,7 +46,6 @@ DEFAULT_MNT_DIR = '/mnt2/'
 RMDIR = '/bin/rmdir'
 WIPEFS = '/usr/sbin/wipefs'
 QID = '2015'
-HDPARM = '/usr/sbin/hdparm'
 
 
 Disk = collections.namedtuple('Disk',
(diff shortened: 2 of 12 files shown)


schakrava:
I reviewed briefly and your changes are looking great @phillxnet!

One way to pass the key is to assign spindownTimes to _this to make it available inside the submitHandler.
`
_this.spindownTime = spindownTimes
`  
After you get the json from the dom 
`_this.$('#add-spindown-disk-form').getJSON()`
you can lookup the key of corresponding spindowntime value the user selected, which is in that JSON and update the JSON by adding a new key, value pair with message. Hope that helps.

Btw, should this feature apply to non rotational media at all? I am guessing not. If so, does your code detect that and disable it automatically?

phillxnet:
@schakrava Thanks, that was just the job and worked perfectly.
I've committed it in https://github.com/rockstor/rockstor-core/pull/1260/commits/c70980dbf6a6528c83bd8b8d5fb936a15546a3ce

Another quick question at your leasure:-
How do I set the background colours of the d3 slider (screen shot):
slider-background-colours-not-showing

The ranges are indicated in the key below the slider and can be hardwired but can't see where to do it.
Thanks and sorry if this is another silly one.

phillxnet:
@schakrava Don't worry about my last background colour question; I've made some headway via:-
d3.slider2().......reclaimable(127).used(1)
So I'll keep chipping away on this one and let you know when it's ready for review.

phillxnet:
Current progress with Disks table for drive Spindown and APM status (screenshot):
disk-spindown-time-and-power-in-table

phillxnet:
- [ ] Set slider to reflect current APM value when loaded
- [ ] add text box input validation (ie only numbers from 0 - 255)
- [ ] Update slider on text box input.

phillxnet:
- [ ] Resolve false positive on TS32GMSA370 for is_rotational - an msata with acoustic management !!! Which I have used as an indicator of rotational nature.
- [ ] Final UI tweak to deal with showing but not setting APM values unless specifically enabled. Involves configuring the "Enable APM settings" tickbox and possibly intervening on values represented by the associated controls.

phillxnet:
@schakrava Almost there now, just need to be able to disable the d3.slider2 ie
this.$('#slider').attr('disabled', 'true');
doesn't do it.
But the same works on the associated text box, ie refuses to take external input.
I need it to present it's value but not allow input, ie no longer respond to drags. Don't see a call / setting for this though.
Cheers.

phillxnet:
This is now mostly functional except for an input validation bug caused by contention between the text box updating the slider first rather than triggering the validation routine. Will have another look shortly.
Quick screenshot of the config screen as is:-
spindown-config-page

phillxnet:
I have re-based this branch on the just released 3.8-13 master branch and to avoid replication of commits I deleted the github remote branch which closed this pr. I will open a new pull request shortly based on the new branch.

PR #1266, opened 2016-04-15
advise_nmtui

ScarabMonkey:
Advises the user to use nmtui as root to fix networking issues at first boot.


ScarabMonkey:
OOps, something went wrong with my github app ... please ignore...

PR #1268, opened 2016-04-15
Advise nmtui

ScarabMonkey:
Advise the user to use nmtui to fix network configuration at first boot.


schakrava:
Thanks @ScarabMonkey, really appreciate your contribution!

PR #1269, opened 2016-04-16
Power drop down menu clearly visible. Fixes #1192

### LINKED ISSUE #1192: Power menu not right aligned
As you can see the drop-down menu is cropped if you use short user names.

rockstor

priyaganti:


priyaganti:
Fixes #1192

schakrava:
Thanks @priyaganti !

PR #1274, opened 2016-04-19
Issue#1196 select multiple disk resize

### LINKED ISSUE #1196: Select mutiple disks easily during resize
Add/remove disk should have a checkbox in the column header which functions as a select all for available disks. Just like when creating a new pool.

priyaganti:
Fixes #1196


schakrava:
Thanks @priyaganti, changes look nice.

PR #1280, opened 2016-04-20
Issue#1202 rockon config redirection

priyaganti:


schakrava:
Thanks @priyaganti , nicely done.

I consolidated CustomChoice and Environment views since they are very similar, fyi.

Issue #1284, opened 2016-04-22
Active Directory info popup slightly misleading


ScarabMonkey:
It mentions the enumerate configuration option - which isn't an option.

ScarabMonkey:
I have submitted a pull request for it :)

Though the file I had to modify seemed like an odd place for this info - perhaps there would be a more generic way for services to put an info thingy onto the services page?

phillxnet:
Linking to the said pull request for ease #1285

schakrava:
You have a good point @ScarabMonkey . @priyaganti is looking into that. It will also make translations easy, we are shamefully behind on that front.

Thanks for the pull request, btw!.

Fixed by #1285

PR #1285, opened 2016-04-22
Issue #1284 - Remove extraneous info

ScarabMonkey:
We always enumerate now.


phillxnet:
@ScarabMonkey Nice catch.
Do you think we need a 'the' in there as well while you are at it.

ScarabMonkey:
Changed to "the Active Directory"

schakrava:
Thank you @ScarabMonkey !

PR #1287, opened 2016-04-22
Issue#1275 handlebars part2

priyaganti:

Issue #1275, opened 2016-04-19
Cleanup handlebar helpers and move to .jst files (handlebars-part2)


priyaganti:

phillxnet:
@priyaganti I've added a couple of columns to the disks table in the referenced issue so just linking here to remind me to keep an eye on your changes so that I might try and ease merging later.

MFlyer:
@priyaganti - @schakrava this to be closed after last PR merged

schakrava:
Fixed by #1287

PR #1292, opened 2016-04-25
drive power down interface. Fixes #885

### LINKED ISSUE #885: drive power down interface

phillxnet:
Adds a per drive power down interface / feature based on idle time using the standard hdparm parameters. Given this is closely related to a drive's Advanced Power Management (APM) setting a display of the current setting and an option to change this setting is also included in the disks table and on the same configuration page. Care has been taken to graphically associate the APM setting with the Power Status (spin down) setting by grouping these columns one after the other and maintaining this relationship on the settings page also.

Hdparm is itself an interface to the kernel's own drive control subsystem so we are sticking to core linux utilities for this feature.

The hdparm switches used are -C to set idle spin down time and -B to read and set APM level.
N.B. It is not possible to read a drives current setting for -C (idle spin down) which complicates matters, the meaning for these values can also vary between drives. The settings used are drawn from man hdparm and are apparently more reliable for newer drives.

Hdparm settings are maintained over a reboot but not over a power cycle. To address this a new stand alone rockstor systemd service is introduced (rockstor-hdparm.service). It does not depend on any other rockstor systemd service and no other rockstor service depends on it. The service is used simply to execute the tested hdparm commands that are otherwise executed on demand via the WebUI. That is if no error message and no non zero return code is received from a proposed (via user config entry) hdparm command then that same command is placed in this systemd service unit to be applied on the next boot to address the power cycle loss of these settings otherwise.

As idle time / spin down is associated with rotational media we do not apply any setting to a device if it fails an included is_rotational test. This test is based on a drives udevadm info readings for rpm and Automatic Acoustic Management (AAM) readings. The kernels own proc reading of rotation has been frequently reported as currently unreliable hence establishing our own mechanism for this. This function (in system/osi) is considered as conservative and will err on the "not confirmed as rotational" side. If the is_rotational function returns false no setting is applied and user confirmation of this is simply to not update our disk table with any requested settings. This could be improved upon but may benefit from future enhancements in a notification system. As is the following info level message is logged:-
"Skipping hdparm settings: device not confirmed as rotational"
This should soon be more readily available by improvement in WebUI log view functionality.

The core of this feature's function is several light weight property extensions to the disk model that provide the power status, hdparm -C setting, and current APM level to the disks page by way of the additional 2 columns. There is also included a pause function that simply executes hdparm -y to request a drive immediately enters standby mode. No full sleep mode is activated by the included mechanisms (ie hdparm -Y) as such a state requires that a drives interface be re-set and was considered overly heavy handed and potentially more problematic. The active settings change path takes the users input and tests it for sanity then attempt to run the proposed hdparm command one switch at a time; starting with the -B setting as this is more often problematic due to some drives not supporting this APM level set option. If any errors or non zero return codes are received then this setting is not included in the final systemd entry. The -C switch is then also tested in isolation and again if any errors or non zero return codes are encountered then at this stage no entry will be submitted to the systemd file. This means in some circumstances we may have only -C settings for a device but if the -B setting passed our tests then it is by default included; often simply re-affirming the default setting unless the user has specified otherwise. The core method to update the systemd service is implemented in update_hdparm_service which is utilized only by set_disk_spindown that tests the viability of the proposed hdparm switches.

To facilitate the option to not include a -B setting we use the 0 value for that switch. This is a symmetrical arrangement in that if an hdparm -B reading fails or returns 'not supported' the reading mechanism conflates these responses into the 0 value (see get_disk_APM_level). This is otherwise an unused APM value (ie 1-255) and so we can maintain a simple interger for this value. This way we avoid applying any -B option to a drive that has otherwise failed to cleanly respond to our -B request or has done so but with a 'not supported' response. Like wise the mechanism is employed to remove any existing -B entry by passing zero as the requested value. The same reading mechanism translates the off reading to 255 which requires our interface logic to translate this back to 'off' but this was considered a cleaner system than dealing with the raw values of 0 (our flag) or 1-255 (written values) or 'off' which represents a reading of the the written value 255. This way we maintain symmetry with reading (using our low level reader) and writing and only translate in the UI for readability. On the UI front the APM settings are disabled by way of their tickbox option and the tick box is also disabled when an APM read of 0 (error or not supported) is encountered.

Note that all hdparm commands use the dev by-id names to maintain correct targets over system reboots. Several general purpose mechanism have been included in this pr to facilitate translating the currently db drive.names of sda / sdb to these more stable by-id drive names which are created by the udev subsystem and represent simple symbolic links to their associated sda / sdb type names. It is intended that these translation functions be employed on a wider basis going forward.

By default we offer a 20 minute idle spin down time if a prior setting is not found. Given prior settings cannot be established using system calls we establish the single point of truth to be direct reading from the device and the systemd file where reading from the device is not possible (ie with -C settings). This avoids using the django db and was intended to keep the whole hdparm feature as light as possible. If the systemd file does not exist it is created from the empty template held in /opt/rockstor/conf/rockstor-hdparm.service. When newly established the service is also enabled but not started as the associated commands have already been executed by way of the tests prior to their entry in the systemd file. The service being enabled is sufficient to ensure that they will be run on next boot which is all that is needed. If on the other hand no entries are found in an existing /etc/systemd/system/rockstor-hdparm.service file, ie by way of UI instigated edits, then the service is disabled and the service file removed in order to prepare for new settings and the re enabling of the service in that instance.

In light of issue #1275 and it's associated pr #1287 which re-factors the disks table creation code modified in this pr I have prepared some handlebar helpers and included them in this pull request, they are however untested as they rely on what is as yet un-merged UI code.

Testing of the supplied spin-down values has been mostly successful with values of 20 minutes and less as higher values are less reliable to non functional. It was also found that with no WebUI open the drives are much more reliably suspended, however opening the WebUI and viewing the disks page after drives have entered standy mode and even refreshing this view via browser or via use of the "Rescan" button doesn't then wake the drives that have entered standby. So there appears to be some command active that inhibit the drives entering standby (while the WebUI is open) but doesn't then activate these drives once they are in standby mode. This was a hard one find. Also note that even with the WebUI open multiple drives did successfully enter standby when set to 30 seconds or 1 minute but no longer. However 20 mins and below seem to work find when the WebUI isn't active. This has mostly been tested when viewing the disks page and is quite a time consuming business. Hence the submission of this pr at the beginning of a testing cycle.

Reference in the user interface is also made to the recently merged "Task execution time windows" feature.

Note: This pr submission text is also intended as the basis of a wiki entry upon successful review and merge. The wiki entry can then serve as a technical manual on this subsystem.

### DIFF
--- a/conf/rockstor-hdparm.service
+++ b/conf/rockstor-hdparm.service
@@ -1,0 +1,9 @@
+Description=Rockstor hdparm settings
+After=suspend.target
+
+[Service]
+Type=oneshot
+ExecStart=
+
+[Install]
+WantedBy=suspend.target basic.target
--- a/conf/rockstor-hdparm.service
+++ b/conf/rockstor-hdparm.service
@@ -3,8 +3,11 @@ Description=Rockstor hdparm settings
 After=suspend.target
 # Rockstor-hdparm - the automated maintenance of this file requires a clear line
 # at the end of both the Unit and Service sections, but no other clear lines.
-# It should also be exactly 13 lines long including these comments but not
-# including the ExecStart lines and their obligatory following one line comment.
+# The line count of the template file in <rockstorroot>/conf is used to evaluate
+# if an 'all entries removed' scenario has been reached. If any modification are
+# made to the template file please also delete the generated file,
+# /etc/systemd/system/rockstor-hdparm.service, and recreate any existing
+# configuration via Rockstor's WebUI.
 
 [Service]
 Type=oneshot
--- a/conf/rockstor-hdparm.service
+++ b/conf/rockstor-hdparm.service
@@ -1,9 +1,13 @@
+[Unit]
 Description=Rockstor hdparm settings
 After=suspend.target
+# Rockstor-hdparm - the automated maintenance of this file requires a clear line
+# at the end of the Unit and Service sections, but no other clear lines.
+# Also requires at least one ExecStart entry in Service section with a comment
+# directly after each ExecStart entry.
 
 [Service]
 Type=oneshot
-ExecStart=
 
 [Install]
 WantedBy=suspend.target basic.target
--- a/conf/rockstor-hdparm.service
+++ b/conf/rockstor-hdparm.service
@@ -3,8 +3,8 @@ Description=Rockstor hdparm settings
 After=suspend.target
 # Rockstor-hdparm - the automated maintenance of this file requires a clear line
 # at the end of the Unit and Service sections, but no other clear lines.
-# Also requires at least one ExecStart entry in Service section with a comment
-# directly after each ExecStart entry.
+# It should also be exactly 13 lines long including these comments but not
+# including the ExecStart lines and their obligatory following one line comment.
 
 [Service]
 Type=oneshot
--- a/conf/rockstor-hdparm.service
+++ b/conf/rockstor-hdparm.service
@@ -2,7 +2,7 @@
 Description=Rockstor hdparm settings
 After=suspend.target
 # Rockstor-hdparm - the automated maintenance of this file requires a clear line
-# at the end of the Unit and Service sections, but no other clear lines.
+# at the end of both the Unit and Service sections, but no other clear lines.
 # It should also be exactly 13 lines long including these comments but not
 # including the ExecStart lines and their obligatory following one line comment.
 
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -1115,26 +1115,26 @@ def get_disk_power_status(disk):
     any request will wake a fully sleeping drive.
     Drives in 'sleeping' mode typically require a hard or soft reset before
     becoming available for use, the kernel does this automatically however.
-    :param device: disk name eg sda.
+    :param device: disk name as stored in db / Disk model.
     :return: single word sting of state as indicated by hdparm -C /dev/<disk>
     and if we encounter an error line in the output we return unknown.
     """
     # if we use the -C -q switches then we have only one line of output:
     # hdparm -C -q /dev/sda
     # drive state is:  active/idle
-    # Note however that if there is an bad/missing sense data then it will be
-    # on the first line but then we can't trust what follows anyway then.
-    hdparm_command = [HDPARM, '-C', '-q' '/dev/%s' % disk]
-    # todo consider throw=False for production
-    out, err, rc = run_command(hdparm_command)
-    # should be only one line but trawl anyway.
-    # todo once tested loose the for loop.
-    for line in out:
-        fields = line.split()
-        # our line of interest as 4 fields when split by spaces, see above.
+    # in some instances an error can be returned even with rc=0, we ignore this
+    # ie SG_IO: bad/missing sense data, sb[]:  70 00 05 00 00 00 00 0a ......
+    # We may want to ignore / return unknown when len(err) != 1
+    out, err, rc = run_command([HDPARM, '-C', '-q', '/dev/%s' % disk],
+                               throw=False)
+    # if len(err) != 1:
+    #     logger.debug('hdparm has output to standard error of %s', err)
+    if len(out) > 0:
+        fields = out[0].split()
+        # our line of interest has 4 fields when split by spaces, see above.
         if (len(fields) == 4):
             return fields[3]
-    return 'Unknown'
+    return 'unknown'
 
 
 def blink_disk(disk, total_exec, read, sleep):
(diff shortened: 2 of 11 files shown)


phillxnet:
Please note that this pr moved some non btrfs functions from fs/btrfs to system/osi.
This may have some unforeseen repercussions on building / packaging.

maxhq:
Hi @phillxnet,
I really like your great effort regarding the power down functionality!
After a quick look through the code I stumbled upon two lines where you hardcode the expected line count of the systemd template - I see it was late friday night ;-).
https://github.com/phillxnet/rockstor-core/blob/885_drive_power_down_interface/src/rockstor/system/osi.py#L985
https://github.com/phillxnet/rockstor-core/blob/885_drive_power_down_interface/src/rockstor/system/osi.py#L1003
I think this is likely to cause headaches in the future and you should open the template and count the lines. Or better: grep for `ExecStart` as this seems to be the trigger according to your code comments (by the way: great comments!).

Regards,
Jens

phillxnet:
@maxhq Thanks for the review, much appreciated; and yes I did feel a little guilty about this one. But I had assumed nobody was watching ;-).
More seriously my thought on this at the time was that given we control the template and that the template states in comments there is an expected length of 13 (to provide this very check as it goes) then I though we were done. But as you say there is a little fragility there going forward given we can't even add more comments so I agree a little test first and proceed may help. Also we can unite the two '13' checks which should have been done anyway. It does mean we add another file read and line count on every edit but this is not a frequent thing and the file is tiny as well so yes.

As to working directly from ExecStart I actually started off doing it that way and having an empty ExecStart within the template but ended up preferring a blank (no ExecStart) template as this made more sense to me and cleaned thing up a tad. It also allowed for this easy check of line counts rather than using the expensive and slow 're' pattern matching just to see if we had an empty file, ie all configs removed by use request. I tried initially to embed the 'back to template' into the main parsing code loop but as indicated in the code comments I thought this was complicated enough already so sided with the 2 step approach to try and keep the code a little easier to maintain.

I will also as per your comments check my own in-code comments re ExecStart expectation in the template.

@schakrava On review from @maxhq I plan to add an additional template line count and unify the resulting value in the two conditionals that reference it.

@maxhq I meant to say, as you now see, that I didn't manage to get your great idea of per pool power down in as this turned out to be a step too far to get this pr ready for review in time before the rest of Rockstor moved sufficiently far on to make merging it more problematic. As is there is already a UI element that will need to be hand merged and re-tested. But I think if this works out and gets in we can pop that into another nice to have feature request issue. Thank again for looking over this and your generous comments. Many eyes and all.

phillxnet:
v2 changelog: (last 3 commits)
As suggested by @maxhq I have added a template line count and unified this value's use.
A check has now been placed on the existence of this template file (prior todo).
All tested to behave as expected, including deleting the template file to test for graceful failure and relevant logging.

@schakrava v2 ready for review.

schakrava:
@phillxnet I am loving the comments and thorough implementation over all. It's very considerate of you to hand merge #1285, making my task a lot easier.

schakrava:
Also consider creating a separate file for specific group of functions. eg: hdparm.py or udev.py etc.. Just a random thought as you are improving the helper layout.

schakrava:
Thanks for all your work on this @phillxnet, happy to have merged it finally.

Thanks for pitching in @maxhq

phillxnet:
@schakrava Cheers and thanks for you help on getting this one done. @maxhq Thanks from me also.

maxhq:
@schakrava @phillxnet It's very pleasant to work with enthusiastic people like you! I love RockStor for your friendly attitudes, the use of BTRFS, the great UI and the way you use existing tools and functionality of Linux. Thank you!

Issue #1225, opened 2016-03-22
Minor issue on web ui - "Total capacity, allocation and usage" widget not resizing


MFlyer:
As for my title, here are 2 screenshots:

**Widget enlarged don't fit container size**
![widget enlarged](https://cloud.githubusercontent.com/assets/17988902/13952749/802d5d30-f038-11e5-8825-68a4dbd43232.jpg)

Dashboard reloaded widget ok on large format
Widget reduced overfit div
![over](https://cloud.githubusercontent.com/assets/17988902/13952796/c161fbee-f038-11e5-9e6d-e01f3fe71a60.jpg)

All others seem ok
This is a really really minor issue, i know it

phillxnet:
@MFlyer Thanks for reporting, minor or not. Nice find.
Just reproduced here on Chrome Version 49.0.2623.87 and Firefox 45.0 in Fedora 23.

Total Capacity widget in top left, expand (widget contents doesn't rescale), refresh browser and widget contents refreshes to large then reduce widget again and as you found widget contents are not rescaling to the new smaller widget size either.

So looks like this particular widget's contents are not auto resizing like the others do.
Good to know.

MFlyer:
More info about this bug, found where it fails:

Uncaught TypeError: this.setGraphDimensions is not a function

schakrava:
Fixed by #1293

PR #1293, opened 2016-04-26
#1225 Missing setGraphDimensions

MFlyer:
#1225 PR workaround (manually call 3 funcs to obtain resize functionality)


schakrava:
Thanks @MFlyer . Nice fix, really appreciate it. I just made a small cleanup/followup commit. do take a look.

Issue #1296, opened 2016-04-27
Snapshots fail after updating to 3.8-13.04


maxhq:
Since the update I get the following error messages from `/opt/rockstor/bin/st-snapshot XX`:
```
Traceback (most recent call last):
  File "/opt/rockstor/bin/st-snapshot", line 43, in <module>
    sys.exit(scripts.scheduled_tasks.snapshot.main())
  File "/opt/rockstor/src/rockstor/scripts/scheduled_tasks/snapshot.py", line 77, in main
    cwindow = sys.argv[2]
IndexError: list index out of range
```

MFlyer:
Hi @maxhq, that sounds strange.

On the fly workaround that could solve: edit your task and just submit it again

Please let me know if this solve it

Flyer

ScarabMonkey:
I can confirm: editing each task and re-submitting does indeed fix this.

MFlyer:
Thanks @ScarabMonkey for confirmation :blush: and sorry again for the missing check (while testing i didn't got that probably because snapshots started after i had already save tasks with new format ! )

maxhq:
Sorry but this still does not work for me (RockStor 3.8-13.06).
I opened and saved all scheduled snapshot tasks but now I get the following errors from cron:
```
Traceback (most recent call last):
  File "/opt/rockstor/bin/st-snapshot", line 44, in <module>
    sys.exit(scripts.scheduled_tasks.snapshot.main())
  File "/opt/rockstor/src/rockstor/scripts/scheduled_tasks/snapshot.py", line 78, in main
    if (crontabwindow.crontab_range(cwindow)): #Performance note: immediately check task execution time/day window range to avoid other calls
  File "/opt/rockstor/src/rockstor/scripts/scheduled_tasks/crontabwindow.py", line 33, in crontab_range
    hour_start = int(range_windows[0]) if range_windows[0] != '*' else 0
ValueError: invalid literal for int() with base 10: 'hp'
```

schakrava:
pr is merged but it did not ship in [.06](https://forum.rockstor.com/t/dev-log-for-3-8-14/1412/7). Planned to be part of next one coming later today.

MFlyer:
Hi @maxhq that's a really strange one.

What does that line?? If value not equal to * then it has to be 0-x number (got via a split() so we convert it to int()).
So we have 2 options: value is * or a number (string format converted to number).

I don't understand why you get int trying to convert 'hp' ?!?
[...]

PR #1297, opened 2016-04-27
#1296 Add control to pool_scrub and snapshot to avoid missing cwindow…

MFlyer:
To @schakrava  #1296 Patch to avoid errors on missing sys.argvs - set execution window to default "always"

To users: Before patch merged and avail on rockstor release please edit your task and just submit again (that will auto generate missing arguments)


schakrava:
Thanks for fixing this regression @MFlyer, really appreciate you owning it!

MFlyer:
Hi @schakrava , it's always a pleasure and sorry again for that missing check!! :smiley:

PR #1298, opened 2016-04-27
#1301 AD tweak: less checking on NTP

ScarabMonkey:
Only check for NTP service when enabling AD - not before being allowed to configure it.
Unlike some other services, AD is only enabled when you choose to enable it - rather than automatically after configuring it.


ScarabMonkey:
referencing https://github.com/rockstor/rockstor-core/issues/1301

schakrava:
Thanks @ScarabMonkey . This is very reasonable. Appreciate your contribution!

schakrava:
Fixes #1301

PR #1307, opened 2016-05-07
Handlebars part4 to fix scheduled tasks 

priyaganti:


schakrava:
Thanks @priyaganti , nice work. Looks a lot cleaner now :)
I see that you've added some html coolness to @phillxnet 's recent power management screens. They look nice to me. phil, please take a look.

phillxnet:
@priyaganti Thanks for cleaning up my rather messy disk spin down and custom smart options pages. They are now much improved. I've screen captured before and after as a reminder and will endeavour to do better next time (by just copying your example I expect). Thanks.
Before:
spin-down-before
after:
spin-down-after
And custom smart before:
smart-options-before
after:
smart-options-after

Also great work on all these 'under the hood' clean ups. Makes things much easier to work out and work on. Cheers.

PR #1308, opened 2016-05-07
issue#1279 display right Transferred rate column in replication history table

priyaganti:


schakrava:
Nicely done @priyaganti. Thank you!

PR #1309, opened 2016-05-07
remote rock-on info not updating deprecated volumes. Fixes #1294

### LINKED ISSUE #1294: remote rock-on info not updating deprecated volumes
A system that had previously retrieved the older version of the plex rock-on ~~had plex installed and then un-installed~~ still shows the older plex config volume options, ie transcode share is requested. This share is not present in the new plex.json rock-on definition. This persists after deleting and recreating the rock-ons-root and repeated used of the Update button / browser refresh and rebooting. The system still prompts for the older plex ~~info, ie the~~ transcode share.
I have confirmed that the url is correct and that the new definitions are available via the generated url's.
```
[26/Apr/2016 17:08:47] DEBUG [storageadmin.views.rockon:361] we are looking to retrieve the following url =http://rockstor.com/rockons/plex.json
[26/Apr/2016 17:08:48] DEBUG [storageadmin.views.rockon:365] this is the json retrieved <Response [200]>
```
This was observed in a system updated to 3.8-13.03.

phillxnet:
In the case of our first deprecated volume definition within a rock-on we are not successfully removing the volumes requirement locally, ie the plex rock-on was updated to require one less volume; however that deprecated volume (mapped to /transcode in the container) was still used as a prompt during Plex installs or re-installs on all systems that booted or performed a rock-on metadata update prior to April 13th and this commit:-
https://github.com/rockstor/rockon-registry/commit/dca3235dd672e63af2b0f45157fe44b61a3e0b9e
where the rock-on definition file was updated.

This pull request address this by removing all prior knowledge of volume requirements during an update but only when the associated rock-on is not installed and when the remembered volume requirements exceed in number the retrieved volume count from the rockon-registry.

I have tested this pr on a clean build and on multiple systems that booted or performed a rockon metadata update prior to the above referenced volume deprecation within the Plex rock-on. No default or additional storage volume info was affected on installed plex rock-ons but upon that rock-on being un-installed both the deprecated and the additional storage mappings were removed as intended and updated to the registry definition of volumes requirement. This is akin to prior patch behaviour except that the deprecated transcoding volume requirement was also removed. There after no prompt was issued for the transcoding volume during plex install. Existing installs that used the transcoding volume mapping were un-affected by updates after this patch, at least until that Plex Rock-on install was un-installed.

In addition to the logging included within this pr (and removed prior to pr submission) the following sql command was used to verify the db's knowledge of the required volumes:
```
select storageadmin_dvolume.container_id, storageadmin_dvolume.dest_dir, storageadmin_dcontainer.name from storageadmin_dvolume join storageadmin_dcontainer on storageadmin_dcontainer.dimage_id = storageadmin_dvolume.container_id where storageadmin_dcontainer.name like '%plex%';
```

@schakrava Pull request ready for review.
Fixes #1294

### DIFF
--- a/src/rockstor/storageadmin/views/rockon.py
+++ b/src/rockstor/storageadmin/views/rockon.py
@@ -246,29 +246,54 @@ class RockOnView(rfc.GenericView):
             v_d = c_d.get('volumes', {})
             cur_vols = [vo.dest_dir for vo in
                         DVolume.objects.filter(container=co)]
+            logger.debug('dealing with updating %s', co.name)
             logger.debug('container description volumes = %s', v_d.keys())
             logger.debug('cur_vols _create_update_meta = %s', cur_vols)
-            #cur_vols can have entries not in the config for Shares mapped post
-            #install.
+            # cur_vols can have entries not in the config for Shares mapped post
+            # install.
+            # If we have more volumes defined in the rock-on definition than
+            # we have previously seen for this rockon, ie volumes added in newer
+            # definition, then remove our existing volumes record.
             if (len(set(v_d.keys()) - set(cur_vols)) != 0):
+                # but only if the current state is 'available' (to install) or
+                # 'install failed', otherwise raise warning about changing an
+                # installed rock-ons.
                 if (ro.state not in ('available', 'install_failed')):
                     e_msg = ('Cannot add/remove volume definitions of the '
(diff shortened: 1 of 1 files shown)


schakrava:
Nicely done @phillxnet! helpful comments too, as usual. Thanks!

phillxnet:
@schakrava Cheers.

PR #1310, opened 2016-05-08
#1296 Edit to sys.argv len, to be > 2 and not 1

MFlyer:
Not strictly related to #1296 but found while testing.

@schakrava sorry for this - inattention. Confirmed anyway that just editing and submitting a task autofix adding a run always

Flyer


schakrava:
Thank you @MFlyer. I should've caught this during review ;)

Issue #1302, opened 2016-04-29
3.8-13.04 update caused network widget on dashboard to not work


barnhill:
Widget used to show network traffic does not work after the 3.8-13.04 upgrade

MFlyer:
Hi @barnhill can you post a screenshot and/or browser console logs??

thanks

phillxnet:
@barnhill Thanks for reporting this; however I think this is a duplicate of #1289 where we already have a little investigation and a more accurate starting point for the occurrence of the issue.
@MFlyer Do you agree? In which case then maybe one or the other should be closed and all info copied into the remaining issue, ie version numbers / findings etc. A small and hopefully relevant console log excerpt is available in the referenced issue.

MFlyer:
Hi @phillxnet , totally agree with you - didn't remember about that issue.

@barnhill can you please close this issue and referer to the other one (as told by @phillxnet on the other side we already have more infos to start with)

Flyer

barnhill:
I had the previous version and it worked just fine.  Any ideas on how that would be possible?

phillxnet:
@barnhill As far as I was aware this network widget broken issue started at 3.8-12.13 hence my suspicion this was a duplicate. Are you saying all was well with your install on 3.8-13.03 (the previous version to 3.8-13.04? When ever you update it will jump to what ever is the newest version in the updates channel you are subscribed to. Could you have jumped a few versions?

I have so far assumed that the widget was broken by #1253 as directly after that pr was when I first noticed this and opened the referenced issue.

MFlyer:
Hi all @phillxnet and @barnhill , done some checking over network widget and found why actually not working:

On #1253 @schakrava added storageadmin.networkconnection over storageadmin.networkdevice (every connection is done over 1 or more devices - teaming/bonding or single device connection, as it was before update), meanwhile network widget still collects initial data from smart_manager_netstat that actually is empty.
Widget view js file has first data grabbing over func `getInitialData` that ask for `/api/sm/sprobes/netstat/?format=json&page_size=1` giving empty results array, so while updating data (every sec) js pushes to dataBuffers being empty/without references for interfaces.

To @schakrava : if you can handle .py side of netstat class i can work on view side :smile: (some errors on getData too -> `this.t2 = new Date(data.results[0].ts)` has to be `this.t2 = new Date(data.results[0]['ts'])` because results returns  an array and each element contains a dictionary, so you have to reference by key)

Flyer

[...]

Issue #1289, opened 2016-04-23
Broken network widget in dashboard (3.8-12.13)


phillxnet:
The Network dashboard widget is now non functional in 3.8-12.13 with repeated errors of:-
"storageadmin.js:18826 Uncaught TypeError: Cannot read property 'push' of undefined"
```
18825   _.each(data.results, function(d) {
18826      _this.dataBuffers[d.device].push(d);
18827    });
```

phillxnet:
Linked to possible duplicate #1302.

MFlyer:
To @schakrava : issue to be closed like 1302 via PR #1323  :wink:

phillxnet:
@MFlyer Sorry forgot to close this issue, it was hanging around as a duplicate origin and although older than it's duplicate it got less attention. Thanks for diligently labelling your commits as relevant here. 

Closing as fix committed by @MFlyer and merged by @schakrava via pr #1323

Issue #998, opened 2015-11-03
Dashboard crashes if left open for a long time


schakrava:
Behavior reported on both firefox and chrome browsers.

Not sure if it only happens on dashboard or other screens. It could be that we are not purging some data of dashboard widgets and as it grows over time, browser kills the tab -- just a guess at this point.

http://forum.rockstor.com/t/dashboard-crashes-chrome-tab/640

MFlyer:
Hi @schakrava, i was with a Rockstor Dashboard window open for a long time and I realized that because of cpu fan screaming, so i decided to have some tests:

Removing line charts or having just 1 out of 3 seems to solve and after some documenting and more testing i suppose it's because of d3 behaves on svg.

Reading around I found we should use instead an html5 canvas solution ( [canvas.js](http://canvasjs.com/docs/charts/chart-types/html5-line-chart/) is a good/easy one ), that better perform over chrome/safari/firefox (read some benchmarks having better frame per second with canvas, in the order of 2:1)

Flyer

MFlyer:
Hi all,
working on Dashboard widgets probably I discovered why we have a "client cpu eater":

- CPU widget runs over jQuery Flot for barchart (canvas) and d3.js (svg) for linechart (tons of updates drawing chart line & consuming cpu)
- Net stats widgets with a line chart with multiple lines (always d3.js svg) and this probably is our bottleneck (more than cpu widget)

Actually coding to chart.js over cpu widget (both barchart and linechart), next step is for net widget and others

To @schakrava : can I remove jQuery Flot or used elsewhere in rockstor??

Flyer

MFlyer:
Hi all @schakrava @phillxnet and @priyaganti ,
happy to announce that with 66be6c9 had a nice working dashboard left open 2 hours (CPU, Net and Memory Widgets) without any browser crash or freeze :sparkles: 

Working on Memory widget to complete it, than moving to last 4 widgets

Flyer/Mirko

MFlyer:
Some news
Finished Top Shares widget update on #1439 and all is fine.
Going on with Pool Capacity and Usage Widget (Top Shares clone)

Flyer

PR #1311, opened 2016-05-09
#1302 #1289 

MFlyer:
To @schakrava - As explained https://github.com/rockstor/rockstor-core/issues/1302#issuecomment-217856840 actually NetworkConnectionCollection doesn't point to devices so it's an empty list @phillxnet can you try this mod on a system with 2 interfaces over bond/team? :smiley: 
#1302 #1289

### DIFF
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -1,3 +1,21 @@
+"""
+Copyright (c) 2012-2013 RockStor, Inc. <http://rockstor.com>
+This file is part of RockStor.
+
+RockStor is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published
+by the Free Software Foundation; either version 2 of the License,
+or (at your option) any later version.
+
+RockStor is distributed in the hope that it will be useful, but
+WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+General Public License for more details.
+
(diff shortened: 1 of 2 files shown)


phillxnet:
@MFlyer  Great. Currently I'm involved in another issue but after that and a bit more hw re-arrangements there after I can give it a go. But will be away tomorrow so it's not going to be that quick I'm afraid.

schakrava:
@MFlyer  you are right that stats are being collected by the data-collector for devices(via /proc/net/dev) and not for connections. Nice find.

schakrava:
With your patch, the numbers at the bottom of the widget are updating every few seconds as expected. This is good, it means data is coming through from the backend and being parsed properly for that part of the graph.

However, there's still a frontend bug which is preventing from the graph to be updated. So the graph is always blank. May be you'll find it interesting to dive in and see what's going on?

Here's a screenshot that might help a bit.
network_widget

MFlyer:
Hi @schakrava, I'll check it asap! - Going on with last mods on logs manager

MFlyer:
Hi @schakrava, got some news:
Tested and working on chrome, not on firefox

While logging:
Chrome at "time 0" get data.results with length = 2, while Firefox data.results has length to 0 raising errors (?!?)

I've also noticed that all line graphs (cpu too) don't render.

Generally speaking (pls ref. to #998) Firefox doesn't like svg ( poor performance and partial support [MDN page](https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_1.1_Support_in_Firefox) ) so we should think about moving to canvas

Mirko/Flyer

Issue #762, opened 2015-07-28
download logs on demand


schakrava:
Provide a feature to download latest log.tgz file from the web-ui.

MFlyer:
Hi @schakrava, can i work on this?

I can see the assignee, but think it's not active, am i wrong?

Flyer

schakrava:
You are more than welcome to, @MFlyer. this is indeed an important feature and it's safe to assume that @gkadillak is not actively working on it.

MFlyer:
Hi All @schakrava @phillxnet @priyaganti , here is a first preview of what planned for logs access via Rockstor Web UI [Log on demand preview](https://youtu.be/ENMv5gBtjtg)

I thought having it with just a link to download rockstor logs like a clone of actual error system, so i've decided to let user choose which logs to donwload on demand plus adding a "live log reader" - to be added modal window for loading files cat/tail

Any suggestion about layout & logs to add? :blush: Actually I've got template and view good skeleton and adding on them now it's easy, backend part is the missing one (I'm used to start always from frontend beacuse it's my "babau" and my frontend coding is a little rusty comparing to backend)

phillxnet:
@MFlyer This is great, well done. A small suggestion to assist with understanding what's what though:
you could put (kernel) in brackets after Dmesg akin to what you have done to explain smbd/nmbd. Same could be done for the Yum entry ie (packaging) or (updates), also for Nginx ie (WebUI).
Also the addition of a general system / systemd log akin to journalctrl would be good, currently just nicer time stamps but going forward there is a greater split between these. There might also be the option to employ systemd's log filtering capabilities (extensive), but that's most likely not a first throw feature in my opinion.

Not looking as closely as your self so take all of the above with a grain of salt.

This is going to be a great addition.

MFlyer:
Hi @phillxnet, just added your suggestions (brackets adds).

Talking about systemd-systemctl /journalctl : totally agree with you, but got some doubts for outputs

Ex. this is a `systemctl list-units -t service` just to limit items, but it still outputs a long log with long items too (fsck)

![system](https://cloud.githubusercontent.com/assets/17988902/14636318/07b35fb8-062b-11e6-969f-a55a58efddb7.jpg)


We risk to have few or too much infos, depending on how it get implemented :confused: 

We could get a <select> with services like <option>s (green bg if loaded and active, red if failing) and cast service details on select change, but probably that would overload users with datas

[...]

Issue #701, opened 2015-06-28
Alerts and notifications


schakrava:
Details need to be thought out, but creating this issue as a start with this helpful feedback from a user.

Ideally I'd have rockstor configured to send me an email when things like this happen.  It's good to know asap that you have a potentially sick hdd and the kernel is already smart enough to send mail to root, let's get that mail to my gmail address 

Other important information might be notification of reaching a configurable capacity threshold.  Errors found during scrubs.  Errors during replication jobs.  I'm sure there's more.

The last thought is a scheduled status notification.  I came from ZFS on napp-it where I would receive a weekly email telling me the status of my pools (basically zpool status -v which also happens to have the status of the latest scrub as well as checksum error information) and the usage of my pools.

maxhq:
Hi everyone,

I found a solution using `logcheck` that works very nicely for me!

Logcheck is a script invoked by `cron` that scans through your `/var/log/messages`, filters out unimportant messages and notifies the remaining ones via email.
Logcheck comes with a standard filter rule set (`/etc/logcheck/ignore.d.server/*`) that can easily be adjusted/enhanced by adding an own config file (`local-rockstor` in my case).

You need to do two things:
1. Set up email alerts in the web interface.
2. Install and configure `logcheck`

### Setting up logcheck
#### Installation
```
yum install logcheck
```
#### Configuration
```
# Make logcheck send mails to root (forwarded to your configured address) and attach logs
sed -ri 's/^(SENDMAILTO)=.*/\1=root/;s/^(MAILASATTACH)=.*/\1=1/' /etc/logcheck/logcheck.conf

# Make logfiles readable by logcheck
chown root:logcheck /var/log/messages /var/log/secure
chmod 640 /var/log/messages /var/log/secure

# Tell logrotate to make new files readable by logcheck
sed -i '/{/a \    create 0640 root logcheck' /etc/logrotate.d/syslog
```
#### Create ruleset to filter out Rockstor specific messages

```
cat <<'EOF' > /etc/logcheck/ignore.d.server/local-rockstor
^\w{3} [ :[:digit:]]{11} [._[:alnum:]-]+ kernel:( \[ *[[:digit:]]+\.[[:digit:]]+\])? BTRFS info
[...]

PR #1312, opened 2016-05-09
Issue#762 logs on demand

MFlyer:
PR on #762  Issue for Log utility - to be decided if to go on with this (like a version 1) and then update or wait till rotating logs integration


MFlyer:
Hi @schakrava, code revision for #762 closed over 169b09f, PR ready for your revision before merging :smiley: 

Important remarks:
- Spawning has been added for all functions, allowing for example current log download while still reading it on a cat -n and obviously better processes management by gevent
- Cat and tail -n * calls always end (we assume that for true, cos' it can't be different), so we don't care about "safe killing", otherwise done for tail -f (it hangs forever till user action and we fire it on modal closing or on disconnect - accidental browser end/crash/etc)
- Error handling for empty or 0 size logs is transparent to user (no debug info on console/error raising, but direct info to user via log reader modal)
- I didn't forget about @phillxnet suggestion for systemd info (I totally agree about its usefulness! :wink: ), but imho it makes more sense under "Services" and not on logs
- If you think to move LogManagerNamespace to another file and import it...yes, do it! :+1: (that would divide widgets from LogManager & eventually future websockets)

Flyer/Mirko

src/rockstor/smart_manager/data_collector.py
```diff
@@ -23,6 +46,265 @@
...
+            self.livereader_process.kill()
+            self.livereader_process = None
+            
+        #Emit current state only for checking purpose, no funcs handle on frontend
+        self.emit('logManager:logskilltailf', {
+            'key' : 'logManager:logskilltailf', 'data': {
+            'status' : self.livereading
+            }
+        })
+        gevent.sleep(0)
```
schakrava: Is gevent.sleep(0) necessary here? can you explain your rationale please?

src/rockstor/smart_manager/data_collector.py
```diff
@@ -23,6 +46,265 @@
...
+                live_out = self.livereader_process.stdout.readline()
+                self.emit('logManager:logcontent', {
+                    'key': 'logManager:logcontent', 'data': {
+                    'current_rows' : 1,
+                    'total_rows' : 1,
+                    'chunk_content' : live_out,
+                    'content_size' : 1
+                    }
+                })
+                gevent.sleep(0)
```
schakrava: Can you please explain the rationale behind putting gevent.sleep(0) in the while loop here?

schakrava:
Rotated logs section is appearing multiple times in the dropdown. Why is that?
screenshot from 2016-05-18 16 43 28

schakrava:
When I click on more than one item in the Available Logs and download, the requested_logs.tgz file only has the log file corresponding to the first item. Is this a bug or am I using/expecting wrong things?

schakrava:
My one word reaction to this pull request is an emphatic wow!!!

I couldn't have done it any better myself. I am really enjoying just playing with this feature, viewing and downloading various logs. I can see my browser memory usage go up and browser lock up because the data is being piped fast. I am sure we can make this a bit more efficient -- something to look into in near future. May be sleeping a bit more than the default 0 seconds could help?

Please take a look at all of my comments above and let me know your response. happy to test again with any new changes. @phillxnet and @priyaganti , you might want to deploy this branch, it's really fun to test!

MFlyer:
Hi @schakrava , I'm pleased to see your reaction to this PR :smiley: 

Talking about gevent.sleeps : 
- the first one on the emit is not required (there's not a rational reason eheh)
- the second one on the live reading with tail -f : while we are live reading in the modal window we can't do anything else except for downloading the log file (link inside modal), so we don't wont our func to sleep showing us all stdout asa there's something to render, but you're right saying it makes no particular sense to gevent.sleep inside while loop! (if modal get closed loop get killed, so that gevent.sleep isn't necessary at all)

Talking about "bugs" on logs list and avail-downloadable logs : going to check on Firefox, cos' under chrome it was all fine

MFlyer:
Found why you get duplicated elements on logs list under firefox:


While on Chrome (and Chromium too, i think) websockets stay alive, on firefox clicking the download link kills the websocket (see image); Rotated logs list is fetched on connection time on data_collector, so when the link get clicked our websocket dies, firefox reopens it and we get again rotated logs list...and so on every time you download a log file (both from archive builder and log download from modal window) :sob: 
websocketkilled

Going to investigate and find a workaround

> When I click on more than one item in the Available Logs and download, the requested_logs.tgz file only has the log file corresponding to the first item. Is this a bug or am I using/expecting wrong things?

We expect a tar file with all selected logs and this is what i get both on FF and Chrome

MFlyer:
https://github.com/rockstor/rockstor-core/pull/1312#issuecomment-220191294 - @schakrava : solved with cc3af6a :wink: Tested both on chrome/firefox under win (I love and "hate" guys from Mozilla eheheh, but finally they're right saying "Ok, if you click a link you're leaving so you don't need that websocket anymore and we close it!")

MFlyer:
Gevent.sleep(0) cleaning on 5a4b175 (those 2 referenced by @schakrava + others)

Last missing "bug" I'm not able to reproduce is requested_logs file having only one log: both on Chrome/Firefox tar file has all requested logs (opening with winrar under win).
Possible workaround: create requested_logs like requested_logs_UTC and check if it's just a kind of "file caching" <- @schakrava can you investigate if it's like this? :smiley: 

Flyer/Mirko

P.S. for fun: yesterday was sunny, tomorrow will be sunny, yesterday i was at work, tomorrow i'll be at work, today i'm on holiday (going to post office, get tickets for theater, etc etc)....today in Milan is rainy xD

schakrava:
Thanks @MFlyer . I am unable to reproduce the problem either. May be i did not test properly yesterday...

Anyway, everything looks great to me! http://lgtm.in/i/Qegg5iK6q

henfri:
Hello,
Wouldn't it make sense to optionally Filter the Logs with the help of logcheck?

MFlyer:
Good morning @henfri,
Logs manager behaviour is to let user read logs / download them on demand, while [logcheck](http://linux.die.net/man/8/logcheck) acts in another way (mail root about log changes): do you agree on that?

Flyer

henfri:
Yes and no.

Yes, logcheck mails log contents (not really changes, but triggered by time).
But the neat thing about logcheck is that it filters out the rubbish, noone is interested in. 
Thus, the conbination of your Userinterface and logchecks abilities to filter (frankly, it's just regexes that you can use directly in the userinterface without any of the logcheck binaries/scripts)

It is really great, because it helps uncovering issues and to focus.

phillxnet:
@henfri and @MFlyer We do have an issue that relates to the use of logcheck over at #701 where @maxhq also suggested it's use along with some example configs. Seems like a great program but with regard to @MFlyer 's very recent log gui addition I think we should let it settle as is for a bit and maybe consider logcheck in the light of the other issue.

henfri:
Hello Phillxnet,

yes, I am aware. 
I also sympathize with letting this feature (gui) settle. BUT: I am convinced that adding the filtering with the help of the logcheck filters to the GUI would *really* boost the usefullness.

As a user, I want to see important messages. Currently, they are hidden in a crowd of unimportant messages.

Greetings,
Hendrik

MFlyer:
Thanks to @phillxnet for ref to other issue and to @henfri Hendrik for suggestion about logcheck/ logcheck rules integration in UI (will come back to this, that's a promise :wink: )

PR #1315, opened 2016-05-17
make the page size maximum on shares collection. Fixes #1305

### LINKED ISSUE #1305: Replication task creation only displays shares from first page of shares.
I'm attempting to create a new replication task and I've noticed that I can't select any share that is on page 2 from my shares list. When I compare the shares that are available to select from for the replication task, the range of shares matches exactly what I see when I'm looking on page 1 of System > Shares. I've never created an issue so I'm hoping that I've done this correctly.

priyaganti:


schakrava:
Thanks @priyaganti . Once we figure out the datatables problem, we should be able to fix the pageSize issue permanently.

Another thought, is there a way to use paginated dropdowns?

PR #1316, opened 2016-05-20
Add nossd option in python file and to tooltip in pool Edit Compression

### LINKED ISSUE #1313: Allow 'nossd' mount option
I'm currently testing the use of bcache in rockstor and found that, by default, bcache reports that the backing devices are not rotational disks.  This causes btrfs to enable the 'ssd' mount option, by default.  I would like to override this behavior via the 'nossd' option, but the rockstor web interface does not allow this because the option is not in the hardcoded whitelist of mount options.

priyaganti:
fixes #1313

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/pool/compression_info.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/pool/compression_info.jst
@@ -1,14 +1,13 @@
-{{!--display_compression_details--}}
 <table>
     <tr>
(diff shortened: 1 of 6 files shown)


schakrava:
Thanks for opening this pull request @priyaganti . When I edit, currently active mount options are not being populated in the text field. We should populate the field with current info for better UX.

schakrava:
Thanks @priyaganti. everything checks out. nicely done :)

PR #1318, opened 2016-05-20
Datatables pagination

priyaganti:
initial test commit with datatable pagination changes


src/rockstor/storageadmin/static/storageadmin/js/rockstor.js
```diff
@@ -63,6 +63,13 @@ RockstorLayoutView = Backbone.View.extend({
 	});
     },
     
+    renderDataTables: function(event){
+		$('table.data-table').DataTable({
+		    "iDisplayLength": 15,
+		    "aLengthMenu": [[15, 30, 45, -1], [15, 30, 45, "All"]],
+		});
```
schakrava: We should have page sizes in RockstorGlobals instead of hardcoding them there.

src/rockstor/storageadmin/static/storageadmin/js/views/snapshots.js
```diff
@@ -42,7 +42,7 @@ SnapshotsView = SnapshotsCommonView.extend({
 	this.snapshots = this.options.snapshots;
 	this.collection = new SnapshotsCollection();
 	this.shares = new ShareCollection();
-	this.collection.pageSize = RockStorGlobals.maxPageSize; 
+	this.collection.pageSize = RockStorGlobals.maxPageSize;
```
schakrava: Do we still need this? You've already set the pageSize to maxPageSize in the paginated collection(parent class), no?

schakrava:
Nice work @priyaganti . This gives our UI a nice snappy upgrade. I think users will really like the ability to search and sort. Thanks for contributing yet another widespread and significant enhancement!

Issue #1327, opened 2016-05-25
Scheduled task - crontab window - wildcards breaking scripts


MFlyer:
Starting from issue #1296 and @maxhq we found that crontab window string for "run always" (*-*-*-*-*-*) can break crontab running because * is a shell wildcard ( @maxhq having this problem, not me or others, anyway we patch to prevent it )

Solution : backslash before first char to escape it

schakrava:
Fixed by #1328

PR #1328, opened 2016-05-25
#1327 - Patch to first char of crontab window being a wildcard

MFlyer:
This patch helps preventing wildcards on first char of crontab window string from breaking scripts on crontab.
Ref. #1327 started from old issue #1296


schakrava:
Thanks for the fix and really owning this @MFlyer!

Issue #1335, opened 2016-06-02
Users page doesn't show system users shell


MFlyer:
Hi @schakrava , while working on #1290 I noticed that our users page has no value for system users shell so...I added it :wink: 

PR ready

Not a bug, not a problem, but I didn't like all those empty cells (while not really empty or null)!

Flyer

schakrava:
Fixed by #1336. Nicely done @MFlyer !

PR #1336, opened 2016-06-02
Users page - Sys Users without shell info

MFlyer:
#1335 Fixed grabbing shell info over getent passwd

Flyer/Mirko


schakrava:
Nice enhancement @MFlyer, sorry for the delay in my review. LGTM

Issue #1048, opened 2015-12-09
Add support for multi device pool in single raid profile


schakrava:
This may be desired in case of disks of different size. In case of single profile, the metadata defaults to single in Rockstor but dup in btrfs. We should let the user select the metadata profile in some kind of advanced option area in the create form.

http://forum.rockstor.com/t/adding-new-disks/765/3

schakrava:
Fixed by #1337

PR #1337, opened 2016-06-03
Add ability to add/remove drives from single raid profile

bskrtich:
This fixes Issue #1048


schakrava:
Welcome to core contributor club of Rockstor @bskrtich! Nicely done, and you get extra respect for test code. Everything checks out.

I do want to add a note here for future reference, that when removing a drive from a multi drive single profile our pre balance space check may pass, but btrfs balance could then come back with a no space left behind error, in which case the resize fails. No big deal in most cases, but the user will need to run a balance before retrying.

schakrava:
@bskrtich I've mentioned your contribution in our forum. Are you a user there? If so, please reply to the topic so we can properly attribute your contributions!

https://forum.rockstor.com/t/dev-log-for-3-8-14/1412/30

Issue #1051, opened 2015-12-13
Use disk serial throughout ui not device name


phillxnet:
E.g. when creating a pool we should present the disk serial numbers as the device names are essentially arbitrary and change from one boot to the next.
ie rather than vda, vdb, etc with table header of Disk Name use device serial numbers with the table header of Disk Serial
The serial numbers are not arbitrary and are often printed on the end of the drive anyway.

phillxnet:
I think there is still a place for the device name on the disks screen as this represents the appropriate level for device names.

phillxnet:
As pr #1320 changes to by-id (boot stable) device names which in that case contain the serial number of the drive (where available) it also addresses this issue, and as we are still using a device name, all be it one that contains the serial, we are also covering the other point raised in this issue: to maintain device names as well, only it will now be a boot to boot stable one that is also likely to contains the model number as well.

PR #1326, opened 2016-05-23
pool disk removal wizard shows no disks Fixes #1325

phillxnet:
I have build current trunk with this patch set applied and was then presented with a disk table within the remove disk pool resize dialog. I also tested adding and removing a single disk from a 2 to 3 disk raid1 (respectively).
No other testing done I'm afraid as needed this for another pr I'm working on.
Sorry for scant testing and details but seems like a fairly safe (and small) change set.

@schakrava and @priyaganti Submitting for Review.

Fixes #1325

Issue #1325, opened 2016-05-23
pool disk removal wizard shows no disks


phillxnet:
To reproduce (current trunk code):
Create a 3 disk raid 1 pool
Resize pool
Remove disk
Expect a table of disk members to be shown but non shown.
Web browser console shows missing handlebars mathHelper.

I have fixed this locally by copying a handlebars mathHelper from another location and will now submit a pr against this issue with my crude fix for review.

PR #1338, opened 2016-06-04
revise internal use and format of device names. Fixes #1320

### LINKED ISSUE #1320: Revise internal use and format of device names
In order to increase the flexibility and utility of the db Disk.name field it is proposed that it be re-deployed to hold by-id device names as opposed to the current ‘sd’ type names. This in combination with the Disk.role field is intended to aid in supporting additional storage options such as LUKS full disk encryption.

Ie change:
/dev/sdX derived from a db entry of sdX
to:
/dev/disk/by-id/wwid-for-device derived from a db entry of wwid-for-device.

A World Wide Identifier (WWID) is a persistent and system independent identifier required of all SCSI standards compliant devices.
Reference:-
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Storage_Administration_Guide/persistent_naming.html

phillxnet:
Changes the format of disk names saved in db Disk.name field from sda type to by-id type.
Builds on existing conversion methods established in pr #1292 to translate transient sda type names to by-id type names.

It was found that in some corner cases such as virtio drives with no serial that by-id type names were not available. In this case the vda type name is stored instead. However this creates an issue with assumptions on drive path now being /dev/disk/by-id from the previous /dev type path. As this affected smart calls via the wrapper of all smart calls introduced in #1206 to find the base device and the enhancement of this introduced in #1221 we simply disable smart calls on devices that have been found to have no serial ie ascribed with fake-serial-. Also to account for this possible existence in the Disk.name field of devices that don't have by-id type names these devices have also been excluded in all future pool creation by adding a check for fake-serial to disk availability for pool selection. This avoids future pools being created with disks that we can't track due to no serial number and hence no by-id type name. The existing user facing serial warning has been updated to indicate this.

Note that as the structure of by-id type names affords a syntactic method for acquiring the base device it was also possible to remove system calls to lsblk to establish these base devices for use in all SMART system calls. As only transient use is made of sda type names, mainly when reading directly from lsblk in scan_disks() and from the immediate output of some btrfs commands and given all internal uses is now transitioned to by-id we have an enhancement in log diagnostic capability as all calls involving drive names are now directly traceable to the device given by-id type names contain model and serial number and are reboot safe. On this point many more modern drives exhibit multiple by-id type names and to stabilise this I have chosen to use the longer of however many are found. This has so far proven to be the one containing model and serial rather than the shorter wwn-<long-number>. This also addresses an observed and rare instability in by-id name presentation order as an earlier implementation of by-id name retrieval such as that in pr #1292 simply used the first instance listed.

An important element assumed in this pr is that no device without a reboot safe by-id type name can from here on constitute a component of new pools. This move improves code clarity and give us a sound datum on which to base our device naming.

All current disk operations have been tested as working with this pr as is (given the addition of the changes in #1326 (already merged) which were required in order to test disk removal from pools).

Also note that a significant number of methods / functions have been moved from fs/btrfs to system/osi. This was by way of existing TODO:'s and significantly reduces the possibility of circular dependencies by reducing the number of import statements across these two files in particular. The main advantage currently being the segregation of all directly btrfs related functions. Many fresh builds have proven the imports as presented in this pr to be sound. There is further improvement to be had separation wise such as setting up an hdparm specific functions file in for example system/hdparm and moving appropriate methods from system/osi to there, but this requires additional TODO:'s for review purposes and may be included later on in this pr for implementation in a future pr.

~~N.B. this pr still contains debug logging which has yet to be removed but is initially intended to serve in the earlier stages of the review process. Once more testing / review has taken place this logging will be removed.~~


phillxnet:
@schakrava pr ready for review, as previously discussed it may be best for this to be considered as a testing channel beginner. I am still testing the pr myself but pretty sure I've already covered all that it may affect.

N.B. does not include the fixes for no disks shown in disk removal wizard ie issue #1325 and now merged pr #1326 which are required to test that function here.

There are also many moves of code from fs/btrfs to system/osi and I think they are primarily responsible for the no 'auto merge' flag here as otherwise I would have expected this to merge OK.

Trivial side effects: the by-id names are much longer so the Disk Widget gets them all on top of one another when more than 2 disks are active. Considered this to be worthy of another issue as essential function is there still and a proper fix may be to establish an index type arrangement in the widget or redo the top part of the widget some how.

phillxnet:
Fixes #1320 and by a side effect also:
Fixes #1051 as the preferred (longest) by-id device names contain the device serial number.

phillxnet:
Found un-migrated hard coded dev path in snapshot role back code; working on this now.

phillxnet:
N.B. updates to testing code may be desired post this pr as:-
src/rockstor/storageadmin/tests/test_disks.py
src/rockstor/storageadmin/tests/test_disk_smart.py
src/rockstor/storageadmin/tests/test_pools.py
src/rockstor/storageadmin/tests/test_pools.py
have 'sda' type naming for api calls which could be changed to reflect more appropriate by-id type names along with the test environment: presumably defined by the json files in:
src/rockstor/storageadmin/fixtures

schakrava:
@phillxnet now that we are past 3.8-14, I think we are ready for this pr. Can you please rebase with latest master and let me know once it's ready for me to play and test?

phillxnet:
I am about to re-based this pr's branch on the just released 3.8-14 (tagged) master branch and to avoid replication of commits I will remove the github remote branch, voiding this pr. I will open a new pull request shortly based on the new branch.

phillxnet:
Please address all future comment to the newly opened pr containing the same commits but rebased on 3.8-14.
#1357

Issue #1343, opened 2016-06-10
Possible security flaw over userdel


MFlyer:
While on #1290 found out that when deleting managed users actually their samba credentials stay alive in samba db - Think this isn't desired

schakrava:
Fixed by #1344

PR #1344, opened 2016-06-10
#1343 Added autodel of samba credentials on user delete - tested and working

MFlyer:
Hi @schakrava , have nothing to tell - subject says all :wink: 

Important: PR #1336 before this - working on same files

### DIFF
--- a/src/rockstor/system/users.py
+++ b/src/rockstor/system/users.py
@@ -73,7 +73,7 @@ def get_users(max_wait=90):
             if (len(ufields) > 3):
                 charset = chardet.detect( ufields[0] )
                 uname = ufields[0].decode( charset['encoding'] )
-                users[uname] = (int(ufields[2]), int(ufields[3]))
+                users[uname] = (int(ufields[2]), int(ufields[3]), str(ufields[6]))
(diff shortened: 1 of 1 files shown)


src/rockstor/system/users.py
```diff
@@ -100,6 +100,9 @@ def userdel(uname):
     except KeyError:
         # user doesn't exist
         return
+    #Ensure user get deleted from samba pass db
+    cmd = [SMBPASSWD, '-x', uname]
+    p = subprocess.Popen(cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
schakrava: Hi @MFlyer . Why not use run_command utility method instead of calling subprocess.Popen?

MFlyer:
Hi @schakrava , your right about run_command! Done over 4285515 : no need of try except because all managed users have a samba entry so every delete action will have a return code of 0 :wink: 

Flyer

schakrava:
Thanks @MFlyer, nicely done as always!

MFlyer:
Hi @schakrava , it was a pleasure and had fun as always! :)

Issue #1352, opened 2016-06-17
replication receive error getFrequency helper missing


priyaganti:

schakrava:
Fixed by #1353

PR #1353, opened 2016-06-17
remove getFrequency helper call from replication receive.

priyaganti:


priyaganti:
fixes #1352

schakrava:
Thanks @priyaganti ! works!

Issue #1043, opened 2015-12-04
dev name change breaking mounts


phillxnet:
Corner case where drive name change breaks mount points.
In looking into reports of samba not starting even after pull request:-
https://github.com/rockstor/rockstor-core/pull/1032
which dealt with a bootstrap fail.
I have reproduced a very low frequency issue of mount point failures due to drive name change.
In my instance samba still started however the share was empty. This does not fit with a report 3.8-9.07 still having a failed samba start by @kimbl on the forum:-
http://forum.rockstor.com/t/pool-not-mounted-on-reboot/731/4
but may be related as the pool is not mounted as expected on boot.

Example of this issue:-

btrfs fi show
Label: 'rockstor_rockstor'  uuid: 0faa780f-a339-43b7-a49f-59cfa98548b5
	Total devices 1 FS bytes used 1.74GiB
	devid    1 size 25.47GiB used 3.27GiB path /dev/sda3

Label: 'time_machine_pool'  uuid: 8f363c7d-2546-4655-b81b-744e06336b07
	Total devices 3 FS bytes used 29.92GiB
	devid    1 size 149.05GiB used 19.00GiB path /dev/sdb
	devid    2 size 153.38GiB used 24.01GiB path /dev/sdc
	devid    3 size 149.05GiB used 19.01GiB path /dev/sdd

btrfs-progs v4.2.1

In the above we have the only non rockstor_rockstor pool as consisting of sdb, sdc, and, sdd

However Rockstor has attempted to mount our samba share (which is in the time_machine_pool) via reference to sda, however as can be seen sda is part of the system drive and as such doesn't contain any btrfs only the standard partitioning.

This results in may mount errors for the various mount points eg:-

[04/Dec/2015 15:38:34] ERROR [storageadmin.views.command:80] Exception while mounting a share(samba_share) during bootstrap: Error running a command. cmd = ['
/bin/mount', '-t', 'btrfs', '-o', u'subvol=samba_share', u'/dev/sda', u'/mnt2/samba_share']. rc = 32. stdout = ['']. stderr = ['mount: /dev/sda is already mou
nted or /mnt2/samba_share busy', '']

this is soon there after followed by:-

[04/Dec/2015 15:38:35] DEBUG [storageadmin.views.command:126] Bootstrap operations completed
[04/Dec/2015 16:19:53] DEBUG [smart_manager.data_collector:276] Sysinfo has been initialized
[...]

PR #1354, opened 2016-06-20
Issue#1290 password recovery system via web ui

MFlyer:
#1290 PR ready for review.

Hi @schakrava, here is my working Password recovery system via Pincards.

Actually it's like this:

- Every managed user and root user can have a Pincard for password reset
- Root user (or generally uid 0 user) password reset requires OTP confirmation too (Email alerts enabled and one time password sent to admin mail, to be added to required pins)
- Users are shown Pincard only one time, no way to recover a Pincard (every click for Pincard means a new Pincard generated)
- Pincards are in ID-1 format (or at least like ID-1 format), generated via html5 canvas and can be saved as png image or via text table (avail under Pincard image)
- "Reset password" box on login page with all info has been extended with Pincard reset, give it a try (it's easier than explain it :smile: )
- Password reset nicely tested and working on managed user, managed user with webui access, root user

Note: obviously for OTP reset on root user we assume all email settings are right!

Mirko

PR #1357, opened 2016-06-24
revise internal use and format of device names. Fixes #1320

### LINKED ISSUE #1051: Use disk serial throughout ui not device name

### LINKED ISSUE #1320: Revise internal use and format of device names

phillxnet:
Changes the format of disk names saved in db Disk.name field from sda type to by-id type.
Builds on existing conversion methods established in pr #1292 to translate transient sda type names to by-id type names.

It was found that in some corner cases such as virtio drives with no serial that by-id type names were not available. In this case the vda type name is stored instead. However this creates an issue with assumptions on drive path now being /dev/disk/by-id from the previous /dev type path. As this affected smart calls via the wrapper of all smart calls introduced in #1206 to find the base device and the enhancement of this introduced in #1221 we simply disable smart calls on devices that have been found to have no serial ie ascribed with fake-serial-. Also to account for this possible existence in the Disk.name field of devices that don't have by-id type names these devices have also been excluded in all future pool creation by adding a check for fake-serial to disk availability for pool selection. This avoids future pools being created with disks that we can't track due to no serial number and hence no by-id type name. The existing user facing serial warning has been updated to indicate this.

Note that as the structure of by-id type names affords a syntactic method for acquiring the base device it was also possible to remove system calls to lsblk to establish these base devices for use in all SMART system calls. As only transient use is made of sda type names, mainly when reading directly from lsblk in scan_disks() and from the immediate output of some btrfs commands and given all internal uses are now transitioned to by-id type we have an enhancement in log diagnostic capability as all calls involving drive names are now directly traceable to the device given by-id type names contain model and serial number and are reboot safe. On this point many more modern drives exhibit multiple by-id type names and to stabilise this I have chosen to use the longer of however many are found. This has so far proven to be the one containing model and serial rather than the shorter wwn-. This also addresses an observed and rare instability in by-id name presentation order as an earlier implementation of by-id name retrieval such as that in pr #1292 simply used the first instance listed.

An important element assumed in this pr is that no device without a reboot safe by-id type name can from here on constitute a component of new pools. This move improves code clarity and give us a sound datum on which to base our device naming.

All current disk operations have been tested as working with this pr, just prior to re-basing on 3.8-14.

Also note that a significant number of methods / functions have been moved from fs/btrfs to system/osi. This was by way of existing TODO:'s (added in #1292) and significantly reduces the possibility of circular dependencies by reducing the number of import statements across these two files in particular. The main advantage currently being the segregation of all directly btrfs related functions. Many fresh builds have proven the imports as presented in this pr to be sound (prior to re-basing). There is further improvement to be had separation wise such as setting up an hdparm specific functions file in for example system/hdparm.py and moving appropriate methods from system/osi to there, but this requires additional TODO:'s for review purposes and are included later-on in this pr for implementation in a future pr. This is in response the the same suggestion from @schakrava in pr #1292 

Known issues:
A cosmetic side effect given the by-id names are much longer is that the Disk Widget gets them all on top of one another when more than 2 disks are active. Considering this to be worthy of another issue / pr as essential function is there still and a proper fix may be to establish an index type arrangement in the widget or redo the top part of the widget some how.

Fixes #1320 and by a side effect also:
Fixes #1051 as the preferred (longest) by-id device names contain the device serial number.
It is also assumed that this pr will sort out the root cause of #1043 "dev name change breaking mounts" but this will require additional testing before confirmation.

Please see the previous version of this pr (#1338) opened 20 days before this one for developmental history to date. All currently relevant info is replicated in this pr.


phillxnet:
@schakrava Rebased on latest master (3.8-14 git tag) and re-tested as requested.
On regular, bios raid, and mdraid system disk installs all disk functions were as expected.
Submitting as ready for review.

MFlyer:
To @schakrava - Cross reference to #1354 - Both PR have mods over data_collector

src/rockstor/fs/btrfs.py
```diff
@@ -782,346 +809,25 @@ def device_scan():
...
 
 
 def btrfs_label(disk):
-    o, e, rc = run_command([BTRFS, 'filesystem', 'label', '/dev/%s' % disk])
+    o, e, rc = run_command(
+        [BTRFS, 'filesystem', 'label', '/dev/disk/by-id/%s' % disk])
     return o[0]
 
 
 def btrfs_importable(disk):
```
schakrava: even though this is not a function you introduced, it's misleading and not used. Remove it if you can please.

src/rockstor/storageadmin/models/disk.py
```diff
@@ -18,14 +18,18 @@
...
 
 class Disk(models.Model):
     """Pool can be null for disks that are not part of any pool currently"""
     pool = models.ForeignKey(Pool, null=True, on_delete=models.SET_NULL)
-    """typically sda, sdb etc.. max_length = 0 supports 100s of disks"""
+    """ Previously the name field contained sda, sdb etc..
+    Revised to contain device names for use with the udev created links at
+    /dev/disk/by-id/ which in turn are symlinks to sda, sdb etc.
+    eg ata-QEMU_HARDDISK_QM00005 ie mostly derived from model and serial number.
+    """
```
schakrava: I support this change, but strongly prefer that primary user facing name of a Disk should be the more readable/humanized one such as sda, sdb etc.. Since they are not persistent, it makes sense to store the persistent by-id name in the db instead. However, I'd like to suggest a property that retrieves the readable name during runtime as the model is feteched.

This also requires some UI changes, wherein the readable name is presented to the admin for various management matters like creating a Pool, adding disks etc.., with the Disk.name in an additional column or in a tooltip. This way, we achieve a pleasant user experience while not compromising the true identity of the Disk.

src/rockstor/system/osi.py
```diff
@@ -786,48 +1234,136 @@ def set_disk_spindown(device, spindown_time, apm_value,
...
+    validation mechanisms and are intended to have more repeatable behaviour but
+    only work on a single device at a time.
+    A single call to this method can provide all current by-id device
+    names mapped to their sdX counterparts with the latter being the index.
+    :return: dictionary indexed (keyed) by sdX type names with associated by-id
+    type names as the values, or an empty dictionary if a non zero return code
+    was encountered by run_command or no by-id type names were encountered.
+    """
+    byid_name_map = {}
+    out, err, rc = run_command([LS, '-l', '/dev/disk/by-id'],
```
schakrava: just a recommendation -- instead of using run_command here, we can simply use python to generate this map. This entire function reduces to

import os, os.path
dirname = '/dev/disk/by-id'
m = {}
for os.listdir(dirname):
    m[os.path.split(os.path.realpath(os.path.join(dirname, d)))[-1]] = d

phillxnet:
@schakrava Thanks for the review.

Since you review I have removed the unused function in fs/btrfs that you indicated along with another I found in the same file. I also removed the comment inconsistency in the other function you commented on. Alas I am unable to translate it to the neater os.listdir as there was a requirement to have a 'long' list via 'ls -l' command. But this did help with my development of the suggested run time property extension of the Disk model to provide short name (ie sda type) retrieval via this extension. I have called this property temp_name to emphasise their temporary nature. The function of this property has been tested via logging prior to commit and works as expected.

As discussed outside this thread I have not enabled / re-surfaced the short names to the UI as that is best done by others more able (and quicker) in that area and is arguably better done is another pr given the vintage and size of this pr. So in the interests of getting this one out I am leaving the long and arguably cumbersome by-id type names in the UI. This mostly affects the readability of the disk names as given in the Dashboard Disk Activity widget as indicated earlier in this pr.

Given the above caveats this pr is again ready for review / merge.

@MFlyer the only additional files I have altered that impact on your large pending pr are serializers.py but I don't think this will cause any merge issues as our changes were in different areas and relatively small.

Sorry I dragged this one out folks and hope it helps.

MFlyer:
hi @phillxnet concurrent mods actually are on serializers and data_collect, but I totally agree with you about differents areas not making merge issues :+1: 

I'm really curious to see latest changes by you PR :)

schakrava:
Thank you @phillxnet !

phillxnet:
@schakrava Cheers.

PR #1365, opened 2016-07-03
Issue#1362 local time on ui

### LINKED ISSUE #1362: Add local/current time on the UI
Here's a user report

What i find a little annoying is that there is no current time shown on the Dashboard. I can see uptime, but how can i tell if my system is running on the right local time? Yes, i have this set at the Installation, but to know if time is correct and to maybe set another timeserver it would be nice to have this in the Dashboard or under System Menu to control and set.

How can i check the current time and timeserver? Should i go to the Terminal and try the right commands for centOS?

MFlyer:
Hi @schakrava this is a starting 'sunday evening relax time' :smiley: solution for having local time on UI (data_collector sends current localtime + timezone datas every 20 secs).
PR tested and working.

Avoided setInterval - see my #1362 post

Screenshot
time

Future update idea : add possibility to click on current time and manually change time / timezones via systemd `timedatectl`

This fixes #1362 
Mirko

### DIFF
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -584,7 +584,7 @@ class SysinfoNamespace(BaseNamespace, BroadcastMixin):
             self.emit('sysinfo:localtime', {
                 'data': time.strftime('%H:%M (%z %Z)'), 'key': 'sysinfo:localtime'
             })
-            gevent.sleep(20)
+            gevent.sleep(40)
 
     def send_kernel_info(self):
             try:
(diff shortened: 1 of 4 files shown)


schakrava:
Nicely done @MFlyer . Just a couple of quick comments. 1. since the precision is at minute level, we can probably get by sending the info every 40 or 50 seconds instead of 20 seconds. Also, "Local Time" seems a bit confusing to me. What do you think about calling it "System Time" or something else on those lines?

MFlyer:
Hi @schakrava, no problem moving to 40 secs (IMO 50 secs too close to 1 min, so if you update on x.x.59 you'll be always 1 min before current real time).
Agree with "System Time" too

schakrava:
Thanks @MFlyer. Happy to merge!

PR #1367, opened 2016-07-05
improve dashboard disk activity widget for by-id names. Fixes #1366

### LINKED ISSUE #1366: improve dashboard disk activity widget for by-id names
As from version 3.8-14.02 testing updates the new by-id names are frequently overlapping in the dashboard disk activity widget as they are much longer that the temp names used previously. There are a number of trivial changes that can be applied to this widget that should help with this issue, at least for the time being, including making better use of the available width.

phillxnet:
With the introduction of by-id names in pr #1357 the dashboard 'Disk Activity Widget' suffered from mostly overwritten writing detailing the drive names and their snapshot percentage of use in the horizontal 'Top Disks' bar. This is mostly addressed by increasing the disk bar width to make use of previously unused horizontal space and removing the text percentage indicator as this is redundant given the graphical display and costs 9 horizontal characters per disk name. The by-id name is then subdivided so that only the guaranteed unique serial component is used as drive identifier. Given the increased use of available widget width along with dropping the 9 characters per name on the text percentage we have a roughly similar though slightly higher text density.

Further improvements can be had in this widget but this pr represents a move in the right direction and largely, but not wholly, reverts the regression introduced by the use of the much longer by-id names.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/dashboard/disk_utilization.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/dashboard/disk_utilization.js
@@ -306,7 +306,7 @@ DiskUtilizationWidget = RockStorWidgetView.extend({
         if (d.name == 'root') {
           return '';
         } else {
-          return d.name + ' (' + (d.dx*100).toFixed(2) + '%)';
+          return d.name;
         }
       })
       .attr('fill-opacity', 1.0);
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/dashboard/disk_utilization.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/dashboard/disk_utilization.js
@@ -52,7 +52,7 @@ DiskUtilizationWidget = RockStorWidgetView.extend({
       this.disks.pageSize = RockStorGlobals.maxPageSize;
 
     this.topDisks = [];
-    this.topDisksWidth = this.maximized ? 400 : 200;
+    this.topDisksWidth = this.maximized ? 520 : 240;
     this.topDisksHeight = 50;
 
     this.selectedDisk = null;
@@ -170,6 +170,7 @@ DiskUtilizationWidget = RockStorWidgetView.extend({
     var _this = this;
     this.disks.each(function(disk) {
       var name = disk.get('name');
+      var temp_name = disk.get('temp_name');
       _this.disksData[name] = [];
       for (var i=0; i<_this.dataLength; i++) {
         _this.disksData[name].push(_this.genEmptyDiskData());
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/dashboard/disk_utilization.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/dashboard/disk_utilization.js
@@ -85,10 +85,10 @@ DiskUtilizationWidget = RockStorWidgetView.extend({
       yaxis: {
         min: 0
       },
-			series: {
-        lines: { show: true, fill: false },
-        shadowSize: 0	// Drawing is faster without shadows
-			}
+      series: {
+        lines: {show: true, fill: false},
+        shadowSize: 0 // Drawing is faster without shadows
+      }
     };
     this.dataGraphOptions = {
       grid : {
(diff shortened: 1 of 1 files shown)


phillxnet:
Fixes #1366 
@schakrava pr ready for review.  It's not a fix all but should be mostly good for current limit of top 5 disks when maximised (as before) and top 3 disks when not. Although top 2 disks when not maximised might be more practical here.

phillxnet:
Proof of improvement:
4 disks prior to pr (4th disk is white and doesn't extend to the edge of the widget):
before-4-disks
Then 4 disks using the whole widget width and only the serial component of the by-id name:
4-by-serial-for-space-demo
Also demonstration of 3 disks displayed via their serial in non maximised widget:
3-disks-by-serial-default

phillxnet:
In the above pictures there is only really one device with a particularly large serial, and with many readings there will still be overlap but not nearly as bad, but it is a move in the right direction and maintains consistency with the drop down for naming. We could in the future consider trimming the serial when long and returning a percentage indicator however there is confusion in the percentage readout especially when the write tick is enabled.

phillxnet:
Tried to extend the width of the graphs but there seems to be some 'intertwined goings on' there, ie changing "this.dataLength" didn't work out so stuck to 'top disks' text issue in this pr.

phillxnet:
Note that I have added an as yet unused temp_name variable that may come in for maybe an alternative (another tick) display option. Currently as we don't display this anywhere else just yet it seemed inconsistent for the time being.

schakrava:
Thanks @phillxnet . Why not display the shorthand/temp_name in the widget?

phillxnet:
@schakrava Simply because this was intended as a quick fix and given the short temp_names are not (currently) available any where else in the Web-UI I wanted to maintain naming consistency. And the in widget drop down serves as a key to reference the serial back to the by-id for further device info. If I had reverted fully to short temp_names then it would be a complete disconnect naming wise with the rest of the UI so this seemed like the way to go for now and less of a backward step. We could at a later date add a tick to switch to short temp_names but without an in place key it would promote a ping pong UX as users move between a page (presumably Disks) and the dashboard in order to find out what device is using for example the temp_name sdd at the time they see it being busy. This is assuming of course that the Disks page later gets a short name column / mouse over or what ever.

I had considered placing in brackets by each name in the drop down selector the equivalent temp_name and then displaying the temp_name in the bar but that seemed like a more complex change than what I did here, and given the obvious fail of the full by-id as was I wanted a quick but consistent fix.

Shall I chip away some more? I think what we currently have in the testing channel (this widget wise) is obviously broken (from my by-id changes) and I wanted to make it quickly less broken: hence the pr. And then when/if temp_names can be indexed to actual devices within the Web-UI we could add them to the widget as a tick option, or go the bracket reference option route which I didn't fancy coding when I was here last.

Hope that make sense.

phillxnet:
@schakrava Where do we stand on this one. I see it as an improvement on what is currently broken visually (and released) and I wanted this in as a quick fix. I'm not likely to re-visit this for a bit as now on the unit testing for a while. Can this not serve as a stop gap until we implement the ideas already discussed. At least then we don't have writing over itself as is the case now (at least much less so anyway). As indicated it doesn't only address the name change but also makes better use of space (the width of the widget) and removes a confusing 100% display on one drive taking up only a segment of the display with others. It also changes the number of disks displayed depending on the widget's state. These are all improvements we are likely to want to keep, even if we revert the names used, or better still do another tick to select between this and old temp_names display style.

MFlyer:
Adding cross reference to #998

schakrava:
Finally got to test it. The display of name is much better now. You were surgical with the display on this one :) 

I think we should just do what btrfs-progs does and just show sda names by default and provide a way to look at longer unique names easily. we can deal with that separately. Thanks for the pr @phillxnet !

MFlyer:
Now ready to go on with last widget over Chart.js :smile:

phillxnet:
@schakrava Cheers and thanks, mostly agree on dev names as well; not ideal 'as is' but better than I left it after the by-id changes at least.
@MFlyer well done on your continuing dashboard work by the way.

Issue #518, opened 2014-10-27
Console access from the Web-UI


schakrava:

MFlyer:
Hi all @priyaganti @phillxnet @schakrava , here is a preview for ongoing shell access implementation on Rockstor, hope you'll enjoy it :smiley: 

https://youtu.be/ekGDnPxj13E

Todo list

- [ ] Increase nginx security to avoid direct access to shell page outside Rockstor (testing)
- [ ] Add shellinabox service management over services page
- [ ] Add shellinabox service customization (define different port and maybe black on white shell css)

phillxnet:
@MFlyer This looks great, you are really knocking out some great stuff. Good luck with the security side, not something I can help with unfortunately (the security side). I take if from the text within the video that initial login is to be limited to admin user only, with su to root there after, is that right? Does this mean that the admin password is over https (as normal) and the root password is then inside a ssh session of it's own which is in turn also inside the https?

MFlyer:
Hi @phillxnet , security side is a challenge to me too eheheh (had to find a way to let shell only inside Rockstor without locking it at all - not so easy because of shellinabox structure and probably only possible with some nginx hacks)

>  I take if from the text within the video that initial login is to be limited to admin user only, with su to root there after, is that right? Does this mean that the admin password is over https (as normal) and the root password is then inside a ssh session of it's own which is in turn also inside the https?

First part you're right!
Shellinabox configured to only allow access from localhost with LOGIN auth that is like a tty login (root user hasn't login auth so we ensure 2 logins: any web-ui created user next su with root) where you should never auth directly with root (our choice :) )
Less secure alternative: SSH auth allowing direct root login (not our choice!)

**HTTPS**

Actually there's a new nginx location named shell (48a953f) that acts like on socket.io address, shellinabox not directly on https because of proxy_pass with current django-nginx already on https (django-nginx over https -> access to /shell address -> nginx proxy_pass moves to http://127.0.0.1:4200, not exposed to frontend, like socket.io), so yes, we're always under https :+1: 

Flyer

MFlyer:
Nginx security / avoid direct access to shell : 02c5a92 :wink: 

Obviously someone hacking could modify http request header, but:

1. This assumes he/she already knows about Rockstor on lan
2. He/she knows Rockstor hostname/ip 

And if 1 & 2...he/she has to know a Rockstor web ui account infos (uname&pass) and Rockstor root pass

I think current protection can be ok

Flyer

[...]

PR #1388, opened 2016-07-11
Issue#518 console access via web ui shellinabox

MFlyer:
Hi @schakrava,
PR has been tested and seems ready for your review.

For a full reference please look this https://github.com/rockstor/rockstor-core/issues/518#issuecomment-230730771

Please check buildout parts cos' I'm not a buildout guru eheh :)

Flyer

### DIFF
--- a/src/rockstor/scripts/prep_db.py
+++ b/src/rockstor/scripts/prep_db.py
@@ -37,6 +37,7 @@ def register_services():
         'NUT-UPS': 'nut',
         'ZTaskd': 'ztask-daemon',
         'Bootstrap': 'rockstor-bootstrap',
+        'Shell In A Box': 'shellinaboxd',
         'Rockstor': 'rockstor',}
 
     for k,v in services.items():
--- a/src/rockstor/scripts/prep_db.py
+++ b/src/rockstor/scripts/prep_db.py
@@ -18,6 +18,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 from smart_manager.models import Service
 from storageadmin.models import Setup
+from django.conf import settings
 
 
 def register_services():
@@ -38,12 +39,19 @@ def register_services():
         'ZTaskd': 'ztask-daemon',
         'Bootstrap': 'rockstor-bootstrap',
         'Shell In A Box': 'shellinaboxd',
-        'Rockstor': 'rockstor',}
+        'Rockstor': 'rockstor'
+        }
+
+    services_configs = {
+        'shellinaboxd': '{"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}'
+        }
 
     for k,v in services.items():
         try:
             so = Service.objects.get(name=v)
             so.display_name = k
+            if v in services_configs:
+                so.config = services_configs[v]
         except Service.DoesNotExist:
             so = Service(display_name=k, name=v)
         finally:
--- a/src/rockstor/scripts/prep_db.py
+++ b/src/rockstor/scripts/prep_db.py
@@ -18,7 +18,6 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 from smart_manager.models import Service
 from storageadmin.models import Setup
-from django.conf import settings
 
 
 def register_services():
--- a/src/rockstor/storageadmin/static/storageadmin/js/router.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/router.js
@@ -96,7 +96,7 @@ var AppRouter = Backbone.Router.extend({
 	    "add-afp-share": "addAFPShare",
 	    "afp/edit/:afpShareId": "editAFPShare",
 	    "rockons": "showRockons",
-		"shell": "showShell",
+	    "shell": "showShell",
 	    "images": "showImages",
 	    "containers": "showContainers",
 	    "appliances": "showAppliances",
@@ -919,12 +919,6 @@ $(document).ready(function() {
     var kernelInfo = function(data) {
 	    $loadavg.text('Linux: ' + data);
     };
-	
-    var displayLocaleTime = function(data) {
-
-        $('#local-time > span').text(data);
-
-    }	
 
     var displayLoadAvg = function(data) {
 	    var n = parseInt(data);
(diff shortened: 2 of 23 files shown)


MFlyer:
Current popup status on chrome/chromium - firefox: shell window seen like a popup so blocked by both browsers (allow popup for rockstor machine).
Alternative: make it like trigger a click event because this prevents popup blocks.

MFlyer:
Updates on popup blocks:
Triggering a click event fails (both Chrome/chromium and Firefox find it's not a real user click, so we get a popup block)
Test done removing ShowShell route, adding a 'popupper' id to shell link in navbars and mastering it via on click event: nicely open a popup without blocks, but this mess up router.js structure / unit logic (no approute, func for on click event and mess while fetching shellinabox service data)

Final solution (to avoid router.js "bad" coding): Added a tooltip telling users to allow Rockstor machine on popup blockers (on 0b0e998)

To @schakrava : nothing more to code on my side, ready to your review :+1:

src/rockstor/storageadmin/static/storageadmin/js/router.js
```diff
@@ -757,7 +758,30 @@ var AppRouter = Backbone.Router.extend({
...
+		//Special router function for shell
+		//We fetch shellinaboxd service model and check for detach option
+		//If not present or false we go with normal layout to maincontent
+		//otherwhise (popup enabled) we open it in a new detached window
+		var _this = this;
+		_this.shell_service = new Service({name: 'shellinaboxd'});
+		_this.shell_service.fetch({
+		success: function(collection){
+			var config = JSON.parse(collection.get('config'));
+			if ('detach' in config && config.detach){
```
schakrava: This is awesome @MFlyer! Nicely done. I did a quick test and noticed a minor issue. If the service is not configured and you click on System -> System-Shell, nothing happens in the UI and there's an error in the browser-console because the service config is null. It would be good UX to prompt the user if shellinabox is not ready when they click on there. For example, if it's not configured, then promt them to configure and if the service is not started, give them the option to do so.

MFlyer: You're right, gotcha! :+1: 
Had tests simulating a fresh new install, but without deleting config field vals so didn't see it, sorry, gonna solve asap :smiley:

src/rockstor/system/shell.py
```diff
@@ -0,0 +1,64 @@
...
+def update_shell_config(shelltype='LOGIN', css='white-on-black'):
+
+    fh, npath = mkstemp()
+    with open(npath, 'w') as tfo:
+        #Write shellinaboxd default config
+        tfo.write('# Shell In A Box configured by Rockstor\n\n')
+        tfo.write('USER=root\n')
+        tfo.write('GROUP=root\n')
+        tfo.write('CERTDIR=/var/lib/shellinabox\n')
+        tfo.write('PORT=4200\n')
```
schakrava: Can we read the port number from settings.py instead? If you like to shoot for perfection, you can read all four attributes from it :)

src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst
```diff
@@ -40,6 +40,7 @@
       <li><a href="#config-backup"><i class="glyphicon glyphicon-floppy-save"></i> Config Backups</a></li>
       <li><a href="#email"><i class="fa fa-envelope"></i> Email Alerts</a></li>
       <li><a href="#logsmanager"><i class="fa fa-gears"></i> Logs Manager</a></li>
+      <li><a href="#shell"><i class="fa fa-terminal"></i> System Shell</a></li>
```
schakrava: I agree that System navbar item is the right place for this. But what do you think of a small terminal icon right next to System Time? It looks nicer IMO, just an idea.

src/rockstor/storageadmin/static/storageadmin/js/router.js
MFlyer: Thinking about current first run of shellinaboxd we have some defaults settings for it (black on white and LOGIN connection as on shellinaboxd conf file). What do you think @schakrava about a prep_db.py change to full reflect defaults vals? Changing / adding a new func next to 
```python
def register_services():

File: src/rockstor/system/shell.py (hunk 2 above)
Comment (MFlyer) [inline src/rockstor/system/shell.py]: Direct conf info on /conf/settings.conf.in ? why not? :)

File: src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst (hunk 3 above)
Comment (MFlyer) [inline src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst]: Eheh here comes another problem I had to challenge with: Rockstor webUI available space

I was not sure about shell icon so got a "code" icon to have a different icon (from font awesome, open to suggestions !)
Current system menu is becoming too long (IMO) and you're right we could move shell icon next to system time, but on that bar we have hostname, management info and system time on left, kernel and uptime on right, so i'm not sure if it's better to have shell on system menu or on this bar... @priyaganti @phillxnet ??? :smiley:

Comment (schakrava) [inline src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst]: I think we should have it in under System. But in addition to that, it would be nice to have a single click access by providing a small icon on the horizontal bar right under the main navbar.

Comment (MFlyer) [inline src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst]: Ok, gonna add it :)

File: src/rockstor/system/shell.py (hunk 2 above)
Comment (MFlyer) [inline src/rockstor/system/shell.py]: Added on 007dac3 :)

File: src/rockstor/storageadmin/static/storageadmin/js/router.js (hunk 1 above)
Comment (MFlyer) [inline src/rockstor/storageadmin/static/storageadmin/js/router.js]: Default config got via prep_db on 852d737 and 41ca9cd

There's also a new dictionary (services_configs) everyone can use to set a default config field val on services table :)

Flyer

File: src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst (hunk 3 above)
Comment (MFlyer) [inline src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst]: Hi @schakrava , small icon on navbar via 5a015b4
If ok for you we can procede :smiley:

File: src/rockstor/scripts/prep_db.py [hunk 4]
```diff
@@ -37,12 +37,20 @@ def register_services():
         'NUT-UPS': 'nut',
         'ZTaskd': 'ztask-daemon',
         'Bootstrap': 'rockstor-bootstrap',
-        'Rockstor': 'rockstor',}
+        'Shell In A Box': 'shellinaboxd',
+        'Rockstor': 'rockstor'
+        }
+
+    services_configs = {
```
schakrava: +1 for this idea. After leaving my comments I was thinking that it may be better UX if we just assume default config and start the daemon if it's never been configured before and the user happens to access the shell via web-ui. But this method works too. Feel free to improve it further in a future PR. As far as I am concerned, this is impressive as is, very useful feature!

Issue #1392, opened 2016-07-12
Changing hostname does not update postfix config


greghensley:
After changing the hostname via the WebUI, the file `/etc/postfix/generic` still contains re-mapping directives for the old hostname.  As a result, mail is sent to the configured SMTP forwarder without remapping the sender.  Since the default sender is `<user>@<hostname>.localdomain`, most ISPs' SMTP forwarders will reject the message as the `.localdomain` TLD is not resolvable.

This should be considered silent failure, since users will simply stop receiving notifications.  The only ways for users to detect that a problem exists is by either sending a test email from the WebUI, or accessing the root account from a shell and checking its mail spool (postfix returns the undeliverable mail).

Removing and re-adding the email configuration in the Rockstor WebUI correctly updates the postfix config.

MFlyer:
Hi @greghensley working on it, probably going to solve it today

Flyer

PR #1393, opened 2016-07-13
Issue#1392 hostname change missing update on postfix

MFlyer:
To @schakrava Code adds / mods explained on 71920be for #1392 

Simply added a call to update_generic() on every hostname change (obviously only if email account on our table)

Ready to merge

Flyer


schakrava:
Thanks @MFlyer , nicely done as usual :)

PR #1395, opened 2016-07-15
Fixes #1391 - Changed how we check Active Directory status

### LINKED ISSUE #1391: Services page bottleneck on Active Directory status
Hi @schakrava , 
while looking again to services page loading i asked myself why it's slower then others so performed some checks on base_service page and found out this:

```
[12/Jul/2016 14:37:50] DEBUG [smart_manager.views.base_service:86] Sorting for Rockstor
[12/Jul/2016 14:37:50] DEBUG [smart_manager.views.base_service:86] Sorting for S.M.A.R.T
[12/Jul/2016 14:37:50] DEBUG [smart_manager.views.base_service:86] Sorting for Samba
[12/Jul/2016 14:37:50] DEBUG [smart_manager.views.base_service:86] Sorting for Active Directory <--
[12/Jul/2016 14:38:03] DEBUG [smart_manager.views.base_service:86] Sorting for NTP
[12/Jul/2016 14:38:03] DEBUG [smart_manager.views.base_service:86] Sorting for NUT-UPS
[12/Jul/2016 14:38:03] DEBUG [smart_manager.views.base_service:86] Sorting for SNMP
[12/Jul/2016 14:38:03] DEBUG [smart_manager.views.base_service:86] Sorting for SFTP
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for Rock-on
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for Replication
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for Shell In A Box
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for NFS
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for LDAP
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for NIS
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for AFP
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for ZTaskd
[12/Jul/2016 14:38:04] DEBUG [smart_manager.views.base_service:86] Sorting for Bootstrap
```

As you can see checking Active Directory status (net ads status) is our bottleneck so here my suggestion: avoid initial status check on collection.fecth because it's a kind of repetition with next listener from socket.io services (collect services status on first run, add socket.io listener and get statuses again: we can skip first check to immediately render page)

What do you think about this?

Flyer

MFlyer:
#1391 Previous check for active directory status was with net ads status
That was ok, but get too long, because enum objects too
Actually checking status via wbinfo -t (rpc call) and
wbinfo -a (authentication with admin user) and these 2 cmds together
are faster then net ads status

### DIFF
--- a/src/rockstor/system/services.py
+++ b/src/rockstor/system/services.py
@@ -33,6 +33,7 @@ SYSTEMCTL_BIN = '/usr/bin/systemctl'
 SUPERCTL_BIN = ('%s/bin/supervisorctl' % settings.ROOT_DIR)
 SUPERVISORD_CONF = ('%s/etc/supervisord.conf' % settings.ROOT_DIR)
 NET = '/usr/bin/net'
+WBINFO = '/usr/bin/wbinfo'
 AFP_CONFIG = '/etc/netatalk/afp.conf'
 
 
@@ -159,9 +160,18 @@ def service_status(service_name, config=None):
                            throw=False)
     elif (service_name == 'active-directory'):
         if (config is not None):
-            cmd = [NET, 'ads', 'status', '-U', config.get('username')]
-            return run_command(cmd, input=('%s\n' % config.get('password')),
-                               throw=False)
+            # 2 steps Active Directory status check
+            # First checks secret via rpc callable
+            # Second checks via auth for admin username
+            # If both give us 0 rc Active Directory is running
+            wbinfo_trust_cmd = [WBINFO, '-t', '--domain', config.get('domain')]
+            wbinfo_auth_credentials = '{}@{}%{}'.format(config.get('username'), config.get('domain'), config.get('password'))
+            wbinfo_auth_cmd = [WBINFO, '-a', wbinfo_auth_credentials, '--domain', config.get('domain')]
+            wbinfo_trust = run_command(wbinfo_trust_cmd, throw=False)
+            wbinfo_auth = run_command(wbinfo_auth_cmd, throw=False)
+            active_directory_rc = 0 if (wbinfo_trust[2] == 0 and wbinfo_auth[2] == 0) else 1
+            
+            return '', '', active_directory_rc
         # bootstrap switch subsystem interprets -1 as ON so returning 1 instead
         return '', '', 1
 


src/rockstor/system/services.py
```diff
@@ -33,6 +33,7 @@
 SUPERCTL_BIN = ('%s/bin/supervisorctl' % settings.ROOT_DIR)
 SUPERVISORD_CONF = ('%s/etc/supervisord.conf' % settings.ROOT_DIR)
 NET = '/usr/bin/net'
+WBINFO = '/usr/bin/wbinfo'
```
schakrava: just to double check, you've installed samba-winbind-clients package for /usr/bin/wbinfo, correct?

MFlyer: Hi @schakrava no manual installation for wbinfo, it was already there (probably from default samba install)

Flyer

schakrava:
I have another question for you @MFlyer . I agree that ```net ads status``` is costly. The man-page also suggests that it's for developers and suggests the usage of ```net ads testjoin``` instead.  So here's my question. How is your new method using the 2 wbinfo commands better than ```net ads testjoin```?

MFlyer:
Hi @schakrava nice question, as always!
Prior to decide, I had an empiric test with `net ads testjoin` vs wbinfo and it seems slower then 2 wbinfo commands (tests done without times measure, but i observed a longer waiting for `net ads testjoin`)
Except for that,  `net ads testjoin` grant us we've joined a domain, but doesn't tell us if we correctly read from domain (you can have a machine correctly joined to AD and not working on winbindd, instead if winbindd is ok reading/authenticating users that means your machine has joined domain too . A => B but B ≠> A

schakrava:
Thanks for the explanation @MFlyer. I think your/wbinfo method is better.

MFlyer:
Thanks @schakrava :) It's always a pleasure working on Rockstor! :+1:

Issue #1083, opened 2016-01-10
GMail detects Rockstor as a 'Less Secure' application


victoriasavi:
With the default setup Rockstor can't use GMail, login is not allowed.

> Jan 10 02:36:39 192 postfix/smtp[4285]: 180BFF29E: to=<my-email@gmail.com>, relay=smtp.gmail.com[74.125.136.108]:587, delay=0.92, delays=0.03/0.04/0.86/0, dsn=4.7.14, status=deferred (SASL authentication failed; server smtp.gmail.com[74.125.136.108] said: 534-5.7.14 <https://accounts.google.com/ContinueSignIn?loginlink[editted]> Please log in via your web browser and?534-5.7.14 then try again.?534-5.7.14  Learn more at?534 5.7.14  https://support.google.com/mail/answer/78754 73sm6146114wmm.7 - gsmtp)

This error turns you to this page: https://support.google.com/mail/answer/78754

After reading that I got to this support page: https://support.google.com/accounts/answer/6010255

Enabling the special gmail address for rockstor I made for 'Access for less secure apps' the issue was resolved and the test email now arrives.

Issue #978, opened 2015-10-22
Send test email reports success even if failed.


dragon2611:
It seems you will get the "Test e-mail sent successfully" dialogue in the web interface even if actually authentication to the SMTP relay failed completely.

The send test email function actually needs to validate the email was infact delivered successfully rather than just handing off to postfix and reporting success.

schakrava:
Thanks for reporting this @dragon2611. Do you know how to validate if the email was actually delivered?

dragon2611:
Possibly parse the logs to confirm it was accepted by the remote server?

Sent from Outlook Mobile<https://aka.ms/blhgte>



On Fri, Dec 11, 2015 at 7:14 PM -0800, "Suman Chakravartula" <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for reporting this @dragon2611<https://github.com/dragon2611>. Do you know how to validate if the email was actually delivered?

-

victoriasavi:
I had the same issue. I suggest looking into the log to find out what went wrong.

` tail -f /var/log/maillog `

For my case I made a new issue: #1083

Issue #738, opened 2015-07-14
Cannot edit existing scheduled task


therealprof:
I created a scheduled backup task but for whatever reason I'm unable to change it after the fact and instead I get the following error:

> Error! 
> Unknown internal error doing a GET to /api/sm/tasks/1 
> If you need more help on this issue, email us at support@rockstor.com with the following information:
> The error message: "Unknown internal error doing a GET to /api/sm/tasks/1"
> A step by step description of the actions leading up to the error.
> Download the tarred and zipped server logs by clicking here, and attach it to the email.

I have checked the logs but there's no evidence whatsoever actually pointing to this particular problem.

modelrockettier:
Is this a duplicate of #728?

therealprof:
Oh sorry, I checked the existing ones but missed that one.

PR #1402, opened 2016-07-20
Issue#1083 issue#978 testmail and gmail auth

MFlyer:
Reference to #1083 and #978 

To @schakrava 
Google accounts problem : added reference to google pages (how to allow access to unsecure apps)
"Mail sent" : modified with a kind message like "mail queued" (no grant about message arriving.

Except these 2 small things, our new smtp auth check function tell if entered smtp configs are ok ( try a real login to smtp server :wink: )

Note: this had to be done on config page because smtp.login requires smtp password, not stored on Rockstor db (so doing we don't store it and this is good)

Flyer

P.S.: Don't care about first commit reference to #738 - just a mistake

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/email/email.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/email/email.jst
@@ -104,7 +104,10 @@
   </div>
   <div class="alert alert-danger auth-results" id="auth_failed" style="display: none;">
   <strong>SMTP Authentication failed!</strong> Please update your settings before saving.
-  <br/>If using a Google account please remember to <a target="_new" href="https://support.google.com/a/answer/6260879">Allow less secure apps to access accounts</a>
+  <br/>If using a Google account please remember to enable
+  <a target="_new" href="https://support.google.com/mail/answer/185833">App Passwords</a> after 
+  <a terget="_new" herf="https://support.google.com/accounts/answer/185839">Google 2-Step Verification</a> or
+  <a target="_new" href="https://support.google.com/a/answer/6260879">Allow less secure apps to access accounts</a>
   </div>
 </div><!--panel-body-->
 </div><!--col-md-8 -->
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/email/email.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/email/email.jst
@@ -94,11 +94,20 @@
     <div class="form-group">
       <div class="col-xs-8 col-xs-offset-4">
         <a id="cancel" class="btn btn-default">Cancel</a>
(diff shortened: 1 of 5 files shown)


phillxnet:
@MFlyer Only had a quick look but in one of the commits so far you recommend that gmail's security setting requires lowering; last I knew this was not necessary but a work around to accommodate poor / old clients and we shouldn't need this (it is often done in error just to get things working). See my forum post for details, in there I reference a GitHub shell script that I used which at least way back when I tried it did not require lowering my gmail settings, I reference the script just as it contains the required configuration. Took me ages to find out too. I used it to automate the setup of the same mail server program as Rockstor uses for an Ubuntu server's notifications.
https://forum.rockstor.com/t/error-notifications-and-status-notifications-email/143/2
https://forum.rockstor.com/t/error-notifications-and-status-notifications-email/143/4

Well done on replacing those pesky TAB characters, we are getting there on those.

phillxnet:
@MFlyer It may be that the modifications necessary to support the more secure gmail requirements are outside the scope of this pr, in which case we can always open another issue to address what is necessary. And apologies if this is all nonsense now, but I did run into this and did solve it way back and so it would be good to implement the same in Rockstor if it is still a valid method.

Great stuff here as usual by the way.

MFlyer:
Hi @phillxnet , thanks for your checks!! I had same doubt about lowering gmail security and found your solution proposed on web too, but pls check current Rockstor `/etc/postfix/main.cf` that is running like your.

Possible solution: Do a real check for gmail accounts (not just a warning) and provide a custom function for that.

Flyer

MFlyer:
Good morning @phillxnet ,
had some testing over gmail/google starting from your github page :sunglasses: (i see it's from 2014) and actually gmail account access has become hard, with only 2 options:
- Enable less secure apps access
- Enable Google 2-step verification (every time you login will require pass+one time password via sms or similar) that let you add auth for other apps (create a special new password for your gmail account instead of one normally used)

Finally,
I think our current PR can be considered OK

phillxnet:
@MFlyer Yes very much advised to use special password and that is what I did. Very bad practice to use master password for gmail account in an appliance of any sort. I also had 2 factor enabled and as an appliance can't ask for the 2nd factor it is required to setup a special purpose password. It then doesn't have as many rights as the master password and can be used in circumstance where a second factor is not required. From memory that was my setup, ie 2 factor and special app password; which has been general practice on gmail for some time now. Nice as you can easily label the various 'apps' that have access and disable that particular access without messing up all other access points.

You new text is great. Sorry to be meddling here, your stuff is always good but it was just my little chip in as I think we don't want to be encouraging lowering security settings; as you mentioned yourself.

Cheers.

The real check login is a great improvement here, well done.

MFlyer:
Thanks @phillxnet :)

schakrava:
Nice work @MFlyer and thanks for your collaboration @phillxnet . LGTM.

Issue #1406, opened 2016-07-23
single to raid1 pool resize not reflected in Web-UI


phillxnet:
Thanks to @grebnek in the following forum thread for reporting this bug.
https://forum.rockstor.com/t/finally-build-my-own-nas/1811

Steps to reproduce (Rockstor version 3.8-14.07):-

Had a single disk pool with single raid level.

Resize Pool - Add Disk - Yes to change RAID level - selected raid1 from drop down and ticked additional disk - Next

Waited until balance complete.

Pools - rock-pool - indicates "Raid Configuration: single"
as does the pools column in the Pools page overview.

Expected result = indication of raid1 for this pool.

Confirmation of correct balance from single to raid1:
```
btrfs fi df /mnt2/rock-pool/
Data, RAID1: total=112.00GiB, used=110.03GiB
Data, single: total=1.00GiB, used=0.00B
System, RAID1: total=32.00MiB, used=48.00KiB
Metadata, RAID1: total=1.00GiB, used=390.86MiB
GlobalReserve, single: total=144.00MiB, used=0.00B
```

This replicates @grebnek findings.

phillxnet:
Please update the indicated forum thread with significant development on this issue.

grebnek:
Maybe I can share the results of my research with you, to help a little bit.
The Problem seems to be the second Data block in the output of btrfs fi df /mnt2/pool.

The function pool_raid(mnt_pt) in fs/btrfs.py iterates over this output 
and only use the last value of the data block.

These single data chunks seem to be normal, according to btrfs [FAQ](https://btrfs.wiki.kernel.org/index.php/FAQ) and I could get rid of them with another balance. I still have to try it.
see "Why do I have "single" chunks in my RAID filesystem?"

phillxnet:
@grebnek Well done. I was about to try another balance here on the system I setup to replicate your findings so I will report back shortly. My current task is actually to take a closer look at all the functions within btrfs.py so your report is actually very timely. I'll post here if it worked for me, obviously we need to parse better and I will keep this in mind when I get to that function in the unit testing I am ear marked to do there.

phillxnet:
[...]

PR #1408, opened 2016-07-26
issue#1406_single_to_raid1_not_reflected

grebnek:
Fix for Issue [Issue 1406](https://github.com/rockstor/rockstor-core/issues/1406)

After a change of the raid level the poll in the Rockstor UI reflects the wrong, old raid level.
pool_raid(mnt_pt) in fs/btrfs.py iterates of the output over the output of btrfs fi df and only returns
the raid level of the last data block. This block is unused and will be removed if another balance is started

```
btrfs fi df /mnt2/rock-pool/
Data, RAID1: total=112.00GiB, used=110.03GiB
Data, single: total=1.00GiB, used=0.00B
System, RAID1: total=32.00MiB, used=48.00KiB
Metadata, RAID1: total=1.00GiB, used=390.86MiB
GlobalReserve, single: total=144.00MiB, used=0.00B
```
Updated pool_raid(mnt_pt)  to only use the raid of the first block
and to ignore 'DUP' if metedata is duplicated on a single device.
```
btrfs fi df /
Data, single: total=2.02GiB, used=1.58GiB
System, DUP: total=8.00MiB, used=16.00KiB
System, single: total=4.00MiB, used=0.00B
Metadata, DUP: total=343.50MiB, used=80.80MiB
Metadata, single: total=8.00MiB, used=0.00B
GlobalReserve, single: total=32.00MiB, used=0.00B
```

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -99,7 +99,10 @@ def pool_raid(mnt_pt):
     for l in o:
         fields = l.split()
         if (len(fields) > 1):
-            raid_d[fields[0][:-1].lower()] = fields[1][:-1].lower()
+            block = fields[0][:-1].lower()
+            raid = fields[1][:-1].lower()
+            if not block in raid_d and raid is not 'DUP':
+                raid_d[block] = raid
     if (raid_d['metadata'] == 'single'):
         raid_d['data'] = raid_d['metadata']
     return raid_d


phillxnet:
@grebnek Nice, well done, and thanks.

@schakrava With the above changes applied the pool_raid() function now passes the as yet un-merged unit test for the same.
I.e. before we had an existing FAIL:
"AssertionError: get_pool_raid_level() miss identified raid level raid1_some_single_chunks"
after we have a pass for this and all existing raid level identification test data.

(The test logs the proposed new name.)

phillxnet:
@grebnek I have reviewed your pr and as per your prompt in the pr description added a further test case to account for the default system disk format of single data and dup metadata ie:-
```
btrfs fi df /mnt2/rockstor_rockstor
Data, single: total=3.37GiB, used=2.70GiB
System, DUP: total=8.00MiB, used=16.00KiB
System, single: total=4.00MiB, used=0.00B
Metadata, DUP: total=471.50MiB, used=165.88MiB
Metadata, single: total=8.00MiB, used=0.00B
GlobalReserve, single: total=64.00MiB, used=0.00B
```
As it turns our you have also uncovered another bug as previous to your patch pool_raid() returned:
```
{'data': 'single', 'system': 'single', 'globalreserve': 'single', 'metadata': 'single'}
```
Where the data raid level is correct but the system and metadata are not. As it goes this doesn't have any knock-on as we only currently use the data raid level (as far as I'm aware).
After you patch the return for the same is:
```
{'data': 'single', 'system': 'dup', 'globalreserve': 'single', 'metadata': 'dup'}
```
Which looks right, again as a result of only taking the first 'block' entry.

But there is a part of your patch that has no effect, however from my testing so far all is working ok anyway.
ie the additional clause of:
```
and raid is not 'DUP'
```
but the 'raid' variable you are testing is always going to be lower case (.lower()) and so this is always true.

I have tried running the existing test data of all raid levels and as is just taking the first entry seems to have solved the existing known issues.

I propose that you change:
```
if not block in raid_d and raid is not 'DUP':
```
to a simpler
```
if block not in raid_d:
```
Which also addresses a PEP8 membership test syntax issue and gets you another commit of course :).

Sorry if I've missed something here re the DUP clause, if so could you explain what you intentions with that clause were.

schakrava:
Hello @grebnek . Thanks for this pr, especially given that it requires interpretation of still-evolving output format of btrfs-progs. Welcome to the club of code contributors! Happy to merge.

PR #1411, opened 2016-07-28
fixes #1410 - Added missing js dependency

### LINKED ISSUE #1410: Email Alerts page missing dependency
To @schakrava : starting from the title, issue related to missing dependency over latest mods (submit not working)

Flyer

MFlyer:
@schakrava Finally solves missing dependency over #1410 - Email Alerts

Tested, just need merge

Flyer

### DIFF
--- a/src/rockstor/storageadmin/templates/storageadmin/base.html
+++ b/src/rockstor/storageadmin/templates/storageadmin/base.html
@@ -112,7 +112,7 @@
     <script src="/static/js/lib/jquery.dataTables.min.js"></script>
     <script src="/static/js/lib/dataTables.bootstrap.min.js"></script>
     <script src="/static/js/lib/clipboard.min.js"></script>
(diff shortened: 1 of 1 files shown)


phillxnet:
@MFlyer Just a little nit pick here but it looks like we are all accidentally popping in tabs by accident occasionally. Just mentioning as @schakrava has picked me up on the same in the past. I believe the project convention is only spaces (usually in groups of 4 but some old files use only 2) but no tabs.

Well done on taking owning of this one by the way, and the quick fix.

MFlyer:
Hi @phillxnet had it "on the fly" so probably with wrong tabs! :/

schakrava:
looks good @MFlyer . Thanks for taking care of this. I should've caught this in the first place :)

MFlyer:
No problem @schakrava , to be honest i thought jslibs were managed something like js views -> rockstor.js (dependencies solved/script tags added on build process), so now I've learned how to handle it ;)

Issue #1412, opened 2016-07-30
Share usage not reported correctly while pool usage is


sfranzen:
I have one share on a 2-disk simple btrfs volume. In the shares view, its usage is reported as 16.00 KB, while the pools view correctly lists 12.99 GB in use.

Note: if this is relevant, the share in question is /home. I've tinkered with my setup a bit by moving /home from its default place on the USB stick with the OS, and allowing it to be shared so I can use samba's [homes] section.

Btrfs details:

```
[root@rockstor ~]# btrfs filesystem usage /home
Overall:
    Device size:                 298.10GiB
    Device allocated:             17.03GiB
    Device unallocated:          281.07GiB
    Device missing:                  0.00B
    Used:                         12.99GiB
    Free (estimated):            283.10GiB      (min: 283.10GiB)
    Data ratio:                       1.00
    Metadata ratio:                   1.00
    Global reserve:               16.00MiB      (used: 0.00B)

Data,single: Size:15.00GiB, Used:12.96GiB
   /dev/sda        7.00GiB
   /dev/sde        8.00GiB

Metadata,single: Size:2.00GiB, Used:27.02MiB
   /dev/sda        1.00GiB
   /dev/sde        1.00GiB

System,single: Size:32.00MiB, Used:16.00KiB
   /dev/sda       32.00MiB

Unallocated:
   /dev/sda      141.02GiB
   /dev/sde      140.05GiB
```

sfranzen:
Did some poking around and discovered that this is due to the quirks of btrfs. In particular: I had experimented with btrfs quota as a way of putting a limit on user directory sizes. This means that my /home directories were actually subvolumes, with their own qgroup. Since rockstor currently determines a share's size by looking up its ID in the output of `btrfs qgroup show <share_pool>`, in this case it only reports the (negligible) usage of the /home subvolume, not of any child subvolumes.

I have resolved this for now by placing the user files back into regular directories and deleting the subvolumes, but this raises the question of how to correctly report the share usage in such a use case. Unfortunately, btrfs does not provide straightforward ways of doing so at the moment. The most robust way seems to be `btrfs filesystem du -s <path>`, but this will be extremely slow for shares with many files. Any thoughts?

sfranzen:
[...]

PR #1415, opened 2016-08-02
Adjust share usage reporting

sfranzen:
A possible way of dealing with #1412. This extends the `share_usage` function by accounting for data located in subvolumes of the share's path.


schakrava:
Thank you for this pr @sfranzen and welcome on board as a code contributor!
Sorry it took me a while to get to this, but LGTM!

PR #1417, opened 2016-08-05
Fixes #1252 adding samba shadow localtime param

### LINKED ISSUE #1252: Windows previous version tab: Time offset
You decided to save the automatic share snapshots in a format like

`.share_201604091525`

The timestamp format is local time, as configured during the installation.
By default, samba is set to expect UTC here, as described in the doc here:

https://www.samba.org/samba/docs/man/manpages/vfs_shadow_copy2.8.html

This leads to an time offset in the windows prevous version tab by the differnce of the user local time including DST.

Adding 
`shadow:localtime = yes`
to each share entry in smb.conf fixes this. Please add this entry.

Anyway i would recommend to use the default naming "@GMT-%Y.%m.%d-%H.%M.%S"
for the snapshots and  do the formatting in the GUI, with timestamp timezone translation.
Having timestamps in local time will lead to further problems if the user changes the machine timezone and so on. Thats probably the reason for v_shadow_copy's default settings.

MFlyer:
Related to #1252 

My opinion to @phillxnet and @schakrava 
Having all in UTC/GMT probably would be the perfect solution, but solving just adding localtime param it's easier and less expensive:

- moving to gmt probably would require some snapshots code rewrite
- having old snapshots with local time and new ones with gmt will mess all (think about this pr like a "conservative patching" letting us not to write a complex script to allign old snaps format with new snaps)

Flyer

### DIFF
--- a/src/rockstor/system/samba.py
+++ b/src/rockstor/system/samba.py
@@ -70,6 +70,7 @@ def rockstor_smb_config(fo, exports):
             fo.write('    shadow:basedir = %s\n' % e.path)
             fo.write('    shadow:snapdir = ./\n')
             fo.write('    shadow:sort = desc\n')
+            fo.write('    shadow:localtime = yes\n')
             fo.write('    vfs objects = shadow_copy2\n')
             fo.write('    veto files = /.%s*/\n' % e.snapshot_prefix)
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks for the pr @MFlyer!

Issue #1331, opened 2016-05-29
Rockon logging


Jaco1960:
The rockon dockers are started with the syslog logging option. This prevents analyses of the docker logs. Can you change this to json-file or journald?

phillxnet:
@schakrava This looks like it is due for closure as Fixed by @sfranzen pr #1420
@Jaco1960 You are welcome to close this issue as such if you agree.

PR #1420, opened 2016-08-09
Docker journald logging

sfranzen:
See #1331. Run the docker service with journald log driver. Also fixes the `-d` argument that will become deprecated.

Unfortunately, this does not change logging for currently installed rock-ons as the `docker run` command is only issued upon installation; re-installation is required.


schakrava:
Thank you @sfranzen . This is a welcome pr, especially given that we should be moving to a newer version(I've been using 1.11) of docker-engine soon.

PR #1423, opened 2016-08-19
Fix: argument to docker run should still be -d

sfranzen:
@schakrava Important, please merge/release ASAP. I overlooked that the command in this case was `docker run`. Passing `daemon` to this will cause it to take this as the image name, failing to install/run. Not sure why this did not appear while I tested this.


fago:
I just ran into this in version: 3.8-14.10
Error caused by this:
>Error: image library/daemon:latest not found

fago:
Note: I manually patched rockon_helpers.py and verified that this fixes the problem.

schakrava:
ok finally got around to testing this out. Thanks @sfranzen and @fago!

PR #1437, opened 2016-09-21
Issue#998 dashboard crashes if opened long - Memory Widget - Part 3

MFlyer:
Hi all developers,
here is Part 3 covering issue #998 

It was quite easy and now our dashboard (Part 1, Part 2, Part 3 PRs) seems ok without crashing / freezing :smiley: 

Going on with other wdgets

Flyer


schakrava:
Thank you @MFlyer! looks great combined with your new Chart.js

PR #1439, opened 2016-09-22
Issue#998 dashboard crashes if opened long - Top Shares Widget - Part 4

MFlyer:
Hi @schakrava, 
our #998 saga goes on and here we have a working top shares widget without using d3 or chart.js, just with some js code and Bootstrap Progressbars :smile: 

Ready to merge
Flyer


schakrava:
I really like the new look. The way progress bars load looks great. The look is more crisp, would be nice to see the same crisp-look with mem and cpu widget :) Thanks @MFlyer

Issue #1438, opened 2016-09-21
Error when creating an AFP share and the service is not running


sfranzen:
Forum user Maxime submitted [a bug report](https://forum.rockstor.com/t/solved-error-when-i-create-a-afp-share/2088) with the following error:
```
Traceback (most recent call last):
File "/opt/rockstor/src/rockstor/storageadmin/views/netatalk.py", line 134, in post
systemctl('netatalk', 'reload')
File "/opt/rockstor/src/rockstor/system/services.py", line 64, in systemctl
return run_command([SYSTEMCTL_BIN, switch, service_name])
File "/opt/rockstor/src/rockstor/system/osi.py", line 104, in run_command
raise CommandException(cmd, out, err, rc)
CommandException: Error running a command. cmd = ['/usr/bin/systemctl', 'reload', 'netatalk']. rc = 1. stdout = ['']. stderr = ['Job for netatalk.service invalid.', '']
```

Instead, a more friendly notification should be displayed that the service is not running, just like for the other share types.

Issue #813, opened 2015-08-23
Add AFP export throws error when the service is turned off 


priyaganti:
1. Rockstor version: 3.8-5.11

When the service is turned ON, I am able to add the afp export without any issues.
![afpexport](https://cloud.githubusercontent.com/assets/1999710/9426384/22d2c6d6-48f2-11e5-9975-8dec3665dfb2.png)

sfranzen:
@schakrava This one can be closed as well now.

PR #1442, opened 2016-09-22
Fix DataTables error on AFP shares view

sfranzen:
This was caused by the table displayed in the absence of existing shares being classed `data-table` and stopped the view from loading fully. Partial fix of #1438, which also involves #813.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/afp/afp.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/afp/afp.jst
@@ -39,7 +39,7 @@
         </tbody>
       </table>
       {{else}}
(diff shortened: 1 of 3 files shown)


sfranzen:
This now fixes the other issue and performs a systemctl `reload-or-restart` upon adding a new share. Perhaps this is a more intuitive UX for the other share types as well?

sfranzen:
@schakrava Good idea. My IDE parser no longer accepts python 2, so it's marking these as invalid syntax. Looks like it's mostly "Exception, e" that it's complaining about though.

priyaganti:
Hey @sfranzen Thanks for finding out the issue and fixing it. Very good catch. I didn't notice it.

Issue #1407, opened 2016-07-26
Broken Samba Config


barnhill:
The following error is filling up my smb logs (1 about every 20 secs or so):

[2016/07/25 03:58:06.044671,  0] ../source3/printing/print_cups.c:529(cups_async_callback)
  failed to retrieve printer list: NT_STATUS_UNSUCCESSFUL

Issue #1404, opened 2016-07-21
Broken Samba again


MFlyer:
Hi all,
current env: testing machine with samba and AD integration
after recent changes something is going wrong.
Here a part of my smb.conf file:
```
[global]
    workgroup = SIVAG
    log file = /var/log/samba/log.%m
    security = ads
    realm = sivag.local
    template shell = /bin/sh
    kerberos method = secrets and keytab
    winbind use default domain = false
    winbind offline logon = true
    winbind enum users = yes
    winbind enum groups = yes
    idmap config * : backend = tdb
    idmap config * : range = 1000000 - 1999999
    idmap config SIVAG : backend = rid
    idmap config SIVAG : range = 10000 - 999999
    log level = 3
    load printers = no
    cups options = raw
    printcap name = /dev/null
    map to guest = Bad User
    username = ****** <- NO!
    winbind use default domain = false
    idmap config * : backend = tdb
    winbind enum users = yes
    winbind offline logon = true
    idmap config SIVAG : range = 10000 - 999999
    password = ******* <- NO!
    kerberos method = secrets and keytab
    realm = sivag.local
    idmap config * : range = 1000000 - 1999999
    idmap config SIVAG : backend = rid
    winbind enum groups = yes
    template shell = /bin/sh
    security = ads
####BEGIN: Rockstor SAMBA CONFIG####
[storage]
    comment = Samba-Export
    path = /mnt2/storage
    browseable = yes
    read only = no
    guest ok = yes
####END: Rockstor SAMBA CONFIG####
```
Obviously replaced username and password (config infos for active directory administrator user) fields with *, but those are not samba fields and i get a regular error via email alerts from cron
```
/etc/cron.hourly/0yum-hourly.cron:
Unknown parameter encountered: "password"
Ignoring unknown parameter "password"
```
Some key pairs are duplicated
Addendum: if AD on you can't add smb.conf custom config because there's a control over setting workgroup name while on AD (both on same page so workgroup blocks custom global section values)

Flyer

MFlyer:
To @schakrava : not investigating on this, think related to latest #1383 new smb.conf file construction (I can check this, but i think you'll do it better knowing latest changes)

Maybe related only to dev environment??

Flyer

MFlyer:
To @schakrava : it seems there's some mess with AD enabled and custom settings (without looking at code: custom settings replicates ad settings in global - check my first post)

Flyer

MFlyer:
[...]

PR #1453, opened 2016-09-29
Broken samba config - correctly build smb.conf

MFlyer:
Opening PR to have a single space for code comments
Reference to #1404 #1407 
To @schakrava : not ready to merge, will tell when ready :)

Roadmap:
- [x] Let our Rockstor smb builder know how to loop inside current smb file like on smb shares, better define AD section, default global section and a global custom section (only one to iterate on with custom configs) (1e5e6fd)
- [x] Ensure skipping AD username and password fields in smb config file :exclamation: (1e5e6fd)
- [x] Nicely handle having custom global configs with AD ON too (currently not possible because of a check over workgroup name - not possible with AD ON, so not able to save custom config on smb [global] (1e5e6fd)
- [x] Samba config page - Avoid having workgroup both in workgroup field and custom params field (42dc6f2)
- [x] If AD ON disable workgroup field on samba config page (not really required, workgroup value by server side checks, just for a better UX) (7daefb5 to 9919b42)
- [x] Code cleaning and indent 4 spaces (42dc6f2)

Testing branch:
- [x] Random turn off/on samba and AD
- [x] AD ON add custom samba params
- [x] AD ON change workgroup, expected to have AD workgroup
- [x] AD OFF change/set workgroup on samba conf page and have it
- [x] AD OFF add custom samba params
- [x] Add / remove new samba shares
- [x] Samba config page workgroup field switches enable/disabled on AD current status

### DIFF
--- a/src/rockstor/smart_manager/views/active_directory.py
+++ b/src/rockstor/smart_manager/views/active_directory.py
@@ -219,6 +219,7 @@ class ActiveDirectoryServiceView(BaseServiceDetailView):
                     cmd += ['--update', '--enablelocauthorize',]
                     run_command(cmd)
                 config['workgroup'] = self._domain_workgroup(domain, method=method)
+                self._save_config(service, config)
                 update_global_config(smb_config, config)
                 self._join_domain(config, method=method)
                 if (method == 'sssd' and config.get('enumerate') is True):
--- a/src/rockstor/smart_manager/views/samba_service.py
+++ b/src/rockstor/smart_manager/views/samba_service.py
@@ -19,13 +19,14 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 import shutil
 from rest_framework.response import Response
 from storageadmin.util import handle_exception
-from system.services import systemctl
+from system.services import systemctl, service_status
 from system.samba import (update_global_config, restart_samba,
                           get_global_config)
 from django.db import transaction
 from django.conf import settings
 from base_service import BaseServiceDetailView
-from smart_manager.models import Service
+from active_directory import ActiveDirectoryServiceView
+from smart_manager.models import Service, ServiceStatus
 from storageadmin.models import SambaShare
 from system.osi import md5sum
 from smart_manager.serializers import ServiceStatusSerializer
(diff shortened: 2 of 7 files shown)


MFlyer:
Updates on last coding session:

Main issue was to handle `workgroup` to be choosen between AD workgroup or samba workgroup (manually added to samba conf page)

Current state : if AD is ON don't let custom samba config (https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/smart_manager/views/samba_service.py#L55-L61)

```python
        if (command == 'config'):
            aso = Service.objects.get(name='active-directory')
            if (aso.config is not None):
                e_msg = ('Active Directory service is configured, so '
                         'Workgroup is automatically retrieved and cannot '
                         'be set manually')
                handle_exception(Exception(e_msg), request)
```
Needed new workgroup selection race conditions over samba config page:
If AD is configured and AD is ON (both required, because we can have an old AD config, but AD OFF) then use AD workgroup like samba workgroup, else use user added workgroup.
Main issue: define when AD is ON (not a simple systemctl service), so neded service_status from system/services

MFlyer:
Hi @schakrava , PR is ready to merge.

Finally no func moving from views to samba.py, had a tricky solution (AD status not so easy to collect because not a real 'service' like smb, nmb etc etc, so had to pass current ad services status between views over Backbone router - that was fan because of trying to use backbone router optional params, available from backbone 0.9.9, and i finally realized we're on 0.9.2 after 25-30 mins eheh)

Mirko

schakrava:
Nicely done, as always @MFlyer . Thank you!

I am not able to test all of AD right now, but things look good as far as my testing goes. Can you also describe your testing effort?

MFlyer:
Hi @schakrava ,
checking on first PR message there's a task list related to PR testing:

First checks done switching AD to simple samba / simple samba to AD while having lot of shell cats over samba/smb.conf to check all was working, with a special care about removing USERNAME and PASSWORD vals used on AD from samba custom params
Second test group was on samba custom params and workgroup field:
before this PR Rockstor with AD active wasn't able to accept custom params because of workgroup field, so had to remove workgroup python checks and had some hacks over workgroup selection (if AD is on AD workgroup wins over samba config page workgroup)
Last part was about beautify and add another check over workgroup: Rockstor had workgroup repeated in samba custom params box, so had a new handlebars helper to avoid it, plus (not required, but it's a better way to inform users) workgroup field disable + info box if AD is on.

Huge part of this?? Define when AD is ON or OFF, because it's not a systemctl service and overlaps with simple samba too, so had a long testing over it

Issue #1152, opened 2016-02-05
Dashboard Disk activity widget's writes check box is not persistent


priyaganti:
Issue pointed out by user on forum: 
http://forum.rockstor.com/t/check-writes-by-default-when-refreshing-dashboard/1021

Issue #916, opened 2015-10-02
RFE: Add "all" option to disk activity widget 


roweryan:
Would be really cool :+1:

roweryan:
👍

Issue #831, opened 2015-08-25
Improve labels on dashboard widgets


vinima:
1. Network can be renamed to "Network Activity"
2. CPU to "CPU Usage"
3. Memory to "Memory Usage"

Issue #807, opened 2015-08-21
Dashboard widgets not entirely functional in Firefox


schakrava:
Bunch of helpful errors in the console.

PR #1457, opened 2016-10-02
Issue#998 dashboard crashes if opened long - Disks Widget - Part 7 (last chapter)

MFlyer:
Hi all,
obviously thi PR is not ready to merge, but required to have a place to discuss about Disk Widget.

Looking to disks widget i thought about usefulness of Top disks part of the widget (that one on top, always visible), with that bar blinking over disk reads/writes, so decided to try another way:

I'd like to have your opinions about this possible solution: http://jsfiddle.net/5gHVY/989/

Maybe having all disks stats there (so deleting disk details line graphs only available when widget is maximized) or have only some disks stat + old line graphs.

@phillxnet ?? - Ask you cos' your our "Smart Disks Master" :smile: 

**Edit on 07/10/2016 - Adding references to issues this PR will close**:
#998 
#1152
#919
#1065 <- @phillxnet can we consider this closing with latest disks updates?
#916
#831
#807
#589

### DIFF
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -419,6 +419,7 @@ class DisksWidgetNamespace(BaseNamespace, BroadcastMixin):
     def send_top_disks(self):
 
         def disk_stats(prev_stats):
+            disks_stats = []
             # invoke body of disk_stats with empty cur_stats
             stats_file_path = '/proc/diskstats'
             cur_stats = {}
@@ -455,23 +456,23 @@ class DisksWidgetNamespace(BaseNamespace, BroadcastMixin):
                         else:
                             datum = (float(cur[i]) - float(prev[i]))/interval
                         data.append(datum)
-                    self.emit('diskWidget:top_disks', {
-                        'key': 'diskWidget:top_disks', 'data': [{
-                            'name': disk,
-                            'reads_completed': data[0],
-                            'reads_merged': data[1],
-                            'sectors_read': data[2],
-                            'ms_reading': data[3],
-                            'writes_completed': data[4],
-                            'writes_merged': data[5],
-                            'sectors_written': data[6],
-                            'ms_writing': data[7],
-                            'ios_progress': data[8],
-                            'ms_ios': data[9],
-                            'weighted_ios': data[10],
-                            'ts': str(datetime.utcnow().replace(tzinfo=utc).isoformat())
-                        }]
-                    })
+                    disks_stats.append({
+                        'name': disk,
+                        'reads_completed': data[0],
+                        'reads_merged': data[1],
+                        'sectors_read': data[2],
+                        'ms_reading': data[3],
+                        'writes_completed': data[4],
+                        'writes_merged': data[5],
+                        'sectors_written': data[6],
+                        'ms_writing': data[7],
+                        'ios_progress': data[8],
+                        'ms_ios': data[9],
+                        'weighted_ios': data[10],
+                        'ts': str(datetime.utcnow().replace(tzinfo=utc).isoformat())
+                        })
+
+            self.emit('diskWidget:top_disks',{ 'key': 'diskWidget:top_disks', 'data': disks_stats })
             return cur_stats
 
         def get_stats():
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/dashboard/widgets/disk_util_select.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/dashboard/widgets/disk_util_select.jst
@@ -1,4 +1,4 @@
-<div id="disk-details-ph">
+<div id="disk-details-ph" >
   <div id="disk-select-ph">
     <span class="h4">Disk activity</span> for disk
     <select name="disk-select" id="disk-select" class="inline">
@@ -7,6 +7,5 @@
       {{/each}}
     </select>
   </div>
-  <div id="disk-graph-reads-ph"></div>
-  <div id="disk-graph-data-ph"></div>
+  <canvas id="single-disk-chart" width="500" height="180"></canvas>
 </div>
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/dashboard/widgets/disk_util_select.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/dashboard/widgets/disk_util_select.jst
@@ -1,11 +1,11 @@
-<div id="disk-details-ph" >
-  <div id="disk-select-ph">
-    <span class="h4">Disk activity</span> for disk
-    <select name="disk-select" id="disk-select" class="inline">
-      {{#each disks}}
-      <option value="{{this.name}}">{{this.name}}</option>
-      {{/each}}
-    </select>
-  </div>
-  <canvas id="single-disk-chart" width="500" height="180"></canvas>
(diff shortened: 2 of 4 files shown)


phillxnet:
@MFlyer That demo looks the business and a nice framework for multi factor info presentation. What were the disk stats you were considering for the axis ie the DiskStat1, DiskStat2? as we would need at least a few for this presentation to pay off. It would scale much better but I still think we are going to need a limit on the number of drives we display for, or maybe toggles on what drives we want to see from a list of all drives which might only be an option when in expanded mode due to size.

Well done on re-thinking this one by the way, I'm certainly in favour of this presentation change as then one could see if disk activity is more iops based or more raw bandwidth use; ie few but large writes, and to separate writes and reads which I think is the key advantage of this more multi factor simultaneous presentation.

How about splitting the historical info of the disk activity out into another widget as that way we avoid overloading this one again. Ie your idea for presenting a number of simultaneous and transient stats for instant nature of use assessment but another widget for purely historical data such as the graphs. But the split might be overkill, the historical info of the graphs does serve a different purpose and eases the display/interpretation of transient yet repeated access patterns that could more easily be lost in the multi simultaneous factor presentation idea.

sfranzen:
For me, this is quite a lot of information to digest from a single graph. I think I would prefer separate gauges for the instantaneous data, in the form of horizontal/vertical bars, or perhaps donut charts, along with a scrolling chart for the recent history.

Since I've been working on PatternFly stuff, the outer widget view could look something [like this](https://jsfiddle.net/sfranzen/a9fx318h/), where I've also put in a separate for longer term history. I don't think we currently record such longer time series, but it may be a nice addition.

MFlyer:
Hi all @phillxnet and @sfranzen ,
agree with you about not having too much data over that graph so we have to decide on these
```
reads_completed <- already on graph
reads_merged
sectors_read <- already on graph *512 to get Kb on historical
ms_reading <- candidate?
writes_completed <- already on graph
writes_merged
sectors_written <- already on graph *512 to get Kb on historical
ms_writing <- candidate?
ios_progress
ms_ios
weighted_ios
```
IMHO reads/writes merged not really interesting in this graph
Having historical infos on another widget not a good idea: both top disks and historical infos datas are from same socket.io [(connection to disks data_collector)](https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/storageadmin/static/storageadmin/js/views/dashboard/disk_utilization.js#L31-L32), so separating them means having 2 websockets to same service serving same data.

Other way to have some useful data: move to `iostat`, having some new code on backend too

Open to any suggestion!

sfranzen:
@MFlyer Yes, it would have to be a little bit more sophisticated, maybe a list of average I/O with one data point per day. I found some code apparently intended to record timeseries data for pool usage [here](https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/smart_manager/models/pool_usage.py). It seems unused so far, but I think such longer term data would be a nice additional feature for the dashboard.

MFlyer:
Hi @sfranzen that part is about Pools and not properly disks stats so not sure about having that infos on Disks Widget

sfranzen:
Yeah, meant to say to implement a similar model for disk stats.

MFlyer:
Hi all @rockstor/developers @sfranzen and @tomtom13,
coding on last widget / Disk widget had some checks with our open issues and added a cross ref in PR first message.
Can you pls confirm (or not) closing of those issues and/or check for other issues to close? thanks :smiley:

MFlyer:
Hi all,
having Rockstor moving soon to a new stable release (@schakrava last post on our forum), going to code Disk Widget next 2-3 days.
We'll care about exposing new/differente disks data in a future PR.
M.

schakrava:
Sounds good @MFlyer . Let me know when this is ready to be reviewed. Thank you!

MFlyer:
Backend mods required too - data_collector DisksWidgetNamespace working wrong way.
Emits on every single disk and not entire disk collection, so until today we had a "not at all" topdisks graph (my example over 3 disks: on first disk emit computes topdisk over first_disk@now and other two_disks@now-1, on second disk emit topdisk is on disk_1@now, disk_2@now, disk_3@now-1)

Not a problem at all because emits are nearly instantaneous, but conceptually wrong with lots of _.each cycles over 1 object / time

M.

MFlyer:
Hi all,
had fun with latest commits and last part about historical data (line charts) is quite easy, but need your opinions about data collecting:
currently our data_collector emits disks data every 1 sec and so doing we have not really useful data (data deltas on 1 sec are really small and not interesting at all).
Disk Widget historical data covers 300 shots, so on 1sec -> 5 min
Moving to 5 secs updates -> 25min history
Moving to 10 secs updates -> 50min history

Do you agree moving to 5secs updates or better to a 10secs updates?

M.

sfranzen:
The best option IMO would be to offer various update interval choices through a dropdown select. I haven't really looked at the socket.io server side stuff, but I'm sure the interval can be changed one way or the other.

Doesn't have to be all in a single pull request of course :wink:, just choose for example 5 seconds for now.

phillxnet:
@MFlyer Sorry just seen your edit / update here:
re "#1065 <- @phillxnet can we consider this closing with latest disks updates?"

I don't think we can close that as of yet because the given examples in the issue relate to reporting disk activity on md devices which as indicated is not reported in our base data (/proc/diskstats). We can always refine that at a later date as it will require sourcing, collating, and surfacing the data from the member drives.

All other issues referenced in your original post look relevant except #589 which I can't comment on currently.

Nice bit of bug / issue collation there by the way, looking forward to this one.

MFlyer:
New widget working, having more tests

Fixes #998 - Closing main issue
Fixes #1152 - No more checkboxes, opening a more generic issue about persistent dashboard opts data
Fixes #919 - Added enough IO infos over charts
Fixes #916 - Best draft algorithm grants having all possible info, radar chart has a pure visualization function while single disk chart give more details
Fixes #831 - Not required at all
Fixes #807 - No more "g" errors, related to svg elements (no more used)

MFlyer:
Hi @schakrava ,
with 	b65e145 PR is ready to merge/review :smiley: 

Here a small preview
immagine

Note: finally had no changes to data_collector sleep, had it over a 5 secs update and IMO too slow, so left it 1 sec

Mirko

phillxnet:
@MFlyer That looks the business, well done.
Have you tested the widget's behaviour when requested to display the information from a 'detached' device / disk?

MFlyer:
> Have you tested the widget's behaviour when requested to display the information from a 'detached' device / disk?

:flushed: 
@phillxnet
Ok, let's do some 'on the fly' talking tests:
starting with 3 disks -> disks collection = 3 disks
disk detached -> disks collection not updated so disks 3 while 2

- on update we compute only received data, so TopDisks will always be ok (detached disk won't get any update) :+1: 
- on update if widget maximized we render single disk data, worst case is selected disk == detached disk, but...
```javascript
        var current_disk = _this.disksData[_this.selectedDisk];
        if (current_disk) {
```
so in our worst case user won't have single disk chart updating, but without errors :grin: 

Mirko

schakrava:
@MFlyer the new widget looks really neat. I can't wait to see how it looks on my production systems. Very creative, thank you for a quality pr.

MFlyer:
Hi @schakrava , hope you'll enjoy it :)

PR #1460, opened 2016-10-05
Improve pool usage reporting

### LINKED ISSUE #1454: Pool usage misreported
While testing the coloured progress bars for the dashboard, I filled one pool ("pool1") entirely with data:

```
[root@rockstorvm share1]# dd if=/dev/zero of=data bs=1M count=2k
dd: error writing ‘data’: No space left on device
1808+0 records in
1807+0 records out
1894776832 bytes (1.9 GB) copied, 5.78675 s, 327 MB/s
```

The result is that, although the pool cannot store more data, it doesn't look alarmingly full (a red colour is triggered above 90%):
Imgur

It happens because the pool reports the size of its file contents, while some space is also reserved by btrfs:

```
[root@rockstorvm share1]# btrfs fi show .
Label: 'pool1'  uuid: 6609efe9-6275-427c-baa9-bb0fa8528d63
        Total devices 1 FS bytes used 1.77GiB
        devid    1 size 2.00GiB used 2.00GiB path /dev/sdb
```

This example is a bit contrived with my tiny virtual disks, but nonetheless I would like the dashboard to show accurately when a pool is nearly full and therefore in need of management, so I'll go have a look at the pool code.

Perhaps in the future it would also be nice to display the btrfs data/metadata in more detail, on the individual pool views.

sfranzen:
Will fix #1454. Should now work for any pool configuration; test results to follow.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -875,7 +875,7 @@ def usage_bound(disk_sizes, num_devices, raid_level):
     disk_sizes = [size - bound for index, size in enumerate(disk_sizes)
                   if index <= bounding_q]
 
-    new_bound = usage_bound(disk_sizes, bounding_q + 1)
+    new_bound = usage_bound(disk_sizes, bounding_q + 1, raid_level)
 
     return bound * ((chunks / data_ratio) - parity) + new_bound
 
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -812,6 +812,74 @@ def pool_usage(mnt_pt):
     return used / 1024
 
 
+def usage_bound(disk_sizes, num_devices, raid_level):
+    """Return the total amount of storage possible within this pool's set
+    of disks, in bytes.
+
+    Algorithm adapted from Hugo Mills' implementation at:
+    http://carfax.org.uk/btrfs-usage/js/btrfs-usage.js
+    """
+    # Determine RAID parameters
+    data_ratio = 1
+    stripes = 1
+    parity = 0
+
+    # Number of chunks to write at a time: as many as possible within the
+    # number of stripes
+    chunks = num_devices
+
+    if raid_level == 'single':
+        chunks = 1
+    elif raid_level == 'raid0':
+        stripes = 2
+    elif raid_level == 'raid1':
+        data_ratio = 2
+        chunks = 2
+    elif raid_level == 'raid10':
+        data_ratio = 2
+        stripes = max(2, int(num_devices / 2))
+    elif raid_level == 'raid5':
+        parity = 1
+    elif raid_level == 'raid6':
+        parity = 2
+
+    # Round down so that we have an exact number of duplicate copies
+    chunks -= chunks % data_ratio
+
+    # Check for feasibility at the lower end
+    if num_devices < data_ratio * (stripes + parity):
+        return 0
+
+    # Compute the trivial bound
+    bound = int(sum(disk_sizes) / chunks)
+
+    # For each partition point q, compute B_q (the test predicate) and
+    # modify the trivial bound if it passes.
+    bounding_q = -1
+    for q in range(chunks - 1):
+        slice = sum(disk_sizes[q + 1:])
+        b = int(slice / (chunks - q - 1))
+        if disk_sizes[q] >= b and b < bound:
+            bound = b
+            bounding_q = q
+
+    # The bound is the number of allocations we can make in total. If we
+    # have no bounding_q, then we have hit the trivial bound, and exhausted
+    # all space, so we can return immediately.
+    if bounding_q == -1:
+        return bound * ((chunks / data_ratio) - parity)
+
+    # If we have a bounding_q, then all the devices past q are full, and
+    # we can remove them. The devices up to q have been used in every one
+    # of the allocations, so we can just reduce them by bound.
+    disk_sizes = [size - bound for index, size in enumerate(disk_sizes)
+                  if index <= bounding_q]
+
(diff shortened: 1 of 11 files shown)


sfranzen:
## Test procedure
- Create pool with four disks: 2x 2 GiB, 2x 1 GiB in given RAID configuration;
- Fill pool with data: `dd if=/dev/zero of=/mnt2/pool1/data bs=1k`;
- Verify that pool size is reported as [here](http://carfax.org.uk/btrfs-usage/) and usage is at (or very close to) 100%.

## Test results
- [x] Single: 6/6 GB
- [x] RAID0: 6/6 GB
- [x] RAID1: 3/3 GB
- [ ] RAID10: 2/3 GB
- [x] RAID5: 4/4 GB
- [x] RAID6: 2/2 GB

Looks like one case to fix :sunglasses:.

sfranzen:
Found another issue when using an odd number of drives in a pool, because I had accidentally removed an important line at some point. Made a slightly more exotic test case for 3x 2 GiB and 2x 1 GiB disks.

## Test results
- [x] Single: 8/8 GiB
- [x] RAID0: 8/8 GiB
- [x] RAID1: 4/4 GiB
- [x] RAID10: 4/4 GiB
- [x] RAID5: 6/6 GiB
- [x] RAID6: 4/4 GiB

MFlyer:
> should I just remove the logging output?

Hi @sfranzen , just remove once you've completed your testing

sfranzen:
Ok, found another slightly annoying issue: the usable size shown in the Add Pool view is still calculated wrongly because that is coded into the Backbone view. I don't like duplicate code, so I will:
- [ ] Move the `usage_bound` function from the Pool model into the `fs.btrfs` module;
- [ ] Create simple view function to accept a GET request with an array of disk sizes and a raid level and return the usage bound;
- [ ] Replace current Add Pool size calculations with this result;
- [ ] Rebase to fix my merge conflicts :wink:

sfranzen:
Ok, this is now ready for review/merging and depends on rockstor/rockstor-jslibs#7.

In the end I changed quite a few more things than I originally planned to, but even though they are not very noticeable, I'm quite proud of how the pool creation view looks and functions now. Please try it out and let me know what you think.

src/rockstor/fs/btrfs.py
```diff
@@ -127,7 +127,7 @@ def pool_raid(mnt_pt):
         if (len(fields) > 1):
             block = fields[0][:-1].lower()
             raid = fields[1][:-1].lower()
-            if not block in raid_d and raid is not 'DUP':
+            if block not in raid_d and raid is not 'DUP':
```
schakrava: nice catch :)

src/rockstor/fs/btrfs.py
```diff
@@ -773,49 +792,92 @@ def shares_usage(pool, share_map, snap_map):
...
+
+    # If we have a bounding_q, then all the devices past q are full, and
+    # we can remove them. The devices up to q have been used in every one
+    # of the allocations, so we can just reduce them by bound.
+    disk_sizes = [size - bound for index, size in enumerate(disk_sizes)
+                  if index <= bounding_q]
+
+    new_bound = usage_bound(disk_sizes, bounding_q + 1, raid_level)
+
+    return bound * ((chunks / data_ratio) - parity) + new_bound
```
schakrava: Nice work, thank you!

src/rockstor/fs/btrfs.py
sfranzen: Wasn't me :wink: my editor does automatic PEP8 checking, it complains about a lot of small things.

schakrava:
@sfranzen to your question about logging, we CAN turn off debug messages by setting log level in settings.conf.in prior to building Rockstor in any environment. I think we should at some point not have debug level for production(rpm build) by default but can provide a simple script to turn it on to aid in troubleshooting. see #1478. and thanks for bringing this up.

schakrava:
Thank you for improving code quality and detailed work log @sfranzen . Very nicely done. I am ok merging this pr, but I do like to ask a few questions and perhaps suggest some improvements. It's totally up to you to implement them in this pr or the next(my preference).

1. A minor issue. In the Create Pool form, the first click after selecting the raid level always triggers validation. So if I change the raid level from single to anything else and then proceed to select a checkbox to pick my first disk, it doesn't select the checkbox, but triggers validation and displays an error under the raid level dropdown. It doesn't matter where you click, but the first click after changing the raid level triggers the validation. This is a minor regression and I'll leave it up to you to fix in this pr or the next.

2. In Create pool again, as you select disks, each selection makes an api call to the backend. This is a bit expensive. I get your point about not duplicating usage calculation logic in the frontend, but it may actually be justified in this case. I don't feel strongly one way or the other about this, so I'll leave it up to you to improve it as you wish.

3. usage bound api as a function based api view in pools.py looks out of place. I like the simplicity of it though, something to think about as you tinker more with django to restful design.

schakrava:
@sfranzen I'll wait for your reply on (1). If you like me to wait until you fix it, let me know. If not I'll merge this as is and you can submit another pr.

sfranzen:
@schakrava Thanks for the feedback.

1. Good catch, I'll just fix that within this PR.
1. I thought about that, but since my main experience is application programming I don't really know what kind of expense a remote call carries. I think you're right that this may be pushing the "DRY" principle too far. :wink: It's no problem to do it client side, in fact the algorithm I translated is already in javascript. It will take a little more care than just reverting some commits though, so I'll do it in a future PR if you don't mind.
1. Yes it is, it probably should have gone in `api/commands/usage_bound`.

schakrava:
Thanks @sfranzen. Nice work!

PR #1472, opened 2016-10-12
support long nutanix device names. Fixes #1471

### LINKED ISSUE #1471: support long nutanix device names
Thanks to forum member @dannyboy1121 it has come to light that serial numbers of Nutanix CE block devices (a KVM based platform) can exceed that of Rockstor's Disk.name db field, assuming the consequent by-id name uses the entire serial number. Is is not yet know however if udev is even providing the require by-id names. A request in the following forum thread has been placed to gain further information on the /dev/disk/by-id mappings that are required for all Rockstor versions from 3.8-14.02 onwards.

Please update the following forum thread for significant progress on this issue:
https://forum.rockstor.com/t/adding-virtual-disks-pool-issues/2006

phillxnet:
Forum member @dannyboy1121 reported a by-id device name of 68 characters, exceeding the current testing channel updates limit of 64 characters. This was on a Nutanix CE hypervisor platform. As from pr #1320 and Rockstor version 3.8-14.02 testing channel updates, disk names use the much longer by-id format. This pr avoids a regression that will affect platforms generating > 64 character by-id names. This should also improves future compatibility with other similarly long named by-id device mappings; although no other reports of >64 char by-id names has been received.

This pr doubles the name field within the Disk model to 128 characters and contains the associated South migration file which was tested on a current master branch created db and the name field was confirmed as changing from 64 char limit to 128 char limit via the following command:
```
\d storageadmin_disk;
```

Fixes #1471 

Ready for review.

### DIFF
--- a/src/rockstor/storageadmin/migrations/0047_auto__chg_field_disk_name.py
+++ b/src/rockstor/storageadmin/migrations/0047_auto__chg_field_disk_name.py
@@ -1,0 +1,588 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+
+        # Changing field 'Disk.name'
+        db.alter_column(u'storageadmin_disk', 'name', self.gf('django.db.models.fields.CharField')(unique=True, max_length=128))
+
+    def backwards(self, orm):
+
+        # Changing field 'Disk.name'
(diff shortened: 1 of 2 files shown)


phillxnet:
Please see the associated #1471 issue for the procedure used to generate the migration file and the pre and post "\d storageadmin_disk;" results confirming the desired character limit change on the disk.name field.

@schakrava  I propose that this pr be incorporated in the current testing channel updates to avoid stable to stable version release regressions associated with pr #1320 changes. I do not however anticipate many users being affected by this regression so the risk of db migration late in a testing channel update cycle is to be weight in this light.

schakrava:
Thanks for taking care of this quickly @phillxnet . LGTM

phillxnet:
@schakrava Cheers.

PR #1487, opened 2016-10-18
Minor dashboard update, increase animation timing to grant a better smoothing

MFlyer:
Hi @schakrava,
this small PR is ready and can be merged: had it without opening an issue because is a really really small one

M.


schakrava:
Thanks @MFlyer !

PR #1490, opened 2016-10-18
nut timed shutdown option. Fixes #982 and #1458

### LINKED ISSUE #982: NUT timed shutdown option
Add the facility to GUI config a NUT initiated shutdown after a user chosen length of time. This would override the current default of only shutting down when On Battery (OB) and Low Battery (LB).
Please update the following forum post with progress on this issue:- http://forum.rockstor.com/t/ups-nut-rock-ons/592
Thanks to @g2_ufo on our forum for requesting this feature.

### LINKED ISSUE #1458: improve NUT port config usability
Thanks to @alexey in the following forum thread for highlighting this issue. When configuring a usb UPS the auto suggestion can be misleading; if we change this to the suggestion type used in the user and password boxes then it would be more obvious that this is a tip not an auto fill and would be more consistent with the user and password fields in this regard ie:

change the "auto" suggestion to "'auto' or serial ~~port~~ device name" akin to the tool tip advice.

Please update the following forum thread when this issue is resolved:
https://forum.rockstor.com/t/failed-to-start-nut-ups-service-due-to-a-system-error/2135

phillxnet:
Adds an additional selector on the NUT configuration page to chose one of several "Shutdown Timing" options ranging from 30 seconds to 32 minutes. There is also a default option of "When Battery Low" which equates to the prior behaviour, i.e shutdown only when On Battery (OB) and Low Battery (LB). This default is also enforced in the back-end config file parser.

As the prior instantiation of NUT already enabled NUT's build in scheduler along with an early-shutdown directive (disabled) this pr addresses the front end capability to select from a range of common values and the back end config file parser to edit existing relevant lines within the prior existing template file for all schedules: /opt/rockstor/conf/upssched.conf. Previously this file was wholesale copied to the working location of /etc/nut/upssched.conf but this pr introduces an on-the-fly edit that accounts for user additions of early shutdown lines such as were already provided but in remarked out form.  Given some instructions were available on the community forum on how to enable the early shutdown by un-remarking the last to lines:

```
#AT ONBATT * START-TIMER early-shutdown 120
#AT ONLINE * CANCEL-TIMER early-shutdown
```

the parser has been made aware of this possibility and also checks for and removes duplicate entries concerning the same timer events. This should ensure that user edits of this sort (i.e un-remarking these 2 lines) will be unaffected until a new configuration is submitted, which in turn allows the user to review the new setting within the config page as this will take prescience.

Fixes #982 

While working on this issue and it's associated default it was though prudent to establish a default entry for the port setting. Previously this had no default and a misleading greyed out place holder / example entry of auto which was reported by several users on the forum to be confusing. The new behaviour is to enforce a default of "auto" when no entry is made as this works for most NUT USB drivers and is likely to be the more commonly employed device type going forward. This enhancement was detailed in issue #1458 and the addition of the new default along with an improved text field placeholder text:

Fixes #1458.

This pr has been tested using 4 different NUT drivers across 3 different UPS models and all behaviours were as expected given the various drivers restrictions, ie not all NUT drivers are able to power down the UPS device successfully. This short coming  is now more evident as previously the UPS would be run until the battery was low and hence would be nearing the built in power off mechanism associated with that state anyway. Where as now if set to power off after 1 minute on a UPS that has significantly more battery reserve there can be an appreciable time period prior to the UPS powering down. On drivers where this function was supported / working the UPS device was powered off as expected according to the reported ups.delay.shutdown register which is often around 20 seconds.

The function of Net Server and Net Client were also tested to ensure they obeyed the remit of the early shutdown setting: this was done by having one Rockstor machine in Net Server mode (with the UPS device attached) instruct a second Rockstor machine to shutdown first. The second machine was on the same network and configured in Net Client mode.  Note that the Net Client machine's early shutdown timer would also take precedence but only if it's setting was shorter than that of the Net Server's shutdown timer setting. This is expected behaviour.

Ready for review.

A couple of unrelated (smard) user facing typos were also included.

A reformat from the previous 2 spaces to the project default of 4 spaces was not enacted where relevant in order to aid in code review. This can be done at a later date if required.

### DIFF
--- a/src/rockstor/system/nut.py
+++ b/src/rockstor/system/nut.py
@@ -241,7 +241,12 @@ def establish_config_defaults(config):
         config['upsname'] = 'ups'
     if ('nutserver' in config) and (config['nutserver'] == ''):
         config['nutserver'] = 'localhost'
-
+    # Establish a port default of auto as this works with most USB connected
+    # UPS devices which is a more common connection type going forward.
+    # Note that this setting persists but netclient mode, which doesn't use
+    # the port option, is unaffected by this setting so persistence is harmless.
+    if ('port' in config) and (config['port'] == ''):
+        config['port'] = 'auto'
 
 def configure_nut(config):
     """
--- a/src/rockstor/system/nut.py
+++ b/src/rockstor/system/nut.py
@@ -122,13 +122,22 @@ def config_upssched():
     """
     # the upssched config file
     upsshed_conf_template = ('%s/upssched.conf' % settings.CONFROOT)
-    # would be better if we could set a file creation mask first then copy
+    # As our custom parser for nut.conf, ups.conf, upsd.conf, upsd.users
+    # and upsmon.conf can only work on whole word keys we cannot use it
+    # for editing upssched.conf which can have multiple simultaneous entries
+    # of multi word keys, ie:
+    # "AT ONBATT * START-TIMER on-batt 10"
+    # "AT ONBATT * START-TIMER early-shutdown 120"
+    # TODO: add upssched.conf parser / config editor to replace straight copy
+    # TODO: from template.
+    # Would be better if we could set a file creation mask first then copy
     # TODO: set file creation mask to 640
     shutil.copyfile(upsshed_conf_template, UPSSCHED_CONF)
     run_command([CHOWN, 'root.nut', UPSSCHED_CONF])
     run_command([CHMOD, '640', UPSSCHED_CONF])
     # the upssched command file
     upsshed_cmd_template = ('%s/upssched-cmd' % settings.CONFROOT)
+
     shutil.copyfile(upsshed_cmd_template, UPSSCHED_CMD)
     run_command([CHOWN, 'root.root', UPSSCHED_CMD])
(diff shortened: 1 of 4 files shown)


src/rockstor/system/nut.py
```diff
@@ -146,7 +235,7 @@ def establish_config_defaults(config):
     """
     # An empty config dictionary will be false so raise exception as what else.
     if not config:
-        # todo I am unsure if this is the best way to raise an exception here.
+        # TODO: I am unsure if this is the best way to raise an exception here.
```
schakrava: This is fine, but a more pythonic approach would be to just assume that config is properly passed here and let the interpreter raise the KeyError if that is not the case. The caller of this method or higher up in the view, can raise this exception if the contract of properly configuring the service is broken.

schakrava:
Thank you @phillxnet, for the pr and excellent documentation as always!

src/rockstor/system/nut.py
phillxnet: OK thanks.

phillxnet:
@schakrava Cheers.

PR #1492, opened 2016-10-19
Fixes #1479 by forcing always run_command to be on en_US locale

### LINKED ISSUE #1479: Required locale to US conversion / code to avoid time comparison failing
Reference to [this](https://forum.rockstor.com/t/unable-to-check-update-due-to-a-system-error-time-data-mo-20-jun-2016-does-not-match-format-a-d-b-y/2178) forum thread:

user Uwe_Ganswig (German locale) reported Rockstor failing over update process
rpm_build_info fails [here](https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/system/pkg_mgmt.py#L99) expecting a US/UK abbreviated weekday.

Requires patching

MFlyer:
Hi @schakrava , this should allow Rockstor systems to be on a different locale too (force run_command to always use en_US locale, regardless of system real locale)

Asking to Uwe_Ganswig forum user to test it over de_DE, done some tests over it_IT and seems ok

Mirko

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -84,9 +84,14 @@ def run_command(cmd, shell=False, stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE, stdin=subprocess.PIPE, throw=True,
                 log=False, input=None):
     try:
+        #We force run_command to always use en_US
+        #to avoid issues on date and number formats
+        #on not Anglo-Saxon systems (ex. it, es, fr, de, etc)
+        fake_env = dict(os.environ)
+        fake_env['LANG'] = 'en_US.UTF-8'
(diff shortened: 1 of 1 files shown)


MFlyer:
To @schakrava : we can conside this ready to merge.
Worst case: force using en_US while our Django is already en_US, so won't produce any issue

Actually user Uwe_Ganswig (forum thread) has solved updates issue by patching/testing this PR :)

MFlyer:
Update: after recent tests with forum user Use_Ganswig + local test on my dev env I can confirm user smb/nmb issues not related to this PR, so we can merge :wink:

schakrava:
Thanks @MFlyer! LGTM.

Issue #1459, opened 2016-10-04
Feature request - Emergency patch system on stable channel


MFlyer:
Hi @schakrava,
while writing you [this](https://github.com/rockstor/rockstor-core/pull/1428#issuecomment-251308270) about Rockstor Widgets and thinking about current chart.js state and last merge about smooth animation (not available on last/our chart.js release), I realized we can suffer same issue over Rockstor. Let me explain:

Assumption: before having a stable channel release we test PRs over testing channel (and we do a lot! :metal:), so everything should be ok, but you know this is not always the real truth.
Our testing channel has 2/3 days average update time, while stable channel has a ~1 month life cycle, so having an unexpected important issue (not minor oddities) could force us to have a short time stable branch update.

My suggestion:
have a new kind of emergency system just using `patch` as well described by @phillxnet [howto](https://forum.rockstor.com/t/how-to-apply-a-patch-to-aid-in-development/1772) on our forum, but all via Rockstor Web UI
IMO you should also mail stable branch user link on patch available (I'm thinking about a copy&paste system to apply patches)

sfranzen:
Wouldn't this be possible within git/GitHub by creating a separate branch, e.g. "stable", from which to tag the stable releases? That way, if some serious regression gets through, you can pull just the bug fix from the master branch and make a new release.

phillxnet:
Closing as no community input for 7 years and we do now have something akin to @sfranzen suggestion re  a "stable" & "testing" GitHub branch arrangement. But have yet to settle into this give the massive changes of late in the "testing" branch. But intend to cherry-pick from fixes in testing to master for hot-fixes pre-tested in testing. This was first done re 4.6.1-0. We can do the like in the future once we have reached the next stable and merged testing into master again.

PR #1496, opened 2016-10-21
Fixes a small samba service regression 

### LINKED ISSUE #1495: Small Samba regression
Hi @schakrava , reference to this forum thread:
https://forum.rockstor.com/t/error-running-a-command-cmd-at-samba-share-screen/2224/25?u=flyer

MFlyer:
To @schakrava , sorry again, on dev env had no null field over ad config :/

Fixes #1495

Ready to merge before stable release
M.

### DIFF
--- a/src/rockstor/smart_manager/views/samba_service.py
+++ b/src/rockstor/smart_manager/views/samba_service.py
@@ -69,7 +69,7 @@ class SambaServiceView(BaseServiceDetailView):
                 #if AD configured and ON set workgroup to AD retrieved workgroup
                 #else AD not running and leave workgroup to one choosen by user
                 adso = Service.objects.get(name='active-directory')
-                adconfig = {}
+                adconfig = None
                 adso_status = 1
                 if (adso.config is not None):
                     adconfig = self._get_config(adso)


MFlyer:
To @schakrava, updates after solving with forum user: both this and #1492 can be merged, user issue not on Rockstor (uefi bios + r8169 driver card having nic up/down, moving to legacy bios solved it :smiley: )

M.

tomtom13:
Fixes #1515

tomtom13:
@schakrava can we expedite this pull ? I've tested a 3.8.15 without this fresh out of the box (and also after updating it it .1) and result is that one can NOT setup a successful samba session. Samba is running but one can not get to a share from windows machine either using guest or actual user account because master samba settings are not set correctly !!! (the workgroup bug block the rest defaults to be stored in samba config)

MFlyer:
Hi @tomtom13 if you need it's just a 1 line code on `src/rockstor/smart_manager/views/samba_service.py`, having a reboot after changes should make it work on testing/stable env too! :)

tomtom13:
@MFlyer it's not about me, I know how to patch my box :P it's about normal users, right now anybody installing rockstor will fail to install samba and for 90% home users it's a deal breaker if they can't have samba.

MFlyer:
Hi @tomtom13 I know you can do that :P My #1459 probably was a damned premonition to this :disappointed: 

@schakrava @phillxnet & @sfranzen I believe we seriously have to handle "missing important patches":
my idea with #1459 is to have something requiring user interaction (have a form where users insert PR number then autopatch), but that forces user to manual intervention.
Suggestions? Can we have urgent PR merged and released on stable channel too?? not sure about that.
@schakrava can we add a new ad hoc yum repo just for emergency patching (so doing we won't have a new release like 3.8.15-1 or similar)

schakrava:
@MFlyer and @tomtom13 . Thanks for this important discussion. I wish we merged this before 3.8.15, but it's too late now. There are a few takeaways from this experience. 1) Let's test our stuff thoroughly, and especially for backend, please add tests going forward. @phillxnet has added a bunch of them recently you can use as a reference to get up to speed on that. 2) I need to do a better job as part of our pre-stable release testing and you are right, a personal note about this regression would have helped, but I could've looked as well; shit happens. 3) Since the fix is so small, we can have it as "errata" for now. I am not in favour of pushing hot-fixes into stable. We can also put a console script that can pull an arbitrary gist and make hot-fixes easily applicable.

schakrava:
oh, finally @MFlyer thanks for the fix and sorry couldn't merge it fast enough :(

Issue #1358, opened 2016-06-28
smart parsing error on specific log summary


phillxnet:
Thanks to @kziegler in the following forum thread for reporting this issue:
https://forum.rockstor.com/t/null-value-in-column-details-violates-not-null-constraint-detail-failing-row-contains-4-6-973-8240-active-or-idle-command-aborted-null/1674
On initial inspection a certain log format from a WDC WD10EADS-00M drive is triggering a parsing bug that is preventing the display of SMART data for this drive.

A dump of the various smartctl commands has been requested to assist in addressing this issue.
Please update the above forum thread on significant development with this issue.

phillxnet:
I strongly suspect that this has now been fixed by recently merged #1499 and is in fact duplicated by the more recent and now closed pr #1498.

As the above referenced pr is now merged I am closing this issue as:
duplicate of #1498 
Fixed by #1499 

I will update the referenced forum thread upon the associated pr is release in testing channel.

phillxnet:
The referenced forum thread has now been updated.

PR #1499, opened 2016-10-25
null value in column "details" when parsing smart error log. Fixes #1498

### LINKED ISSUE #1498: null value in column "details" when parsing SMART error log
Thanks to forum member alexey in the following thread for reporting this issue. The SMART error log parser is, under specific hardware conditions, producing unexpected and incompatible output resulting in the following Web-UI error message whenever the "Refresh" button in the smart screen for the specific drive is pressed:

```
Houston, we've had a problem.

null value in column "details" violates not-null constraint DETAIL: Failing row contains ...
```

with the associated Traceback having the penultimate line:

```
IntegrityError: null value in column "details" violates not-null constraint
```

resulting in no SMART info available via the Web-UI for this drive.

As a result of alexey submitting the requested command outputs from the affected drive via the forum thread the reported issue has been reproduced and narrowed down to the error log parsing component of the pre-processing that is required to display the same in the Web-UI.

Please update the following forum thread with significant progress on this issue:
https://forum.rockstor.com/t/smart-not-working/2137

phillxnet:
Using the smartctl outputs supplied by forum member alexey it was found that when no details are specified in the "smartctl -l error" output the resulting 'details' list element was given a "None" value, this caused an exception in the Web-UI code. The solution presented in this pull request involves replacing the None attribution to the 'details' list element with a string = 'No Sector Details Available'.

This change circumvents the "null value in column "details" violates not-null constraint ..." error and also informs the user via the details column that no such information is available.

Fixes #1498 

Thanks to alexey in the following forum thread for contributing the required smartctl outputs:
https://forum.rockstor.com/t/smart-not-working/2137

The resulting code was tested both on the supplied smartctl dumps and on a qemu SATA drive emulation and on a real drive that had existing "smartctl -l error" log entries and a real drive that had not error log entries. All parsing was as expected with no observed regression.

Ready for review.

### DIFF
--- a/src/rockstor/system/smart.py
+++ b/src/rockstor/system/smart.py
@@ -177,23 +177,23 @@ def error_logs(device, custom_options='', test_mode=TESTMODE):
     # log contains errors but otherwise executes successfully so we catch this.
     overide_rc = 64
     e_msg = 'Drive %s has logged S.M.A.R.T errors. Please view ' \
-                 'the Error logs tab for this device.' % local_base_dev
+            'the Error logs tab for this device.' % local_base_dev
     screen_return_codes(e_msg, overide_rc, o, e, rc, smart_command)
     ecode_map = {
-        'ABRT' : 'Command ABoRTed',
-        'AMNF' : 'Address Mark Not Found',
-        'CCTO' :  'Command Completion Timed Out',
-        'EOM' : 'End Of Media',
-        'ICRC' : 'Interface Cyclic Redundancy Code (CRC) error',
-        'IDNF' : 'IDentity Not Found',
-        'ILI' : '(packet command-set specific)',
-        'MC' : 'Media Changed',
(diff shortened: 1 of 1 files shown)


phillxnet:
Please see issue #1498 for details of the "smartctl -l error" output that triggered the rouge behaviour.

phillxnet:
N.B. after the submission of this pr it was noticed that the following redundant and ineffective pre-existing line was not removed.
```
    print ('summary_d %s' % summary)
```
from src/rockstor/system/smart.py

This can be added to this pr if required but is essentially harmless and so could easily be addressed in later 'code cleaning' pull requests.

schakrava:
Thank you @phillxnet ! LGTM

phillxnet:
@schakrava Cheers.

phillxnet:
I have just noticed a duplicate issue that this pr also addresses:
Fixes #1358

PR #1500, opened 2016-10-26
Issue#1278 services config as modal

MFlyer:
Hi @schakrava & @priyaganti ,
this PR is ready to review/merge.

Had it because found it on Looney Bean and quite easy to solve :)

This is how I got modals without full rewriting of views:
- changed router configureservice to open a modal instead of moving maincontent to view
- added some hacks to grant switch on x/cancel buttons / submit successful from modal back to services (app_route.navigate with no trigger)
- resized configure_SERVICENAME templates to better fit on modals

Mirko

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/router.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/router.js
@@ -292,7 +292,11 @@ var AppRouter = Backbone.Router.extend({
 
     configureService: function(serviceName, adStatus) {
         this.renderSidebar('system', 'services');
-        this.cleanup();
+        //Left here like a comment for documentation purpose
+        //Having configurations in modal windows we're leaving service page navigating
+        //to service config page, but we don't cleanup up to avoid closing socketio
+        //Add: this let us have also a nice view of services real state updating while saving/changing configs
+        //this.cleanup();
         var service_options = _.isUndefined(adStatus) ? {serviceName: serviceName} : {serviceName: serviceName, adStatus: adStatus}
         this.currentLayout = new ConfigureServiceView(service_options);
         $('#services_modal .modal-body').empty();
@@ -873,6 +877,12 @@ $(document).ready(function() {
         $('#globalerrmsg').empty();
     });
 
+    //Function to handle services configuration modal closing on
+    //Grant both on submit or cancel we move back to services page,
+    //without trigger so we don't have a page refresh
+    $('#services_modal').on('hidden.bs.modal', function (e) {
+        app_router.navigate('services');
+    });
 
     // Initialize global error popup
     $('#global-err-overlay').overlay({load: false});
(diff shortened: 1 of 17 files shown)


sfranzen:
@MFlyer I also wanted to use plenty of modal dialogs for my snapper view, so I coded [this](https://github.com/sfranzen/rockstor-core/blob/snapper/src/rockstor/storageadmin/static/storageadmin/js/views/common/bootstrap_modal.js) "backbonified" dialog that works like this:
```
HTML:
<div id="myModal"></div>

Inside a Backbone view:
var dialog = new ModalView({
    el: this.$('#myModal'),
    template: window.JST.my_modal_content
}).render();
dialog.once('submit', this.mySubmitHandler, this);
```
It is

MFlyer:
Hi @sfranzen, had same idea about a new modal view (if I'm not wrong there's an existing backbone plugin too), so let's check it also with @priyaganti.
My suggestion about adding a new view : check current Rockstor view to avoid missing parts :)
M.

MFlyer:
Add for @sfranzen : 
for a better coding/code structure the new view should be on `src/rockstor/storageadmin/static/storageadmin/js/rockstor.js` , becoming globally available.

If job won't kill all my time, I'll check it today :)
M.

MFlyer:
Hi all,
added a new #1501 issue for a common Rockstor Modal View on 3.8-16 roadmap, so meanwhile we can marge this and have some good testing on next cycle

M.

sfranzen:
@MFlyer Should it? I mean I don't really mind, but since all of our JS is in the global scope, I personally find it a little clearer to have them in the `common` section, which is added to the JS output before the rest of the views.

MFlyer:
Hi @sfranzen :smiley: 

Just to let you better understand my idea: check rockstor.js code, my idea is to extend a Rockstor view instead of the generic backbone view, so we inherit some defaults too (fetch, etc etc).
We can do this in common subfolder or directly in Rockstor (my idea: better in rockstor.js to have it on top of all)
M.

schakrava:
@MFlyer modals are looking nice! Upon testing, I noticed a minor glitch. For example, follow these steps. 
1. turn off rock-on service if it's previously on.
2. configure the rock-on service. (doesn't matter even if it's preconfigured, just submit on modal)
3. turn the service on again. You'll notice that the switch slightly shifts to the left as expected, but after a few seconds, when the listener gathers the update of all services, it should line up with rest of them. This update is not happening. There is a helpful error in the console. This screenshot might help.

rock-on_service

Could you please fix it? Also, you should rebase to merge @phillxnet 's nut_configure template changes.

MFlyer:
@schakrava :+1:  will check it asap

MFlyer:
PR rebased over last Phil PR, tomorrow going to check glitch!

MFlyer:
First update: find out that with new modal config pages websockets get disconnected, checking this too (maybe partially related to that glitch)

MFlyer:
Great @schakrava , your attention to that glitch made me find this #1522

Solving over python-socketio migration :)

schakrava:
@MFlyer are you still working on this or is it ready for me to test and merge?

MFlyer:
@schakrava ready to test & merge, sorry had my ok go on PR :)

schakrava:
Thank you @MFlyer . Looks great now!

MFlyer:
Thank you @schakrava ,
I was feeling uncomfortable having Rockstor messed up because of previous wrong PR with some pieces from modal :)

PR #1509, opened 2016-11-05
SMART parsing issue report helper script. Fixes #1507

### LINKED ISSUE #1507: SMART parsing issue shell script helper
To aid in debugging user reported SMART parsing errors it has proven useful to retrieve the raw output from the SMART commands used by Rockstor to present the Web-UI parsed information. In the past this has required repeated forum posts indicating the required command set. I propose that a simple shell script be included in /opt/rockstor/src/rockstor/scripts that accepts a device name as a single argument ie:

```
/opt/rockstor/src/rockstor/scripts/dump-smart.sh /dev/disk/by-id/ata-ST3160021AS_5JS1HNX9
```

Where the given device name, in any smartctl viable device naming convention, indicates the problem drive. This would then simplify future reports of such SMART parsing issues by preparing the smartctl output dumps required to address the parsing issues encountered via the built in developer accessible testing mode in:

    src/rockstor/system/smart.py

It is proposed that this initially only cater for HBA smart info retrieval, as opposed to the more complex syntax required when drives are behind some raid controllers.

phillxnet:
A bash shell script intended to aid in reporting SMART parsing
issues by generating the required smart-issue-report.tar.gz
consisting of smartctl command output dumps in a format
compatible with the build in developer test function found in
src/rockstor/system/smart.py

Fixes #1507 

Ready for review.

Reservations: I am unsure if the chosen location for this shell script is appropriate?

### DIFF
--- a/src/rockstor/scripts/dump-smart.sh
+++ b/src/rockstor/scripts/dump-smart.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+#
+# Simple shell script to dump the raw smartctl command output as used internally
+# by Rockstor prior to parsing for Web-UI accessible SMART info.
+#
+# Intended as an aid to reporting SMART parsing issues on:
+# https://forum.rockstor.com/
+#
+
+# Check number of script arguments.
+if [ "$#" -ne 1 ]; then
+    echo
+    echo "Please supply a single device name using the /dev/sdX or /dev/disk/by-id reference"
+    echo "N.B. Use the base device - not a partition reference."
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet . Looks good, though a bit surprising to see you went old-school with bash instead of scripting in python. If this were a python script, we can add a console script and can be invoked from the bin directory. Please consider porting to python if possible.

phillxnet:
@schakrava Yes, agreed, and Cheers. Shall revisit at a later time and port to python as advised.

Issue #1512, opened 2016-11-07
[Minor issue] Share creation - better UI & code refactoring


MFlyer:
Ref to
https://forum.rockstor.com/t/share-size-must-be-an-integer/1912/
https://forum.rockstor.com/t/enhancement-enlage-field-size-on-shares-page/2297/

Had same issue one/two months ago and had some check on it: our issue not on backend, but on frontend js d3 slider.

M.

tomtom13:
@MFlyer could you please remove limit for size on slide OR provide a tick box saying "enable over provision" that will let users exceed current available size ?

MFlyer:
@tomtom13 why ? O_o

tomtom13:
I think I've explained this in forum,

I've had a situation where I've create a pool with single share and filled the FS up to 50%, then went on to create a second share and could not span it across 100% of pool space. I want two shares dynamically fight for space and not use quotas. 

On forum I think @schakrava does agree that it should be to users choice whenever to / not to use quotas and how to configure those.

Let's say user got a 1 share in 1 pool. share is filled to 90%, they want to migrate ALL data to share that they are about to create ... I know that quotas don't work right now, but when they will start working we will end up with a massive can of worms.

tomtom13:
still, @MFlyer you need to understand that quota means that you want something not to be bigger than X, not that it's not bigger than free space.

I've used quotas in my life and you can freely say that you want for example user difrectories not to be bigger than 1GB for 500 users on 300GB disk. Yes this is cheating, because you over provisioned but usually people will not get near a allowed 1GB

Also in current setup if you setup a single pool with 1 share and allocate a 100% to it, who says you can allocate any more space to any new share ?!?!?!?! letting in user create a share with what ever free space is on the disk is alreade creating a over provisioned setup - because first share has 100% and second has a bit more.

MFlyer:
You're right Tom, I agree on that, but actually "downgrading" that slider doesn't make sense: it will be anyway required over quotas (updrage: have slider on quota enabled, no slider if no quota!)

M.

tomtom13:
so how with this slider and quotas enabled I can setup a 1TB pool with 3 shares each allowed 500GB after filling first share to 600GB and than creating reminding shares ? 

I know what you are trying to achieve, but because ewe don't keep track of allocated size for other shares this slider is NOT functioning as intended.

MFlyer:
[...]

PR #1513, opened 2016-11-08
Issue#1512 share creation issues

MFlyer:
This fixes issue #1512 both on decimal error and input being too small

Ready to merge
Mirko


schakrava:
Thanks @MFlyer

Issue #1515, opened 2016-11-10
Regression in 3.8.15 - not able to set samba workgroup - spits out an error


tomtom13:
Just FYI:

`            Traceback (most recent call last):
  File "/opt/rockstor/src/rockstor/smart_manager/views/samba_service.py", line 83, in post
    update_global_config(global_config, adconfig)
  File "/opt/rockstor/src/rockstor/system/samba.py", line 134, in update_global_config
    tfo.write('    %s = %s\n' % (k, ad_config[k]))
KeyError: 'workgroup'`

I've tried a lot different names and everything returns error - even the example from info balon "mygroup" ( or "MYGROUP" :D )

MFlyer:
@tomtom13 : check my comment on PR :)

M.

phillxnet:
@tomtom13 Thanks for opening this and updating accordingly re dup status etc and your contributions on issue #1495 and your prompt but unfortunately redundant pr #1516.

Given the referenced duplicate issue of #1495 is now fixed by pr #1496 are you happy to close this issue as:
Fixed by #1496 ?
Thanks.

tomtom13:
Damn, I thought I done that :/ 
Fixed by #1496

Issue #1481, opened 2016-10-14
[Feature Enhancement] Improve samba shares configuration


MFlyer:
While working on #1467 discovered currently samba shares don't inherit acl configured over Rockstor Pools/Shares, so an ACL system like on shares creation could be desired (actually users have to manually add samba share custom params like samba create mask, force user/group etc, or add SGID bit to Share etc etc)

tomtom13:
Yeah, recently got bitten by creation mask and had to add it my self :/ (sharething was unable to list stuff because of no search bit for "others").

Anyway, I think that samba bit reaaaaaaalllyyy deserves a bit more love because there are a lot of things for those shares that can be configured. 

I know that #1456 was a bit quiet recently, my fault - just bussy at work pushing development of truck forward (had some safety design work to do right after coming out of sicknes :/ ).

MFlyer:
Hi @tomtom13 & all other guys,
adding some steps (split in 2 branches) to cover this issue:

- Grant xattr in a Rockstor domain member
- Grant acl inherited

MFlyer:
Having some tests and it seems working fine (maybe we need a better share management - missing inherit/recursive option on share creation, coded backend not frontend)

Nice find of the day:
mount over Rockstor doesn't show xattr on data disks, but we have it (googled to understand if btrfs has default attrx without answer, but now I can say it has :D )

MFlyer:
Adding reference to #981

MFlyer:
Adding notes:

`acl allow execute always` on [global] or a single share grants exec bit (.exe, .bat, etc) for windows files

MFlyer:
Trying to code on new features/features enhancement finally got back to this :tada: and had some thinking about samba shares + ACLs:

Having ACLs inheritance on #981 : ok, we can have it
Having ACLs inheritance on Samba shares with WIN clients: yep, we can have it "hard coded" (think about a button with "enable extended ACLs for Windows"), but usually Sysadmins having samba+win clients have their way via vfs objects and shares management from win machine and we can't grant a "fully working Rockstor Samba + win clients" env having access only to Rockstor side

MFlyer:
Memo : http://pig.made-it.com/samba-file-rights.html#25963

Marenz:
This might belong to this issue:

[...]

PR #1516, opened 2016-11-11
Issue #1515

tomtom13:
This will just remove a line that was committed by @MFlyer in commit 1e5e6fd07002ddbdc9fc199869f2deade06fead2

I know that there might be a better sollution, but I don't know this code and personally I would preffer to concentrate on networking config. 

Also without this, it's impossible to setup samba on new machine (because it will not accept anny settings and will not punch in any default in either)

### DIFF
--- a/src/rockstor/system/samba.py
+++ b/src/rockstor/system/samba.py
@@ -131,7 +131,6 @@ def update_global_config(smb_config=None, ad_config=None):
             tfo.write('\n%s\n' % RS_CUSTOM_HEADER)
             for k in smb_config:
                 if (ad_config is not None and k == 'workgroup'):
-                    tfo.write('    %s = %s\n' % (k, ad_config[k]))
                     continue
                 tfo.write('    %s = %s\n' % (k, smb_config[k]))
             tfo.write('%s\n\n' % RS_CUSTOM_FOOTER)


MFlyer:
Hi @tomtom13 ,
actually there's #1496 to be merged / not yet merged that solves this

Mirko

tomtom13:
Ok, when I'll get home I'll give it a whirl.

Anyway I've notice that your samba commits are mostly geared at users with active directory etc ... when pushing this direction, would it be possible for you to give options to users like "security = share" through combo box ? Not every rockstor user uses it's box in multi-user setup, and they would like as simple as possible setup. 

Don't get it wrong, we all do stuff that server us the most, just let's not alienate group of users by that.

tomtom13:
Since @MFlyer has an update for that I'll close that one.

MFlyer:
Hi @tomtom13 ,
actually had code more on AD side of Samba because that was the faulty one, but `security = ads` runs only on AD integrated Rockstor, while having default `security = user` on a normal samba environment (https://www.samba.org/samba/docs/man/Samba-HOWTO-Collection/ServerType.html#id2559114)
Obviously we can move to a select (security = user || share) or have a different approach via #1481 (moving acls to share creation & have security = user ??)

Issue #1517, opened 2016-11-11
network stack overhaul


tomtom13:
I would like to use this issue to fix a lot of stuff in networking stack:
- manual settings sometimes don't work
- manual setting can magically suck in additional setting from dhcp and populate it magically inside of gui (yeah, scary ... init)
- related to previous, when trying to fix this problem by going to edit page again and removing ALL dhcp aquired setting leaving only manual IP address that one would wish to have ... after hitting submit all manual setting disappear and you only get left with dhcp ones (checked with ifconfig to be true on interface as well). After togling interface OFF then back ON problem seems tobe fixed.
- bonding only works on initial setup, after reboot it's bonkers
- after setting up a bonding in 802.3ad for 4 interfaces and rebooting in VMware, this made interfaces go into crazy loop and get rockstor stuck. Forced reboot and removal of interfaces just to get to logon screen made rockstor forget my log on detail (possible config corruption ?)
- teaming is just plainly defunct
- manual setting forces to put in a random gateway that makes it put route with higher priority than dhcp one resulting in rockstor not being able to access internet. 
- Clicking a pencil (for edit) button on network connection sometimes takes you to 100% unpopulated network page ... but if you stick on the network page for longer than 30 seconds and click the pencil button it then takes you to a properly populated page.

ps. guy could you please pile on any problems you see with networking setup ? I can join it here.

tomtom13:
Just adding more stuff :/
- manual setting forces to put in a random gateway that maker it hit route with higher priority than dhcp one resulting in rocstor not being able to access internet.

tomtom13:
- Clicking a pencil (for edit) button on network connection sometimes takes you to 100% unpopulated network page ... but if you stick on the network page for longer than 30 seconds and click the pencil button it then takes you to a properly populated page.

![fireshot capture 1 - rockstor - https___192 168 44 129_ network_edit_6](https://cloud.githubusercontent.com/assets/22304458/20243257/f0c7a60e-a947-11e6-97e9-8a690ebc6a82.png)
![fireshot capture 2 - rockstor - https___192 168 44 129_ network_edit_6](https://cloud.githubusercontent.com/assets/22304458/20243259/f2d0b206-a947-11e6-9181-ca3bac0e2681.png)

tomtom13:
[...]

PR #1520, opened 2016-11-13
Issue #1517

tomtom13:
First and simple batch of changes:
- remove gateway requirements for manual interface configuration. Forcing users to put random gateway on closed networks, screws up "route" on host and makes installation unable to access outside world.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/network/new_connection.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/network/new_connection.jst
@@ -111,7 +111,7 @@
           </div>
         </div>
         <div class="form-group">
-          <label class="col-sm-4 control-label" for="gateway">Gateway<span class="required"> *</span></label>
+          <label class="col-sm-4 control-label" for="gateway">Gateway</label>
           <div class="col-sm-4">
             <input class="form-control" type="text" id="gateway" name="gateway" {{#if connection}}value="{{connection.ipv4_gw}}"{{/if}} placeholder="Gateway">
           </div>
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/edit_network_connection.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/edit_network_connection.js
@@ -84,14 +84,6 @@ NetworkConnectionView = RockstorLayoutView.extend({
 			}
 		    }
 		},
-		gateway: {
-		    required: {
-			depends: function(element){
-			    return (_this.$('#method').val() == 'manual');
-			}
-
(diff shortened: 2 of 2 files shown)


schakrava:
@tomtom13, thanks for this. Please help me understand this better, I have a few questions. 

- What are closed networks and is gateway not required in such networks? 

- Is it possible/recommended to have a default gw on such networks such as 0.0.0.0 or something like that? 

- And how does this effect normal static ip config where the config is useless without a gw and results in the system being inaccessible over the network?

tomtom13:
Closed network is colloquial for network that is physically connected to internet by not a single device in this network is not aware how to access outside world (or any other network) because they don't know what is a gateway. Do not mistake it for private network - private network is a network with private addressing space that knows / doesn't know how to access outside networks. Security in closed networks is achieved by means that any outsider (without overtaking an equipment inside of your network) is not aware of it, can't physically send any data (because other equipment doesn't know this network exists and how to reach it) and equipment on that network can't reply back because they don't know how to send data outside of this closed network (lack of gateway).

so for example: 
- let's say you have in you enterprise a large backbone and you want to use it to create a dedicated network which nobody should technically access or know of ... and your equipment does not allow you to create a vlan. 
- you have two machines on 1gb network and want them to talk to each other via dedicated 10GB fibre ... you install fibre interfaces in each machine, join them and then give them IP addresses from outside of you your 1GB network. Every time you will attempt to access other machine using a address from this "closed network" you will effectively push traffic through fibre rather than 1GB

0.0.0.0 gateway .... a very bad idea. Newer, under any circumstances provide a phony gateway. Fist, you end up with 2 routes to outside world, second all IP timeouts and TTL actually get fucked up by this because your machine if it can't access something will try some weird trick on other gateway - behaviour is unpredictable, third if your 0.0.0.0 gateway gets higher priority than your real gateway - you machine won't be able to talk to outside world on other interfaces. 

static IP configurations are not useless without gateway - it all depends what you are trying to chive. Technically speaking if we qualify usefulness of link without gateway, it's equally useless nowadays without DNS - how will you access update server if you can't resolve name to ip ? :) 


Finally - the Ethernet interface is NOT technically required a IP address at all ... IPX comes to mind :)

schakrava:
Thanks for the explanation @tomtom13 . I had to make a tiny backend change, but looks good. Thanks for your contribution!

tomtom13:
Seriously no problem. From standpoint of NAS - Network is a first letter of NAS ... so it's rather important and I would rather move it slowly and have as many eyes watching changes / testing / criticizing. To much at stake here, and we all (specially me) make errors.

Issue #1530, opened 2016-11-16
Dashboard widget resizing is glitchy in Chrome on Mac


schakrava:
A user reported that resizing disk activity widget on dashboard causes the boundaries to overlap. This was reported on Mac 10.11.6 with latest Chrome and Firefox. Apparently the behavior is the same on WIndows 10 with Chrome.

I've confirmed it on Mac with Chrome version 53.0.2785.113. I cannot reproduce it on Linux+Chrome or Firefox.

Finally, to reproduce on Mac+Chrome all you need to do is go to the dashboard and expand Disk activity widget and perhaps another one(eg: Network) if you don't see the issue already. Below screen shot shows the issue.

<img width="1084" alt="dashboard_glitch" src="https://cloud.githubusercontent.com/assets/1123552/20348197/27031f0a-abb9-11e6-850e-39866337dcd8.png">

MFlyer:
Hi @schakrava ,
got some initial tests and these don't help us (not able to reproduce):

WIN 7 + Chrome 54.0.2840.99 (latest)-> No overlapping, all fine
WIN 7 + Firefox 50 (latest)-> No overlapping, all fine

Had some time playing removing/adding widgets, having all maximized, some maximized, some not, but it was all ok.
Any idea if user having some add-on on firefox/chrome??

MFlyer:
Had another relaxing 20mins playing with widgets, but without success.
Tomorrow going to check over a WIN10 system
M.

MFlyer:
Hi @schakrava , while checking templates found this -> 003fb9a

Can you test it if solves over a Mac? a wrapper div always required by canvas (no wrapper => canvas overlapping on resize), but having 2 divs with same ID probably made it crazy

Waiting your checks :)

MFlyer:
> Apparently the behavior is the same on WIndows 10 with Chrome

Good UTC morning @schakrava :) , with my new laptop going to test this on WIN10 too so hopefully will have better chance to catch it.

MFlyer:
Hi @schakrava,
after Xmas flu and a battle vs win10 & wifi finally got my new laptop running with dual boot and...reproduced this issue :)

Hope to solve asap, there's an error raising when resizing from small to big

M.

MFlyer:
Got an update about this, @schakrava & @phillxnet can you check too?
[...]

PR #1533, opened 2016-11-16
#1530 - Removing duplicated div id disk-details-ph, both on disk_util…

MFlyer:
Reference to #1530 

Required testing over Mac

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/dashboard/widgets/disk_util_select.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/dashboard/widgets/disk_util_select.jst
@@ -1,4 +1,4 @@
-<div id="disk-details-ph">
+<div>
     <div id="disk-select-ph">
         <span class="h4">Disk activity</span> for disk
         <select name="disk-select" id="disk-select" class="inline">


schakrava:
Thanks for working on this even though you are unable to reproduce it. I've tried on mac and sad to say the issue is still the same.

MFlyer:
Hi @schakrava,
going to have a OS X El Capitan on a Virtualbox: having some code + asking you to test every time is like walking in the dark!
Here we come with our new game: will a 5GB ram WIN7 survive running OS X in Virtualbox? eheh

phillxnet:
@MFlyer I normally have a rather old macbook pro hanging around who's sole use is to test stuff against Rockstor but I'm afraid it's been out on emergency loan for a couple of weeks. I'll let you know when I have it back and then I should be able to chip in on this one. Running pretty vanilla El Capitan (really slowly). Chase me up if I don't mention it for a bit.

Sorry for letting the side down on this one.

Maybe there is some trusty remote access software I can install to give you access, Note though that I'm double NATed here so that might be a problem. Hackintosh road is definitely a pretty bumpy one; in my experience.

MFlyer:
Hi @phillxnet, thanks :D 
Today Hackintosh session over Virtualbox -> :feelsgood:  xD

PR #1544, opened 2016-11-24
Issue#1524 implement client state dict to avoid zomby threads

### LINKED ISSUE #1524: [Feature enhancement] New python-socketio not handling spawned functions
Hi @schakrava ,
this is a memo/reference for upcoming (next days) PR related to https://github.com/rockstor/rockstor-core/issues/1522#issuecomment-260432234 and #1510 

Testing together on new python-socketio for #1510 is good and now I'll check gevent.spawn + that glitch

Important: on every F5 (page refresh) we get for each `gevent.spawn` 2 active threads instead of 1 as expected, that should be quite safe (not a real cpu eater), but better to avoid it and close this before 3.8.16 release

Mirko

MFlyer:
Hi @schakrava ,
this fixes #1524 by adding an ad hoc threads manager (ready for review and merge)
On f6a8fb2 you have a full description of what has been done, ask if you need other infos :)

Had some tests (one of these made me realize PincardManager regression) and it seems ok: dashboard working, logs reader working, sysinfo, etc

M.

### DIFF
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -49,8 +49,30 @@ from system.pkg_mgmt import update_check
 import logging
 logger = logging.getLogger(__name__)
 
+class RockstorIO(socketio.Namespace):
+    "RockstorIO socketio.NameSpace SubClass"
+    
+    def __init__(self, *args, **kwargs):
 
-class PincardManagerNamespace(socketio.Namespace):
+        super(RockstorIO, self).__init__(*args, **kwargs)
+        self.threads = {}
+
+    def cleanup(self, sid):
+
+        if sid in self.threads.keys():
+            gevent.killall(self.threads[sid])
+            del self.threads[sid]
+
+    def spawn(self, func, sid, *args, **kwargs):
+
+        thread = gevent.spawn(func, *args, **kwargs)
+        if sid in self.threads.keys():
+            self.threads[sid].append(thread)
+        else:
+            self.threads[sid] = [thread]
+
+
+class PincardManagerNamespace(RockstorIO):
 
     def on_connect(self, sid, environ):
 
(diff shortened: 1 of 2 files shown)


src/rockstor/smart_manager/data_collector.py
```diff
@@ -49,8 +49,30 @@
...
+    "RockstorIO socketio.NameSpace SubClass"
+    
+    def __init__(self, *args, **kwargs):
+
+        super(RockstorIO, self).__init__(*args, **kwargs)
+        self.threads = {}
+
+    def cleanup(self, sid):
+
+        if sid in self.threads.keys():
```
schakrava: This can just be `if side in self.threads:`

src/rockstor/smart_manager/data_collector.py
```diff
@@ -49,8 +49,30 @@
...
+            gevent.killall(self.threads[sid])
+            del self.threads[sid]
+
+    def spawn(self, func, sid, *args, **kwargs):
 
-class PincardManagerNamespace(socketio.Namespace):
+        thread = gevent.spawn(func, *args, **kwargs)
+        if sid in self.threads.keys():
```
schakrava: same as my previous comment. This can be `if sid in self.threads:`

schakrava:
Looks nice in my testing too. Thank you @MFlyer ! If you could, please update based on my comments above.

MFlyer:
Hi @schakrava ,
you're right and at first had it `sid in self.threads`, but python raised a "not iterable" error: I was surprised about that so moved to `keys()`.
Trying back to simple sid in self.threads and test + commit changes

M.

MFlyer:
Ok, got it on 463cc3b and know I understand it:

Had those `keys()` while it wasn't a subclass and threads not defined in __init__
As for your suggestion removed all `keys()` and working, here we go! :) 

Happy Thanksgiving to you Suman!
M.

schakrava:
Thank you @MFlyer . We are having a nice Thanksgiving break. Turkey turned out pretty good, I'll e-mail you pictures.

My brain did not register zombY either, but in my defence, I was a zombie from all the holiday turkey consumption :)

PR #1554, opened 2016-11-29
correct disk removal disqualification calculation. Fixes #1553

### LINKED ISSUE #1553: incorrect space calculation on disk removal
When removing a disk from a pool the calculation to disqualify the action is based on if the disk begin removed is larger than the overall pool usage. Where as the action should only be disqualified if the disk being removed is larger than the pool free space: as the error message state:

To reproduce I added a 7.5 GB disk to a nearly empty (1GB used) 473 GB pool.
I was then unable to remove this 7.5 GB as it's size was > than the use space of the pool ( about 1GB), rather than the space free (pool size - usage).

```
Removing these([u'mmc-00000_0x1ceff260']) disks may shrink the pool by 7864320KB, which is greater than available free space 1081856KB. This is not supported.
```

phillxnet:
Previously we refused a disk removal based on disk > pool usage
but the intended calculation was disk > pool free space.
Hence switch usage for (pool.size - usage) = free space.

So now this action is disqualified if disk > pool free space

This accounts for a worst case scenario.

Fixes #1553 

Previously a 7.5 GB disk was not allowed to be removed from a 463 GB (2 GB used) pool.
After pr this was allowed. Logging was used during development to confirm figures were as expected.
Details and user experience are available in the linked issue #1553 .

### DIFF
--- a/src/rockstor/storageadmin/views/pool.py
+++ b/src/rockstor/storageadmin/views/pool.py
@@ -416,7 +416,7 @@ class PoolDetailView(PoolMixin, rfc.GenericView):
                 size_cut = 0
                 for d in disks:
                     size_cut += d.size
-                if (size_cut >= usage):
+                if size_cut >= (pool.size - usage):
                     e_msg = ('Removing these(%s) disks may shrink the pool by '
                              '%dKB, which is greater than available free space'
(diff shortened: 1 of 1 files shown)


phillxnet:
Ready for review.

PR #1557, opened 2016-11-30
Fixes #1556 adding Django env

### LINKED ISSUE #1556: [Bug] Upgrading to Django >= 1.7 requires enabling Django on data_collector
Hi @schakrava , need your help / confirmation for upcoming PR:

While checking Pincard manager after Django update got this while trying to create a new pincard (credit card symbol on users page clicked):
```python
  File "/opt/build/eggs/gevent-1.1.2-py2.7-linux-x86_64.egg/gevent/greenlet.py", line 534, in run
    result = self._run(*self.args, **self.kwargs)
  File "/opt/build/src/rockstor/smart_manager/data_collector.py", line 96, in create_pincard
    new_pincard = save_pincard(uid)
  File "/opt/build/src/rockstor/system/pinmanager.py", line 171, in save_pincard
    flush_pincard(uid)
  File "/opt/build/src/rockstor/system/pinmanager.py", line 163, in flush_pincard
    Pincard.objects.filter(user=int(uid)).delete()
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/query.py", line 536, in delete
    collector.collect(del_query)
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/deletion.py", line 193, in collect
    if self.can_fast_delete(objs):
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/deletion.py", line 155, in can_fast_delete
    for related in get_candidate_relations_to_delete(opts):
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/deletion.py", line 62, in get_candidate_relations_to_delete
    opts.get_fields(include_hidden=True) for opts in candidate_models
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/deletion.py", line 62, in <genexpr>
    opts.get_fields(include_hidden=True) for opts in candidate_models
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/options.py", line 740, in get_fields
    return self._get_fields(include_parents=include_parents, include_hidden=include_hidden)
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/options.py", line 802, in _get_fields
    all_fields = self._relation_tree
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/utils/functional.py", line 59, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/options.py", line 709, in _relation_tree
    return self._populate_directed_relation_graph()
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/db/models/options.py", line 681, in _populate_directed_relation_graph
    all_models = self.apps.get_models(include_auto_created=True)
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/utils/lru_cache.py", line 101, in wrapper
    result = user_function(*args, **kwds)
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/apps/registry.py", line 168, in get_models
    self.check_models_ready()
  File "/opt/build/eggs/Django-1.8.16-py2.7.egg/django/apps/registry.py", line 131, in check_models_ready
    raise AppRegistryNotReady("Models aren't loaded yet.")
AppRegistryNotReady: Models aren't loaded yet.
<Greenlet at 0x22e8cd0: create_pincard(u'11201')> failed with AppRegistryNotReady
```

Googling found that it's related to Django >=1.7 and solved over system.pinmanager this with:
```python
import django
django.setup()
```

Pinmanager code inside data_collector (remember data_collector is supervisord, not on our wsgi) relies on system.pinmanager that loads Django User model (`from django.contrib.auth.models import User as DjangoUser`), but django.setup() now seems to be required

Can you confirm this being ok? :)

Meanwhile, having PR
M.

MFlyer:
#1556 :

To @schakrava : here my PR, hope had it right way (actually Pincardmanager working fine!)

Ready to review
Mirko

### DIFF
--- a/src/rockstor/system/pinmanager.py
+++ b/src/rockstor/system/pinmanager.py
@@ -15,7 +15,8 @@ General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
-
+import django
+django.setup()
 import random, string
 from hashlib import md5
 from pwd import getpwnam


schakrava:
Hi @MFlyer this pr made me realize there's a design flaw in the pincard manager implementation. Let me explain. data-collector is a separate service, and we could use django-ORM to directly update models in the database. You are doing this now, but it creates new postgresql connections and in my experience leads to bad behaviour with resource starvation and becomes hard to debug.

The preferred way is to expose CRUD operations on Pincard model via REST API just like we do it on other resource like Disk, Pool, User etc.. Then, data-collector would use APIWrapper.api_call() to create, update and delete Pincard models. Routing it this way is better than playing hard and fast with ORM. We do this in data-collector already for other operations like updating rockons metadata etc..

Using ORM directly from data-collector just for reads is usually OK, something to keep in mind as well if you just want to read something from db and there's no API...

MFlyer:
Thanks @schakrava ,
will work on this asap (asap = before you close for 3.8.16 ;) )

schakrava:
Hi @MFlyer . Please don't rush on the proper implementation. Do it thoroughly like you normally do. I am planning to close out 3.8.16 soon as I and @priyaganti finish testing and fix any important bugs we encounter in the process. So no clear estimate, but it would be great to get the release out within a week.

I am totally fine with merging this PR for 3.8.16 and replacing it with improvements after that. Sound good?

MFlyer:
Thanks @schakrava ,
thinking about Pincard via APIWrapper not sure still deciding if have it with CBV of FBV (Pincard management is alike Logs Manager, more running of custom functions then models interactions)
M.

schakrava:
Looks like this change is required regardless of our design consideration, because of a django refactor since 1.7. check this out: https://docs.djangoproject.com/en/1.10/releases/1.7/#app-loading-refactor

Issue #1194, opened 2016-03-01
Resize UI bug


schakrava:
reported by a user. [video demo](https://192.195.81.156:22080/index.php/s/NkvZJqkpqPPd6vN)

I’m allowed to resize a pool and choose change raid level but continue without selecting a raid level leading to balance errors. I should get a warning or mandate to choose before continuing forward, and if I go forward but then click back my selections should be sticky.

Error running a command. cmd = ['btrfs', 'balance', 'start', u'-mconvert=', u'-dconvert=', u'/mnt2/Testpool1']. rc = 1. stdout = ['']. stderr = ['the convert option requires an argument', '']

ChristianKueppers:
when resizing a non raid pool with a new disk i get the error "pool migration from single to single is not supported". 
Is it a bug or a limitation?

schakrava:
It's more of a feature than a bug :)

If you like to use multiple drives in single profile, you can use some kind of hardware raid or mdraid underneath.

schakrava:
Fixed by #1558

PR #1558, opened 2016-11-30
add validation to change raid level in add disks wizard view #1194

priyaganti:


schakrava:
Thanks @priyaganti . Looks very nice!

PR #1559, opened 2016-11-30
Fix for #1555 - encode passwords as utf8

### LINKED ISSUE #1555: Non-ASCII password on user creation leads to 'User(...) already exists'
Looking for possible NAS solutions and I stumbled over Rockstor today. I am already successfully using btrfs on my laptop, so this looks very interesting to me.

Anyway. When attempting to create the first user through the WebGUI wizard I used a password with non-ASCII characters which lead to the following issue. (As can be seen in the screenshots, this is reproducible when simply creating a new user.)

screenshot from 2016-11-29 22-13-30
```
  File "/opt/rockstor/src/rockstor/rest_framework_custom/generic_view.py", line 40, in _handle_exception
    yield
  File "/opt/rockstor/src/rockstor/storageadmin/views/user.py", line 152, in post
    usermod(invar['username'], invar['password'])
  File "/opt/rockstor/src/rockstor/system/users.py", line 130, in usermod
    out, err = p.communicate(input=passwd)
  File "/usr/lib64/python2.7/subprocess.py", line 800, in communicate
    return self._communicate(input)
  File "/usr/lib64/python2.7/subprocess.py", line 1401, in _communicate
    stdout, stderr = self._communicate_with_poll(input)
  File "/usr/lib64/python2.7/subprocess.py", line 1465, in _communicate_with_poll
    input_offset += os.write(fd, chunk)
UnicodeEncodeError: 'ascii' codec can't encode character u'\x**' in position 0: ordinal not in range(128)
```
_`u'\x**'` contains the hex representation of the character, censored here._

There appears to be no hint on allowed characters and no password validation whatsoever? Fair enough. Change the password to all-ASCII characters and try again:

screenshot from 2016-11-29 22-14-47

```
```

Oops.

As a workaround I used a different username. This is all for testing purposes on a VM anyway, so it's not a big deal. The username later appears in the list of `Other system users` and needs to be removed via CLI.

I'd suggest either adding validation as it appears to be done for the username or edit the python script to perform whatever encoding has to be performed before the user creation? Unfortunately, I haven't done any python yet myself ..

I am running Rockstor version `3.8.15-0`, kernel `4.6.0-1.el7.elrepo.x86_64`.

ansemjo:
Encode strings that are passed to usermod and smbpasswd to fix #1555 (caused by non-ASCII passwords)
see: http://stackoverflow.com/a/3040117

### DIFF
--- a/src/rockstor/system/users.py
+++ b/src/rockstor/system/users.py
@@ -127,7 +127,7 @@ def usermod(username, passwd):
     cmd = [PASSWD, '--stdin', username]
     p = subprocess.Popen(cmd, shell=False, stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE, stdin=subprocess.PIPE)
-    out, err = p.communicate(input=passwd)
+    out, err = p.communicate(input=passwd.encode('utf8'))
     rc = p.returncode
     if (rc != 0):
         raise CommandException(cmd, out, err, rc)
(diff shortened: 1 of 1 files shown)


schakrava:
Thank for your contribution @ansemjo and welcome to our contributor club!

I couldn't get to your pr in the last cycle, but happy to have merged it as the first pr of our new dev cycle.

Issue #1567, opened 2016-12-04
[Enhancement] data_collector move from Django ORM to CRUD operations (on db writes)


MFlyer:
Ref to #1556 (for further info check comments with @schakrava)

Data collector issue has been patched, but data_collector will have a partial rewrite to have Api Wrapper (CRUD) instead of current ORM way when writing to db

M.

PR #1569, opened 2016-12-05
Fixing Issue#1567 datacollector orm to api migration

MFlyer:
Hi @schakrava ,
had my way to #1567 following your suggestions about CRUD ops

Checking comments on commits you'll find checking pincard existence is always with direct db reading (as you suggested it possible on #1556 )

Note: Pincard view is a small one because the system/pinmanager already had all checks

Tests performed:
- Pincard creation both with a Rockstor user and root -> OK (actually on CRUD it seems faster then before)
- Password reset on login page both with normal user and root -> OK
- Password reset on login simulating wrong pins and/or OTP -> OK

Ready for your review / to be merged

Mirko

### DIFF
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -32,9 +32,8 @@ from os import path
 from sys import getsizeof
 from glob import glob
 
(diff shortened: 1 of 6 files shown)


schakrava:
Thank you @MFlyer. Nicely done as always!

Issue #1583, opened 2016-12-13
root and home shares offer delete button


phillxnet:
I think it would be safer and more consistent to disable the delete button on the root and home shares. As is done with the root/system pool delete button. This could be done by a simply name filter as a standard install will always have them. Like wise the Access Control tabs on these shares could be disabled with an appropriate message indicating they are not available for this kind of change.

phillxnet:
Note that on occasions the root share is named "root00", this has been seen in the current and last versions of the installer and when following the "Mirroring Rockstor OS using Linux Raid" install method. So we might also want to check for this subvolume name. Currently I think this is an installer quirk but one we might as well account for in this issue.

PR #1603, opened 2017-01-05
Preserve System shares root and home from being delete

MFlyer:
Ref to #1583 

Added some custom functions to shares table page and share details page to block bin and delete button

Ready to be merged :)

Mirko

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
@@ -95,7 +95,7 @@ ShareDetailsLayoutView = RockstorLayoutView.extend({
 			return appliance.get('current_appliance') == true;
 		});
         var systemShare = false;
-        if (parseInt(this.share.get('id')) < 3 && this.share.get('pool')['id'] == '1') {
+        if (parseInt(this.share.get('id')) < 3 && this.share.get('pool')['role'] == 'root') {
             systemShare = true;
         }
 		this.subviews['share-usage'] = new ShareUsageModule({ share: this.share });
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
@@ -3,7 +3,7 @@
  * @licstart  The following is the entire license notice for the
  * JavaScript code in this page.
  *
- * Copyright (c) 2012-2013 RockStor, Inc. <http://rockstor.com>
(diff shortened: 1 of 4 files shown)


MFlyer:
Note to @schakrava : had no fixing on bad tabs to let better code reading

M.

src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
```diff
@@ -94,6 +94,10 @@ ShareDetailsLayoutView = RockstorLayoutView.extend({
 		var current_appliance = this.appliances.find(function(appliance) {
 			return appliance.get('current_appliance') == true;
 		});
+        var systemShare = false;
+        if (parseInt(this.share.get('id')) < 3 && this.share.get('pool')['id'] == '1') {
```
schakrava: Can we tighten this logic @MFlyer ? It requires backend changes. There's a role attribute in the Pool model which can be updated as part of the Pool state refresh. We should rely on that attribute rather than the Pool's id being 1. Same sort of business logic should apply in identifying /home and /root Shares. In fact, the OS Pool may have a different name, but home and root almost never do. So you can just check the names. But to make it 100% solid, we can have a property in the Share model that does a runtime check if the subvolume is mounted as /home or /root.

src/rockstor/storageadmin/static/storageadmin/js/views/shares.js
```diff
@@ -138,5 +139,16 @@ SharesView = RockstorLayoutView.extend({
 			}
 			return new Handlebars.SafeString(html);
 		});
+
+        Handlebars.registerHelper('isSystemShare', function(id, pool_id) {
+
+            //During Rockstor installation we create first pool with pool_id == 1
+            //and root and home shares with id 1 and 2; this let us filter on ids
+            //instead of share names
```
schakrava: Your reasoning is correct here, but please see my previous comment. For the OS Pool, the pool.role is already set to "root" during disk scan process. Do you have any concerns using that instead of id == 1?

MFlyer: 956c66b :wink:

src/rockstor/storageadmin/static/storageadmin/js/views/share_details_layout_view.js
MFlyer: 956c66b got it on both, you're totally right - my fault, I really didn't remember about that role field
M.

schakrava:
Thanks @MFlyer !

MFlyer:
It's always a pleasure @schakrava :+1:

PR #1620, opened 2017-01-21
Show sector sizes for 512e disks. Fixes #1590

### LINKED ISSUE #1590: Sector Size empty for 512e drives
For [512e hard drives](https://en.wikipedia.org/wiki/Advanced_Format#512e), the `smartctl` information section contains  
```
Sector Sizes:     512 bytes logical, 4096 bytes physical
```

However, this doesn't match the `Sector Size` item that Rockstor expects, so the field on the Web UI is empty:
screen shot 2016-12-18 at 6 59 21 pm

For reference, `Sector Size`  as would appears for other hard drives:  
```
Sector Size:      512 bytes logical/physical
```
(Not sure about 4k native; I don't have such a device to test, and haven't come across other people's examples.)

chrstphrchvz:
Update the match string to parse smartctl output for differing logical/physical sector sizes, e.g. `Sector Sizes: 512 bytes logical/4096 bytes physical`.

For reference, the corresponding code responsible for the smartctl output is in the "Print sector sizes" section of  `print_drive_info()` in [ataprint.cpp](https://www.smartmontools.org/browser/tags/RELEASE_6_2/smartmontools/ataprint.cpp?rev=3841#L592).

### DIFF
--- a/src/rockstor/system/smart.py
+++ b/src/rockstor/system/smart.py
@@ -54,7 +54,7 @@ def info(device, custom_options='', test_mode=TESTMODE):
                'Serial Number:|Serial number:',
                'LU WWN Device Id:|Logical Unit id:',
                'Firmware Version:|Revision', 'User Capacity:',
-               'Sector Size:|Logical block size:', 'Rotation Rate:',
+               'Sector Sizes?:|Logical block size:', 'Rotation Rate:',
                'Device is:', 'ATA Version is:',
                'SATA Version is:', 'Local Time is:',
                'SMART support is:.* Available',


schakrava:

schakrava:

MFlyer:
Hi @schakrava , not able to test smart (Proxmox not with physical disks), but reading ataprint.cpp code that should be `Sector Size:` or `Sector Sizes:`, no question mark, so

```python
'Sector Size:|Sector Sizes:|Logical block size:'
```

https://www.smartmontools.org/browser/tags/RELEASE_6_2/smartmontools/ataprint.cpp?rev=3841#L589

@phillxnet ?

chrstphrchvz:
@MFlyer, the use of question mark is shorthand for optional 's' in regex, though I see the convention for `matches` (as documented in smart.py) has been to use literal text and `|` operator as @phillxnet suggested. I agree it's equally correct and slightly easier to understand (I'm not a regex expert myself), so I'm willing to revise the patch. I also have the example smartctl output I tested this with: https://gist.github.com/chrstphrchvz/a229c595423eff0e8b07627caa01ec43 (use "download as zip" and extract to /root/smartdumps/).

MFlyer:
Sorry @chrstphrchvz , your totally right about that, don't care about my last message, so on my side ok

M.

MFlyer:
Sorry @chrstphrchvz for my short reply, obviously that question mark regex is ok, just preserving smart.py `matches` coding style :)

@schakrava on my side we can go on and merge

Mirko

phillxnet:
Patch applied and regular "Sector Size:" catch
sector-size-match
patch applied and @chrstphrchvz provided dumps containing "Sector Sizes:" (using smart.py test mode):
sector-sizes-match

Well done on spotting and sorting this @chrstphrchvz.

phillxnet:
@chrstphrchvz Thanks for contributing this by the way, much appreciated.

If you include in a comment the following Fixes #1590 (as you have done in the title) without the quotes then the indicated issue will be automatically closed upon the pull request's merge. This makes for less 'dangling' issues that have already been closed and reduces the manual admin of issue management. Also gives an easy clickable link back to the associated issue for reference.

Cheers.

Oh and nice references back to upstream code people. But this isn't the only generator of that fields output as in https://github.com/rockstor/rockstor-core/commit/3c5ebe89115d50f5b1312ad8e4926f89a00ece82
the "Logical block size:" string match was added. I think this was for some scsi output we had sent in. Hopefully we are almost there now as we haven't had a mismatch reported for some time.

phillxnet:
@chrstphrchvz and @MFlyer in the interests of not letting the side down I am compelled to provide the upstream code reference for the scsi match on this same field:
https://www.smartmontools.org/browser/tags/RELEASE_6_2/smartmontools/scsiprint.cpp?rev=3841#L1516

chrstphrchvz:
Fixes #1590

schakrava:
Thank you @chrstphrchvz and @phillxnet and @MFlyer

Issue #550, opened 2014-12-07
Support full disk LUKS


schakrava:

halemmerich:
I have done some experiments using LUKS and wanted to share my experiences:
My target was a RockStor instance running a pool on encrypted disks and I did not accomplish that so far.
After creating a couple of LUKS disks mapped at /dev/mapper RockStor had a problem because the device names seemingly can only be 10 characters long. As I was using the serial number as the device mapper name the webinterface complained. Using shorter names makes the mapped disks appear in the storage view, but all operations including creation of a pool fail, because in the btrfs commands the device mapper disks are also addressed as /dev/name and not found.
I then tried to create a btrfs over the disks manually but RockStor did not pick it up correctly. The encrypted disks in the storage view are shown as having a btrfs on them, but can not be imported, probably again because of path issues.

phillxnet:
@schakrava Could we change the title of this issue from:-
"Support dmcrypt"
to
"Support full disk LUKS"
From our recent discussion I think this fits in better with our current capabilities and plans plus from: https://gitlab.com/cryptsetup/cryptsetup/wikis/FrequentlyAskedQuestions
we have:- "plain dm-crypt (one passphrase, no management, no metadata on disk) and LUKS (multiple user keys with one master key, anti-forensic features, metadata block at start of device, ...)"
So we gain management and an easy way to identify the device as LUKS encrypted so that scan_disks can hopefully provide an on the spot truth during disks Rescan (I think anyway as yet to confirm). This can then inform the db updater to set the new role flag to identify the device as LUKS formatted and hence trigger looking up it's /dev/mapper/device-name for unencrypted access.
and:
"First, unless you happen to understand the cryptographic background well, you should use LUKS. It does protect the user from a lot of common mistakes. Plain dm-crypt is for experts."
and:
"Advantages [of LUKS ] are a higher usability, automatic configuration of non-default crypto parameters, defenses against low-entropy passphrases like salting and iterated PBKDF2 passphrase hashing, the ability to change passphrases, and others."
My addition in square brackets in the last quote.

- We could later, by way of a feature issue, add a LUKS header backup button, and perhaps some additional LUKS password management.

We will also need the "cryptsetup" program installed which is available in the current repos. Just noting so it doesn't get missed.

phillxnet:
[...]

PR #1622, opened 2017-01-23
enhance disk role/management subsystem. Fixes #1494 

### LINKED ISSUE #1494: enhance disk role/management subsystem
In order to enhance the capability and flexibility of Rockstor's disk management a disk 'role' field was introduced in https://github.com/rockstor/rockstor-core/commit/cc7cfc1750cafa43b3a7c8192faa1bdce5c22f10 . This field was initially used to identify mdraid disk members in #1164 . In pr #1314 this role field was moved to a json format, as advised by @schakrava , enhancing it's general purpose use, (see commit https://github.com/rockstor/rockstor-core/commit/4935f06c99bb41976a0836390b4eb5f67c575328 ).

Given the requirements presented by  #550 (Support full disk LUKS) and #1157 (Support backup to USB devices) it is proposed that we move the existing disk identification system fully over to using the 'role' db field. Currently there is still an 'overloading' of the "parted" flag in storageadmin/views/disk.py DiskMixin _update_disk_state() to signify 'unusable' or unsupported devices or file-systems. This potentially complicates support of extended role devices such as may be used in for example external USB 'import' / 'export' type roles which may very well be partitioned. Moving over to using the role field exclusively should therefor centralise and simplify device management and classification and aid in this subsystems utility.

It is therefor proposed that the 'overloading' of the parted flag be removed and superseded by the disk role field to allow for enhanced device support. ~~In later pull requests~~ this role field could be user assigned via links within the disk page that take the place of the "Click to wipe it clean" cog that instead ascribes one or more compatible roles to a device as well as the existing option to wipe the device prior to the usual default whole disk btrfs device role.

phillxnet:
As the role system enhancements within this pr are non trivial a developers technical manual entry is planned in order to aid in further development / enhancement in this area. As such this pr messaging is constrained (in the interests of brevity) to bullet point abstractions of the changes as full explanations, and intended future use, can be reserved for the developers documentation to come.
A brief exposition of the merits of the approach taken is included.

Bullet points of changes made in this pr:

- extend the existing disk role system to accommodate general purpose use (previously used to flag mdraid members only)
- replace overloaded use of the partition property in the disks model for more specific disk categorization via the role system enhancements.
- add recognition of whole disk LUKS containers and their un-encrypted virtual devices to scan_disks()
- add recognition of bcache backing and caching device formats and their consequent virtual devices to scan_disks() (requires additional udev rules to be included in 'unofficial bcache package')
- add greater partition awareness to scan_disks() with a priority on reporting btrfs in partition.
- add filesystem within partition reporting to scan_disks() again with a priority on btrfs in partition.
- add contextual reporting of above on disk Web-UI page.
- add roles config Web-UI interface to report partition and filesystem status of a device and allow wiping of either specific partitions (once a redirect role is configured) and the whole disk. An introduction of future roles is included in the user facing text.
- add the 'User assigned' redirect role - essentially a flag that causes all disk actions to be redirected to a disk's partition.
- The disk wipe dialog was replaced by the above Roles config UI due to a required interdependence.
- all disk tables were updated to indicate redirect roles.
- disk and pool models and mechanisms were updated to be aware of both auto assigned (LUKS / bcache / root drive label) roles and User assigned roles (currently only redirect role).

The scope of this pr is larger than usual but was required for the following reasons:

- To develop and test the role system a number of auto and manual roles were required, hence the requirement to add recognition for root, LUKS, and bcache devices to prove role subsystem capability in more than one arena.
- To facilitate the testing of a user assigned role a web-ui interface was required in order to prove viability and the intended role mechanism with the interplay of auto and manual roles and added context that a redirect role, for example, brings: ie to the wipefs facility.
- The usability component of facilitating partition use via a redirect role attribution and how that tied into the existing UI also required a 'proof of concept' Web-UI.

The essence of this pr is to extend the disk management and recognition to accommodate more device types and partitions there in and to establish a mechanism by which those devices might be excluded or included contextually. The approach taken restricts Rockstor's use of partitions to one per disk. This is a majority use case in a NAS appliance. This approach also allows for the continued simplicity of a disk 'levels of analysis' approach which helps to keep the UI simple and approachable and avoids users having to be aware of partitions, except in specific use cases. One such case is the interoperability requirement faced when dealing with prior use (partitioned) external media for import or shared or intermittent use scenarios. However there is still a UI bias towards 'whole disk' for pool device members as this is a more elegant / ideal arrangement.

Testing methods employed are to follow in additional comment.

Fixes #1494 

Ready for review.

### DIFF
--- a/src/rockstor/storageadmin/views/pool.py
+++ b/src/rockstor/storageadmin/views/pool.py
@@ -51,6 +51,41 @@ class PoolMixin(object):
             e_msg = ('Disk(%s) does not exist' % d)
             handle_exception(Exception(e_msg), request)
 
+
+    @staticmethod
+    def _role_filter_disk_names(disks, request):
+        """
+        Takes a series of disk objects and filters them based on their roles.
+        For disk with an openLUKS role the openLUKS role value is substituted
+        for that disks name. This effects a name re-direction for crypt disks.
+        :param disks:  list of disks object
+        :param request:
+        :return: list of disk names post role filter processing
+        """
+        try:
+            # Build dictionary of disks with roles
+            role_disks = {d for d in disks if d.role is not None}
+            logger.debug('role_filter role only disks=%s' % role_disks)
+            # Build dictionary of crypt drives with their openLUKS role values.
+            # by using only role_disks we avoid json.load(None)
+            crypt_disks = {d.name: json.loads(d.role).get("openLUKS", None) for
+                           d in role_disks if
+                           'openLUKS' in json.loads(d.role)}
+            logger.debug('role_filter_disk_names crypt_disk=%s' % crypt_disks)
+            # Replace d.name with openLUKS mapper name for crypt type disks.
+            # Substitute openLUKS role value for crypt (openLUKS) mapped disks.
+            # Our role system stores the dm mapped /dev/disk/by-id name for
+            # /dev/mapper mount points so use that value instead name:
+            dnames = [
+                d.name if d.name not in crypt_disks else crypt_disks[d.name] for
+                d in disks]
+            logger.debug('role_filter_disk_names RETURNING=%s' % dnames)
+            return dnames
+        except:
+            e_msg = ('Problem with role filter of disks' % disks)
+            handle_exception(Exception(e_msg), request)
+
+
     @staticmethod
     def _validate_compression(request):
         compression = request.data.get('compression', 'no')
@@ -267,22 +302,7 @@ class PoolListView(PoolMixin, rfc.GenericView):
 
             compression = self._validate_compression(request)
             mnt_options = self._validate_mnt_options(request)
-            # TODO: re-name d.name to d.role openLUKS element value if present.
-            # build dictionary of crypt drives with their openLUKS role values.
-            # TODO: json.loads errors if role=None so CHECK THIS HERE
-            crypt_disks = {d.name: json.loads(d.role).get("openLUKS", None) for
-                           d in disks if 'openLUKS' in json.loads(d.role)}
-            # replace d.name with openLUKS mapper name for crypt type disks.
-            disk_info = {d.name: d.role for d in disks}
-            logger.debug('DISK_INFO HERE=%s' % disk_info)
-            logger.debug('CRYPT NAME TO PATH DICT=%s' % crypt_disks)
-            # substitute openLUKS role value for crypt (openLuks) mapped disks.
-            # Our role system stores the dm mapped /dev/disk/by-id name for
-            # /dev/mapper mount points so use that instead value instead.
-            dnames = [
-                d.name if d.name not in crypt_disks else crypt_disks[d.name] for
-                d in disks]
-            logger.debug('DNAMES PASSED TO BTRFS COMMANDS=%s' % dnames)
+            dnames = self._role_filter_disk_names(disks, request)
             p = Pool(name=pname, raid=raid_level, compression=compression,
                      mnt_options=mnt_options)
             p.save()
@@ -334,7 +354,8 @@ class PoolDetailView(PoolMixin, rfc.GenericView):
             disks = [self._validate_disk(d, request) for d in
                      request.data.get('disks', [])]
             num_new_disks = len(disks)
-            dnames = [d.name for d in disks]
+            dnames = self._role_filter_disk_names(disks, request)
+            logger.debug("POOL DETAIL VIEW PUT dnames=%s" % dnames)
             new_raid = request.data.get('raid_level', pool.raid)
             num_total_disks = (Disk.objects.filter(pool=pool).count() +
                                num_new_disks)
--- a/src/rockstor/storageadmin/views/pool.py
+++ b/src/rockstor/storageadmin/views/pool.py
@@ -16,9 +16,7 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
-"""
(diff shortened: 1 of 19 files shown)


phillxnet:
Basic recognition and UI indicators of root, LUKS containers, bcache backing and caching devices.

For brevity and to demonstrate the combination of the LUKS and bcache device recognition systems:
the LUKS containers are on top of bcache created virtual devices.

Before pr:
before-pr-bcache-and-luks-containers

After pr:
after-pr-bcache-and-luks-containers

And once the LUKS containers are opened their respective un-encrypted virtual devices for also recognised post pr. Note that these open LUKS containers can then be used in pool creation:
luks-pool-disks-page

As the focus of this pr is to establish recognition of these devices, not their configuration. The setup of these configurations is left to the relevant issue. Ie this pr is a pre-cursor / pre-requisite for #550 where the required naming conventions indicated in the last image above, will be managed by the as yet non existent UI components.

phillxnet:
Testing sequence used to validate redirect role mechanism and it's associated UI.

- yum install dosfstools
- 4GB virtio (as then 50% partitions give bare min of 2GB each) serial number "serial-1"
- parted -a optimal /dev/disk/by-id/virtio-serial-1
- mklabel msdos
- mkpart primary fat32 1 50%
- mkpart primary ext2 50% 100%
- quit
- mkfs.fat -s2 -F 32 /dev/disk/by-id/virtio-serial-1-part1
- mkfs.ext4 /dev/disk/by-id/virtio-serial-1-part2
- Then observe these partitions and their file systems in the Roles config UI.

- Set redirect role to ext4 part:
- Confirm disk marked as having redirect but not included in pool creation / resize tables.
- Wipe using roles UI now we have an active redirect role to this partition.
- Check this empty redirected partition now appears in pool create / resize tables.

- import an existing single device pool created with the following command:
- "mkfs.btrfs -L btrfs-in-partition /dev/disk/by-id/virtio-serial-1-part2" ie btrfs fs in a single partition.
- As we have an existing redirect role in play and there is now an externally created btrfs fs on this partition we are offered an import option in disks.
- Check this existing btrfs partition with active redirect role is excluded from create pool and resize tables.

- Test import of this externally created btrfs fs in a partition.

- Add a whole disk member to this pool via resize.
- Delete the now mixed pool.
- Re-import pool via whole disk member.
- Delete the freshly imported mixed pool again.
- Re-import pool via partitioned (and still active redirected role) member.

- Resize pool and remove partitioned member.
- We now have a single (whole disk) device pool.

- Resize pool and add back the prior partitioned member (fs wiped during removal).
- We again have a mixed pool of whole and partitioned members.
- Resize pool and remove the whole disk member.
- (N.B. a disk rescan is sometimes required to update disk page info, separate and prior existing issue.) see #1623.
- We are back to a single partitioned member pool.

- Delete the now single partitioned member pool.
- Wipe partition only, using role config Web-UI (this is prevented while the partition is managed by a Rockstor pool (tested)).
- Confirm in role config UI the prior btrfs fs on that partition is now wiped but the partitions remain.

- Remove redirect role on this disk (ie select Whole Disk) but don’t then delete (delete is prohibited during redirect role changes (tested)).
- Reestablish redirect role to the partition.

- Create new single device pool with this partitioned member this time within UI.
- Add a whole disk member to the pool.

- Disk wipe icon is not displayed for whole disk or partition members but role is still accessible (for partitioned member) which has the new wipe UI in so check safeguards (following block):

- Attempt to wipe the partitioned pool member:
“Selected device is part of a Rockstor managed pool. Use Pool resize to remove it from the relevant pool which in turn will wipe it's filesystem.” Backend prevention of this wipe is also in place and tested.
- Attempt to change the redirect role of this partitioned pool member:
“Existing btrfs partition found; if you wish to use the redirect role either select this btrfs partition or wipe it or the whole disk and re-assign.Redirecting is only supported to a non btrfs partition when no btrfs partition exists on the same device.”
We favour btrfs and this enables our improved user feedback on prior existing btrfs in partition members.

- Remove partitioned member from the mixed pool.
- Disk page update required due to linked issue.
- Attempt to wipe the now unused and already empty (no fs) partition - (Now allowed).
- Remove redirect role (by selecting Whole Disk) - (Now allowed).
- Return to Role UI and attempt to wipe whole disk (active selection) which also removes all partitions - (Now allowed).
- Add newly wiped whole disk (prior partitioned device) to prior pool.
- Web-UI refresh again required re linked issue.

During all states (given the known bug of #1623) appropriate device flagging was observed in both disk a pool Web-UI pages.

src/rockstor/storageadmin/views/pool.py
```diff
@@ -51,6 +53,54 @@ def _validate_disk(d, request):
             handle_exception(Exception(e_msg), request)
 
     @staticmethod
+    def _get_disk_names(disks, request):
```
schakrava: this method is not being called anywhere. Can it be removed? You can also consider list comprehension for creating lists like this.

schakrava:
It's a pleasure to review this pull request @phillxnet . Masterfully done. Sorry for taking long to review and test it. I am half way there and it's looking great.

src/rockstor/storageadmin/views/pool.py
phillxnet: It can be removed if you preferred this. I used it during development and though that it might come in again for future development so left it in place, but it can always be recreated of course.

phillxnet:
@schakrava Cheers. I'll try (again) to keep my pr's to a more manageable size. Thanks for review and advise.

PR #1626, opened 2017-01-24
Fixing Jenkins' Flake8 regressions

MFlyer:
Jenkins' Flake8 doesn't like some files, so we fix them

Ready to be merged @schakrava 

Can we have it running against PR diffs instead of all Rockstor repo structure? This can speed up flake8 task :)
Mirko

### DIFF
--- a/src/rockstor/scripts/initrock.py
+++ b/src/rockstor/scripts/initrock.py
@@ -468,5 +468,6 @@ def main():
     enable_rockstor_service(logging)
     enable_bootstrap_service(logging)
 
+
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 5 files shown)


MFlyer:
Add about building:

http://192.195.81.208:22082/job/build-rockstor-core/31/console

Yesterday had same issue while building share usage commits, had to bootstrap.py and rebuild

M.

MFlyer:
@schakrava & @phillxnet : 

After fetching PR on jenkins, instead of flake8 full Rockstor env we can check it against modified .py files:

`flake8 $(git diff master --name-only | grep -i "\.py" | tr '\n' ' ')`

git diff --name-only always returns `\n` separated files, so had to tr them

Mirko

schakrava:
Thanks @MFlyer, nicely done!

Issue #1630, opened 2017-01-26
GUI Shows 100 for Network Devices when system is using 1000


magicalyak:
The main GUI shows 100 (connected) yet ifconfig reveals the connection to be using 1000 (if you run cat /sys/class/net/{interface}/speed it shows 2000 for my bond0 (it's a LACP bond yet the GUI shows 100).  I suspect it's using something like mii-tool or something old that doesn't have fast ethernet (so it never goes above 100).  If the network GUI could leverage running cat on /sys/class/net/{interface}/speed it should be perfect.

Or...someone forgot to add a 0

phillxnet:
Having a quick look into this now:
From initial investigation the relevant base network info is parsed in src/rockstor/system/network.py in devices() via the following commands:
```
nmcli -t -f device device
```
to first get a list of the devices, and then each of these devices is fed into:
```
nmcli d show devname
```
In an example on real hw where ifconfig shows 1000 on a dev named enp3s0 in this instance we have the output here as:
```
nmcli d show enp3s0 | grep 'GENERAL.STATE:'
GENERAL.STATE:                          100 (connected)
...
So we are parsing nmcli's output correctly.

What is shown is not connection speed but a status code (condition) report.
ie in the following we have the state code of 20 (unavailable):

![network-state-code-report](https://cloud.githubusercontent.com/assets/2521585/22405410/0b41e60a-e63a-11e6-84e0-52ef3cce6335.png)

So to address this issue I have opened an additional issue to specifically address the observed miscommunication of the 'State' title as the column header: #1633 .

@magicalyak An additional feature request issue could be opened to request that interfaces and or connections speeds be reported in an addition column to the existing tables. Please feel free to open such an issue and then link back to this issue if you think that is appropriate. I think this issue can now be closed given the specificity of #1633 which attributes you and the associated forum thread with this lake of clarity finding.

Linking back to the forum thread that initialted this issue:
https://forum.rockstor.com/t/jumbo-packets-possibility/2706/3

phillxnet:
@magicalyak As pr #1634 (for spin off issue #1633)  has now been merged I think this issue can be closed. Please do so if you agree and do open a feature request issue re network device speed if you think that is appropriate.

phillxnet:
@magicalyak and @schakrava I think this one can now be close re my last comment here.

schakrava:
Thanks for your input @phillxnet

PR #1634, opened 2017-01-29
improve state column clarity in network device tables. Fixes #1633

### LINKED ISSUE #1633: improve state column clarity in network device tables
Thanks to Forum members Karney_Chen and @magicalyak in the following forum thread for highlighting this issue. In network device tables a 'State' column reports device state codes: these can be misinterpreted as network device speeds due to a common state code (connected) being 100, which is also a well known ethernet speed. However this column is parsed from nmcli's GENERAL.STATE output and is simply a status code.

Linking to issue #1630 where the source of this columns contents was established.

Forum thread associated with this and the above referenced issue:
https://forum.rockstor.com/t/jumbo-packets-possibility/2706/3

I propose that we change the network device table header from "State" to "Status code" there by clarifying the meaning of the codes there in.

phillxnet:
To avoid an observed misunderstanding of the 'State' column contents in network device tables we rename it to 'Status Code' as this more accurately reflects what this is and helps to avoid the confusion observed by several parties in the linked forum thread and issue:

https://forum.rockstor.com/t/jumbo-packets-possibility/2706/5
The issue this pr addresses was opened as a result of work done in #1630 by myself and @magicalyak .

Patch is tested against current master after being built from scratch.

Please see the first commit for intended user visible column text changes as later formatting changes are also made.

Fixes #1633

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/network/network.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/network/network.jst
@@ -1,149 +1,170 @@
 <script>
-/*
- * Copyright (c) 2012-2013 RockStor, Inc. <http://rockstor.com>
- * This file is part of RockStor.
- *
- * RockStor is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published
- * by the Free Software Foundation; either version 2 of the License,
- * or (at your option) any later version.
- *
- * RockStor is distributed in the hope that it will be useful, but
- * WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


magicalyak:
I didn't think of that! That makes a lot more sense.  Ideally trimming the numbers would probably be best in my opinion. Should I add a rfe for a speed column? After all isn't it always about how fast you go? ;)

phillxnet:
@magicalyak I don't think throwing away any info is necessary here really and actually complicates things a little. Best we stick with showing what nmcli is putting out but yes to your other point: definitely open an issue as a feature request for device / connection speeds column.

Thanks for your input on these issues by the way.

schakrava:
Thank you @phillxnet !

phillxnet:
@schakrava Cheers.

Issue #1632, opened 2017-01-29
[Rockstor Devel Feature] Add Gulp file testing


MFlyer:
Migrating from #1629 (closing it) to have a clean new issue

commits won't have any ref to issue to have it clean and with a different approach, starting from a working gulp + eslint without checks on unused vars (we have to decide how to have it / hack it all Rockstor js like a unique file)

@schakrava , @phillxnet , @priyaganti

MFlyer:
Reference, Django Pipeline:

```
PIPELINE = {
    'DISABLE_WRAPPER': True,
    'JS_COMPRESSOR': None,
    'CSS_COMPRESSOR': None,
    'TEMPLATE_FUNC': 'Handlebars.compile',
    'JAVASCRIPT': {
        'storageadmin': {
            'source_filenames': (
                'storageadmin/js/license.js',
                'storageadmin/js/templates/**/*.jst',
                'storageadmin/js/templates/**/**/*.jst',
                'storageadmin/js/socket_listen.js',
                'storageadmin/js/rockstor.js',
                'storageadmin/js/rockstor_widgets.js',
                'storageadmin/js/rockstor_logger.js',
                'storageadmin/js/paginated_collection.js',
                'storageadmin/js/router.js',
                'storageadmin/js/graph.js',
                'storageadmin/js/d3.slider2.js',
                'storageadmin/js/models/models.js',
                'storageadmin/js/views/common/*.js',
                'storageadmin/js/views/*.js',
                'storageadmin/js/views/pool/**/*.js',
                'storageadmin/js/views/dashboard/*.js',
                ),
            'output_filename': 'storageadmin/js/storageadmin.js'
            },
        }
}
```

PR #1635, opened 2017-01-30
Gulp (Eslint) session 1

MFlyer:
Ref #1632 
Hi @schakrava , @phillxnet and @priyaganti ,
this is the first eslint task over Rockstor js files, covering minor ones like indentation, single/double quotes, etc

Can you please try this and check it? I'll do it too, but having over 90 files I think more testers can grant a better result :)

Mirko


MFlyer:
Testing had only 6c5c3a6 and e28d9a6 fixing a small regression on unuseful console.log

On my side it seems ok and ready to be merged @schakrava

MFlyer:
Suggestion: merge this after @phillxnet huge disk role PR, going to rebase and lint again :)

src/rockstor/storageadmin/static/storageadmin/js/views/setrole_disks.js
```diff
@@ -233,15 +233,15 @@ SetroleDiskView = RockstorLayoutView.extend({
...
+                uuid_message = 'btrfs'
             } else {
                 uuid_message = 'None';
             }
             var selected = '';
             if (this.current_redirect == '') {
                 selected = ' selected="selected"';
             }
-            html += '<option value=""' + selected + '> Whole Disk (' + uuid_message + ')';
+            html += '<option value="' + selected + '"> Whole Disk (' + uuid_message + ')';
```
MFlyer: Here the typo!

MFlyer:
Hi @schakrava , now this is ready to be merged after @phillxnet disk role :)

Rebase : Had conflicts solving over Github, disks.js had 3 conflicts so to avoid messing it had a copy&paste from current master file then a new eslint 😉 

Mirko

MFlyer:
Notes about Eslint:

- indentation : 4 spaces (having it pythonic way)
- strings : single quotes, not double quotes
- redeclarating vars discouraged:

this raises an eslint error
```javascript
if (somenthing) {
    var a = false;
} else {
    var a = true;
}
```
this works:
```javascript
var a;
if (somenthing) {
    a = false;
} else {
    a = true;
}
```
Alternatives:
- Option A
```javascript
var a ='my init value';
if (something) {
    a = 'alternative value';
}
```
- Option B
```javascript
var a = (function(){
if (something) {
return 'this value';
} else {
return 'that value';
}
})();
```

src/rockstor/storageadmin/static/storageadmin/js/views/setrole_disks.js
phillxnet: @MFlyer I think this was 'as intended' ie not a typo as for a "Whole Disk" entry the value to be returned will = "" ie empty string. It's used as an indicator in the backend to signal no redirect: ie Whole Disk. Your change will break that mechanism. Can you double check your changes just in case. Ie first entry in this drop down list is 'hard wired' to have a value of "" to always offer the option to remove / (have no) redirect role defined.

It would be a shame to lose the option to define whole disk as an option :)

phillxnet:
@MFlyer I don't think this is ready to merged just yet as I am in doubt about your changes to my recent setrole_disks.js file. I have indicated where I think your changes may break what was tested as working as intended prior to submission. Apologies if I have this wrong but just trying to avoid breakage.

Please confirm either way so we can nail this one down.

src/rockstor/storageadmin/static/storageadmin/js/views/setrole_disks.js
MFlyer: @phillxnet sorry, you're right, going to move it back to original!

M.

phillxnet: @MFlyer OK, cheers; I should have documented this better in the first place.

phillxnet:
@MFlyer Cheers for addressing that and thanks for cleaning up my recent js var declaration style there. I've only reviewed this pr with respect to the file indicated unfortunately.

Also I don't myself like the Option B (overly convoluted) in "Alternatives" but Option A or it's variation "this works" are more readable to me.

MFlyer:
> `Also I don't myself like the Option B (overly convoluted) in "Alternatives" but Option A or it's variation "this works" are more readable to me.`

I'm on A style too, probably we're "old school" eheh

MFlyer:
a478747 it's a relinting, every time we merge another PR with js adds have rebase & relint eheh

schakrava:
Hi @MFlyer . Thanks for picking this up. It's important that we add js linting to our checks. On a cursory look, your changes look correct but I want to make sure to setup the jenkins job and all of us to be on the same page about running the linter with an agreed upon config file in our dev env.

Here's the config I've generated using `eslint --init`

```
$ cat .eslintrc.json 
{
    "env": {
        "browser": true,
        "es6": true
    },
    "extends": "eslint:recommended",
    "rules": {
        "indent": [
            "error",
            4
        ],
        "linebreak-style": [
            "error",
            "unix"
        ],
        "quotes": [
            "error",
            "single"
        ],
        "semi": [
            "error",
            "always"
        ]
    }
}
```
Please share your thoughts on config, how you run the linter and what you recommend for our jenkins.

MFlyer:
Hi @schakrava ,
this is my current eslintrc.js

```javascript
module.exports = {
    'env': {
        'browser': true,
        'jquery': true,
        'es6': true
    },
    'extends': 'eslint:recommended',
    'rules': {
        'strict': 'off',
        'no-unused-vars': 'off',
        'no-undef': 'off',
        'indent': [
            'error',
            4
        ],
        'linebreak-style': [
            'error',
            'unix'
        ],
        'quotes': [
            'error',
            'single'
        ],
        'semi': [
            'error',
            'always'
        ],
        'no-console': [
            'error',
            {
                'allow': [
                    'log'
                ]
            }
        ]
    },
    'globals': {
        // Rockstor jslibs dependencies
        'Backbone': false,
        'Handlebars': false,
        '_': false,
        'd3': false,
        'humanize': false,
        'io': false,
        'moment': false,
        'Chart': false,
        'options': true
    }
};
```
Notes:
- jquery is a built-in eslint env, so we add it to avoid errors on `$` and let it peacefully parse every jQuery obj
- globals section defines some global js lib we use and not avail like eslint environments : moment, Chart, d3, humanize, etc (read closing note)
- rules section:

1.  strict off -> http://eslint.org/docs/rules/strict (we avoid strict js)
no-unused-vars off, we can have a `var something = 0` and never use it -> http://eslint.org/docs/rules/no-unused-vars (read closing note)
2. no-undef off this skip all our Rockstor global vars (ex. views, etc) and avoid tons of errors -> http://eslint.org/docs/rules/no-undef
3. indent 4 error : 4 spaces to indent, raise error (eslint can raise error or warnings) if different -> http://eslint.org/docs/rules/indent
4. quotes single error: like on indent, we use single quotes for strings -> http://eslint.org/docs/rules/quotes
5. semi forcing use of semicolons (we can skip this) -> http://eslint.org/docs/rules/semi
6. no-console raises errors is we use console, but let us have console.log() func -> http://eslint.org/docs/rules/no-console

Eslint has more options we can set, but probably we have to decide how we want it, let me explain:
over Rockstor Django we pipeline all js files together making them a unique big js file, so `var x` defined over rockstor.js is available (global var) on every .js file coming next (ex. views js files)
From what I understand Eslint can parse a list of files, but every file is a separate process (back to unused-vars and globals options I had to skip over eslint or, thinking about globals, we can solve adding a complete list of global vars - huge!) - Eslint supports multiple files being a unique lib if running with node.js style (ex. check chart.js github project with `require(module)` etc etc)

Here probably Gulp helps:

On this pr there's a gulpfile.js:

```javascript
var gulp = require('gulp');
var eslint = require('gulp-eslint');


gulp.task('lint', lintJob);
gulp.task('default', ['lint']);

function lintJob() {

    var jssrc = 'src/rockstor/storageadmin/static/storageadmin/js/';
    var jsfiles = [
        'license.js',
        'socket_listen.js',
        'rockstor.js',
        'rockstor_widgets.js',
        'rockstor_logger.js',
        'paginated_collection.js',
        'router.js',
        'graph.js',
        'd3.slider2.js',
        'models/models.js',
        'views/common/*.js',
        'views/*.js',
        'views/pool/**/*.js',
        'views/dashboard/*.js'
    ];
    jsfiles = jsfiles.map(function(element) {
        return jssrc + element
    });


    return gulp.src(jsfiles)
        .pipe(eslint('./.eslintrc.js'))
        .pipe(eslint.format())
        .pipe(eslint.failAfterError());
}
```

Gulp gives developers tools to perform tasks: with this gulp file we get all our js files (like on Django Pipeline) and perform eslint on every file (running `gulp` without args has default to eslint too, `gulp eslint` exec eslint, `gulp something` performs tasks named `something`), but we still have them alone.
Gulp has an add-on to build multiple file in a single file and we should use it to eslint output.js file, but we need some hacks to track errors to single input files.

Finally, actually we can have a basic Jenkins job checking common errors (indent, quotes, etc), but probably we should plan a more complex Gulp job (input files -> output.js -> eslint on outpit.js with refs to single Rockstor js files reference)

Gulp: better then a direct eslint command although checking single files, flake8 checks can under Gulp too

Mirko

schakrava:
Thank you @MFlyer , very well done! I have updated our buildout workflow to include gulp. 

You've done an impressive job cleaning up the code. It's a wonderful contribution that raises the quality of our project and sets a great example to contributors making more changes. Thank you so much for that.

I do have general concerns when we bring in new dependencies. Perhaps gulp is a great tool and I may learn to love it over time. However, we already have buildout, our swiss army knife. I like to leverage it as much as possible and minimize external dependencies. I know it well and it rarely has any issues as things change around it. So naturally, I'd like to limit gulp's role to things we cannot accomplish easily with buildout. I wouldn't want gulp to run flake8 or any other command we can run with buildout, for example. Keep this in mind as you work on more build related tasks.

I'll leave it up to you to decide if we really need gulp if all we need it to do is linting our js code. As you can see, your code is merged and we started using gulp :)

MFlyer:
Hi @schakrava ,
thanks for your kind words :)

Back to gulp / not gulp, IMO we're not looking same thing, let me explain.
I consider buildout part of Rockstor (no buildout->no Rockstor), while gulp & eslint like developer tools (no gulp/eslint -> Rockstor running, perhaps with some not clean code, but running!), so my idea to add gulp was more on a "git repo POV" rather then "Rockstor machine POV" and i did not think you would have added gulp to buildout 😆 

My suggestion - because I agree with you about using buildout! : back to my first idea, I'm for no gulp over buildout and leave it only on a "git repo POV" (developers can run it from current repo branch instead of running it over Rockstor dev VM, checking code before PR like we do with flake8), if we want linting on buildout I agree we can perform this with buildout - I'm not a buildout "master", so open to suggestions :)

PR #1637, opened 2017-01-30
Fixes #1636 adding Rock-on ver both on avail and installed Rock-ons

### LINKED ISSUE #1636: [Enhancement - Rockons] Add current Rock-on version on Rock-ons table

MFlyer:
#1636 

Ready to be merged

To @schakrava : don't care about indentation, etc, will rebase Gulp #1635 PR after this and others merged :)

Mirko

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
@@ -324,7 +324,7 @@ RockonsView = RockstorLayoutView.extend({
 		    html += '<div class="col-md-6">';
 		    html += '<a href="' + rockon.get('website') + '" target="_blank">';
             html += '<h3 style="display:inline"><u>' + rockon.get('name') + '</u></h3></a>';
-            html += ' (Rockstor version: ' + rockon.get('version') + ')';
+            html += ' (Rock-on file version: ' + rockon.get('version') + ')';
 		    html += '<p>' + rockon.get('description') + '</p>';
 		    html += '<h4>Current status: ' + rockon.get('status') + '</h4>';
 		    html += '</div>';
@@ -381,7 +381,7 @@ RockonsView = RockstorLayoutView.extend({
 		    html += '<div class="col-md-12">';
 		    html += '<a href="' + rockon.get('website') + '" target="_blank">';
             html += '<h3 style="display:inline">' + rockon.get('name') + '</h3></a>';
-            html += ' (Rockstor version: ' + rockon.get('version') + ')';
+            html += ' (Rock-on file version: ' + rockon.get('version') + ')';
 		    html +=  '<p>' + rockon.get('description') + '</p>';
 		    if (rockon.get('state') == 'install_failed')  {
 			html += '<strong>Failed to install in the previous attempt.</strong> Here\'s how you can proceed.';
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
@@ -322,7 +322,9 @@ RockonsView = RockstorLayoutView.extend({
 		    }
 		    html += '<div class="row">';
 		    html += '<div class="col-md-6">';
-		    html += '<a href="' + rockon.get('website') + '" target="_blank"><h3><u>' + rockon.get('name') + '</u></h3></a>';
+		    html += '<a href="' + rockon.get('website') + '" target="_blank">';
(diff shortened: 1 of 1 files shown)


glenngould82:
Great!

phillxnet:
@MFlyer This is a great addition, well done.

However do you think it would be better to specify that this is a Rock-on file version not a docker image or installed software version.

Ie: change "version" to "Rock-on file version" or the like as we seem to have plenty of space there.

Otherwise it reads like the installed software is at this particular version:
ie:
post-patch-installed-rockon

Where as the installed Plex version in that case is the latest:
plex-version-installed

This will hopefully avoid users thinking we are displaying the version of the actual software (which is I think what is desired in the issue's linked thread):
https://forum.rockstor.com/t/rock-on-requests/41/82
which is quite another issue potentially.

MFlyer:
You're right about that @phillxnet , patching

Mirko

phillxnet:
@MFlyer I don't think that "Rockstor version" covers it really as it has the same fault. 
ie Rockstor has version 1.0 of Plex. That would be a 'No No' for many looking in on the project.
Specifying that it is the Rock-on definition file version, which is actually what it is, is the important part here. In a sense this is semi useless user facing info but has some merit as at least then we and users can track significant changes in these definitions and in the future provide definition file update flags for example. Hence my previous specificity on "Rock-on file version". I know it has been used in a differing sense in the vast minority of Rock-ons but over all that seems to be what it is currently used to define.

Do you agree?

My worry is that false / misleading info is worse than no info. And up-to-date plugin availability is a major feature.

MFlyer:
Totally agree on that Phil, meanwhile had some checks over Rock-ons(docker images) and we can't grant docker image version (you can collect it after downloading docker image with `docker inspect yourimage`, but there's no guarantee about it returning docker image version)

phillxnet:
@MFlyer Cool. Yes it's a difficult one really as the user is really only interested in actual main software version. Either we maintain versions on each and every update which is obviously impractical or we add some kind of additional (optional) mechanism to Rock-ons that returns the version. I.e. a specific command for each rock-on that will return a string of the currently installed main software version. Then during an update we execute this on all running containers and display the returned string, if any, and grey out otherwise or something. Non trivial but would then always reflect the installed version. Doesn't provide version of uninstalled software of course.

phillxnet:
@MFlyer Just has a little dabble but in the case of a running Plex container we have:
```
docker exec -it plex-linuxserver.io /bin/bash -c "LD_LIBRARY_PATH=/usr/lib/plexmediaserver /usr/lib/plexmediaserver/Plex\ Media\ Server --version"
```
which returns a usable:
```
v1.3.3.3148-b38628e
```
So we could add an optional field to the rock-ons definition file that contains a command which returns the main program's version number. In the above case it would be:
```
/bin/bash -c "LD_LIBRARY_PATH=/usr/lib/plexmediaserver /usr/lib/plexmediaserver/Plex\ Media\ Server --version"
```
Obviously would be nice to omit the "/bin/bash -c" but containers will have varying shell programs and locations.

How does this look as an idea for a future issue/pr for 'program version' ?

That way each rock-on definition file could be updated (if possible) with this optional field and then start returning the live version running in that container. And if not included or not available we could direct the user to view the programs UI (via the button) for installed version info. Possibly stashing the last read for containers that have been run in the past.

The the case of "docker inspect plex-linuxserver.io" there is no useful info re plex version, hence the suggestion for the above extension.

Fancy opening an issue and we paste the "program version" stuff from here in there as this issue / pr is essentially done re Rock-on file version.

MFlyer:
Well @phillxnet , that is a good point, but we had to customize it for every single image :-/

phillxnet:
@MFlyer Yes; but how else do we do it? Only really a 'run this for version number' entry in each Rock-on definition with the fail over instructions if the Rock-on doesn't have this entry or it fails. And each Rock-on file is customised for it's specific image already.

MFlyer:
@phillxnet I know there's no other way, I was just thinking something like "what the hell, nobody never 	thought about a version tracking feature in docker" eheh

schakrava:
Thanks for the effort @MFlyer and for adding your thoughts @phillxnet . Since we don't really have a versioning system for Rock-ons, this information is largely useless. At this point we don't have any guidelines or method for incremental updates of rock-ons. It's been in MVP stage, but getting a lot of usage :)

Also note that a Rock-on is a set of one or more docker images configured and deployed to run together. So showing docker image version is ambiguous. This pr is a good start, but we need to think broadly on this one.

MFlyer:
Spring cleaning, closing this one :)

Mirko

Issue #1585, opened 2016-12-14
keyerror in samba config restore


phillxnet:
Under certain circumstance, when restoring a previously saved config backup, the samba service config fails to get restored: a key error is logged:
```
ERROR [storageadmin.util:47] exception: Samba could not be configured. Try again. Exception: 'global_config'
Traceback (most recent call last): File "/opt/rockstor/src/rockstor/smart_manager/views/samba_service.py", line 61, in post
    gc_lines = config['global_config'].split('\n')
KeyError: 'global_config'
DEBUG [storageadmin.util:48] Current Rockstor version: 3.8.16-0
ERROR [storageadmin.views.config_backup:51] Exception occured while creating resource: https://localhost/api/sm/services/smb/config. payload: {'config': {u'workgroup': u'MYGROUP2'}}. 
exception: 500 Server Error: INTERNAL SERVER ERROR. Moving on.
```
Note in the above log that the only config element was that of the workgroup. This is suspected as the trigger to reproduce the issue.

MFlyer:
Hi @phillxnet, probably this issue starts when saving config: samba view had lots of mods not reproduced in config backup func. Will check it

MFlyer:
Having a complete restore (never had it before this) I got a bounch of errors :/

```
[15/Dec/2016 11:24:05] DEBUG [storageadmin.views.config_backup:242] restore starting...
[15/Dec/2016 11:24:05] DEBUG [storageadmin.views.config_backup:244] restore submitted...
[15/Dec/2016 11:24:05] DEBUG [storageadmin.views.config_backup:55] Started restoring users and groups.
[15/Dec/2016 11:24:05] ERROR [storageadmin.views.config_backup:51] Exception occured while creating resource: https://localhost/api/groups. p        ayload: {u'admin': True, u'groupname': u'rockstor', u'gid': 11201}. exception: HTTPSConnectionPool(host='localhost', port=443): Max retries e        xceeded with url: /o/token/ (Caused by <class 'socket.error'>: [Errno 111] Connection refused). Moving on.
[...]

PR #1639, opened 2017-02-02
Samba config restore, missing checks over configuration origin (WebUI or config restore system)

MFlyer:
#1585 issue by @phillxnet ,
added control over config origin (samba config page or system config restore) with if...else to handle data

Ready to be merged, tested with:
- config saving over WebUI -> OK
- config backup and restore -> OK

Mirko


schakrava:
Thank you @MFlyer . Nice catch!

Issue #821, opened 2015-08-24
support translations


schakrava:
There's some ground work needed both on the django side and backbone side before messages can be properly translated.

paul-aviles:
Let me know when you are ready of if I can do anything to help.

Paul Aviles

phillxnet:
Connecting to recent forum thread in case it helps:
http://forum.rockstor.com/t/any-ideas-on-the-translations/887

phillxnet:
linking to another more recent forum thread on the same:
https://forum.rockstor.com/t/rockstor-locale/1282

MFlyer:
https://github.com/wikimedia/jquery.i18n

MFlyer:
https://phraseapp.com/blog/posts/jquery-i18n-the-advanced-guide/
Adding usefule reference, especially for Data API (data-i18n attrib on html tags), not well documented over wikimedia repo

MFlyer:
![sample](https://cloud.githubusercontent.com/assets/17988902/22645088/8f8435e0-ec65-11e6-9967-ae7fbc3165c4.jpg)

Starting tests with jquery.i18n and et voilà "Altri utenti di Sistema" (other system users)

M.

MFlyer:
Adding ref to testing repo for i18n : https://github.com/MFlyer/rockstor-i18n

My idea is to have a Github repo where to share language files, letting users to `git commit` directly from Rockstor WebUI

M.

PR #1643, opened 2017-02-06
Rockstor Translations

MFlyer:
Ref to #821
Hi @phillxnet @schakrava @priyaganti ,
we can merge this to have a starting point (had some tests).

Having translations over all Rockstor requires adding data apis to every *.rst file & define every language.json lib (en not required, Rockstor current default) :)

lang attr removed to avoid fixed en lang (jquery.i18n tests lang attr and use default lang only if that missing)

This requires jslib repo PR

Mirko


schakrava:
Nice start @MFlyer . Looking forward to translation contributions. I'll come up with the proper repo division strategy. We should definitely make it as easy as possible for potential contributors to add translations.

Issue #1645, opened 2017-02-08
[Bug] Maximum count field check missing on snapshot scheduled task edit page


MFlyer:
- Had a new snapshot task, Maximum count field check was ok
- Edit to snapshot let having Maximum count field empty

Test results: snapshot failing, no infinite snapshots

```python
Traceback (most recent call last):
  File "/opt/build/bin/st-snapshot", line 39, in <module>
    sys.exit(scripts.scheduled_tasks.snapshot.main())
  File "/opt/build/src/rockstor/scripts/scheduled_tasks/snapshot.py", line 89, in main
    validate_snap_meta(meta)
  File "/opt/build/src/rockstor/scripts/scheduled_tasks/snapshot.py", line 47, in validate_snap_meta
    raise Exception('max_count is not an integer. %s' % meta)
Exception: max_count is not an integer. {u'writable': True, u'visible': True, u'prefix': u'das', u'share': u'Test', u'max_count': u''}

```

phillxnet:
@MFlyer Yet another nice find.

PR #1646, opened 2017-02-08
Fixing max count for snapshots on snapshot task edit page

MFlyer:
Ref to #1645

Fixed by adding missing id and name html attrs: backbone view already has its validator to check max count if task_type is snapshot (check was missing on edit page)

Ready to be merged @schakrava


schakrava:
Thank you @MFlyer!

MFlyer:
Cheers @schakrava

Issue #1647, opened 2017-02-08
"Comment" field not filled when editing Samba share


maxhq:
When I go to "File Sharing" - "Samba" and edit an existing share the "Comment" field is not filled with the previously entered comment (which is corrently shown in the overview table).

MFlyer:
Thanks @maxhq , solved over #1649

Mirko

maxhq:
Wow, that is fast!
Take care that you do not solve problems before I report them or we'll end up in a time loop ;-P
Thanks a lot!

MFlyer:
Hi @maxhq next time don't close issues, PR for this and other need to be merged by @schakrava (on our commits we always have one "Fixes #issuenumber" and this auto-close issues :wink: )

Mirko

maxhq:
Oh OK sorry, shall I reopen it?

MFlyer:
No problem Max, don't worry :)

PR #1649, opened 2017-02-09
Samba share comment wrong model field name

MFlyer:
Ref to #1647 
We were fetching wrong model field (comments instead of comment), this causing samba export comment not rendered while on edit page

Ready to be merged

Mirko


MFlyer:
Ref to #1635 ,
rebase and relinting required beforge merging

schakrava:
Nice catch. These improvements with your eslint changes are really making the code look beautiful. Thank you @MFlyer !

Issue #1648, opened 2017-02-08
Samba export's "Comment" field only allows ASCII chars


maxhq:
Trying to add e.g. a german umlaut (ü) into the comment field in "File Sharing" - "Samba" - "Add Samba Export" gives the following error:

```
Traceback (most recent call last):
  File "/opt/rockstor/src/rockstor/rest_framework_custom/generic_view.py", line 41, in _handle_exception
    yield
  File "/opt/rockstor/src/rockstor/storageadmin/views/samba.py", line 162, in post
    refresh_smb_config(list(SambaShare.objects.all()))
  File "/opt/rockstor/src/rockstor/system/samba.py", line 88, in refresh_smb_config
    rockstor_smb_config(tfo, exports)
  File "/opt/rockstor/src/rockstor/system/samba.py", line 60, in rockstor_smb_config
    fo.write('    comment = %s\n' % e.comment)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xfc' in position 27: ordinal not in range(128)
```

MFlyer:
Once again thanks @maxhq ,
we've fixed it :)

Mirko

maxhq:
Great!

PR #1650, opened 2017-02-09
Add UTF8 support over samba exports comment param

MFlyer:
Ref to #1648 

Ready to be merged @schakrava 

Mirko


schakrava:
Thanks @MFlyer

PR #1664, opened 2017-03-05
Removing unrequired package import from initrock.py

MFlyer:
Refs to #1663 

Ready to be merged @schakrava , it's a small regression


schakrava:
Thanks @MFlyer , thought I had removed this. Must have not committed it.

PR #1705, opened 2017-05-02
nbd devices do not support S.M.A.R.T.

KaiRo-at:
We are using a fork of Rockstor for a system that uses nbd devices. While most of the support we have for that is specific to our system, the fact that nbd devices do not support S.M.A.R.T. is something that upstream Rockstor can support well, so here's a patch for that. :)

### DIFF
--- a/src/rockstor/storageadmin/views/disk.py
+++ b/src/rockstor/storageadmin/views/disk.py
@@ -283,7 +283,7 @@ class DiskMixin(object):
                 # not have a by-id type name expected by the smart subsystem.
                 # This has only been observed in no serial virtio devices.
                 if (re.match('fake-serial-', do.serial) is not None) or (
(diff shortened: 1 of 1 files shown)


schakrava:

schakrava:

phillxnet:
@KaiRo-at Hello again, Looks good to me: given the caveat  of our other ongoing discussion re by-id names in #1704. But no harm putting this in as is.

Thanks for contributing this, most welcome.

KaiRo-at:
If anyone has a "normal" nbd device, this is still useful, I guess. Of course, if we end up having nice "by-id" names with our solution to #1704, we may need another patch here, but then I guess that may not be something to upstream, depending on how it's done.

schakrava:
yay! upstream contribution from a fork. We feel special :)

Thank you @KaiRo-at

PR #1707, opened 2017-05-02
create SHA256 certs so browser devtools don't complain all the time

KaiRo-at:
When developing things with Rocktor UI, I noticed a lot of warning about SHA1 certs in my browser's developer tools. Here's a patch that moves to SHA256, which does away with those warnings.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -148,9 +148,9 @@ recipe = plone.recipe.command
 cert_loc = ${buildout:directory}/certs
 command = if [[ ! -d ${create-cert:cert_loc} ]]; then
 	mkdir ${create-cert:cert_loc};
-	openssl req -nodes -newkey rsa:2048 -keyout ${create-cert:cert_loc}/first.key -out ${create-cert:cert_loc}/rockstor.csr -subj "/C=US/ST=Rockstor user's state/L=Rockstor user's city/O=Rockstor user/OU=Rockstor dept/CN=rockstor.user" &&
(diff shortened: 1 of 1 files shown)


schakrava:

schakrava:

phillxnet:
@KaiRo-at Thanks again for another nice patch. Looks good to me.
However I'm afraid it's not really in my area so will have to pass this over to @MFlyer again as he is much more knowledgeable in this area.

MFlyer:
@phillxnet / @schakrava : hope to check this & be back coding be this week end! (new job -> new env -> new "headaches" understanding a new django project structure! eheh)
Mirko

MFlyer:
To @schakrava & @phillxnet : tested, can be safely merged
Note: obviously this doesn't solve self-signed cert issue

Thanks ot @KaiRo-at :)

Mirko

KaiRo-at:
Yes, it doesn't solve the self-signed cert issue (on initial connection), but browser (at least Firefox) developer tools complain at every request (including the API ones) if a SHA1 cert is used, no matter is self-signed or not, and no matter if an exception exists. And that is annoying when doing development work and trying to find real messages in the Web Console. That's why I looked into this in the first place. ;-)

schakrava:
agreed @KaiRo-at . Thanks for your contribution!

schakrava:
Welcome to our contributors club! I apologize for delaying my review, but glad to have merged.

Issue #1709, opened 2017-05-06
overflow of disk.role db field


phillxnet:
Thanks to forum members wellman and scanlom for reporting and confirming this issue. In the linked forum post there is evidence of a Disk.model field being insufficiently long:
```
DataError: value too long for type character varying(256)
```
Given the only field in our Disk model that is 256 characters long is the Disk.role:
```
    role = models.CharField(max_length=256, null=True)
```
it is the most obvious candidate.

Currently the suspicion is that we have a combination of many partitions on a device with a long by-id name: resulting in an overly lengthy 'partitions' role value.

Linking to the forum thread for context:
https://forum.rockstor.com/t/drive-discovery/3053
(ongoing discussion in thread to confirm the cause of the db field overflow)

phillxnet:
From the above referenced forum thread we have confirmation of 8 partitions on a single drive where upon removal of the partitions normal function returned. Thanks to scanlom for confirming this.
The output from the requested command:
```
ls -la /dev/disk/by-id/
```
Results in a mock-up role field including just the obligatory 'partitions' role of:
```
{"partitions": {"ata-TOSHIBA_MK6461GSY_31DBT7P0T-part1": "ext4", "ata-TOSHIBA_MK6461GSY_31DBT7P0T-part2": "ext4", "ata-TOSHIBA_MK6461GSY_31DBT7P0T-part3": "ext4", "ata-TOSHIBA_MK6461GSY_31DBT7P0T-part4": "ext4", "ata-TOSHIBA_MK6461GSY_31DBT7P0T-part5": "ext4", "ata-TOSHIBA_MK6461GSY_31DBT7P0T-part6": "ext4", "ata-TOSHIBA_MK6461GSY_31DBT7P0T-part7": "ext4", "ata-TOSHIBA_MK6461GSY_31DBT7P0T-part8": "ext4"}}
```
just including the partitions role gives 408 characters: almost double the current 256 char limit.
37 char dev name and assuming 4 chars per fstype and in this case 8 partitions.
Failure point is upon the 5 partition as the first 4 partitions in the above example take 213 chars.

I have recorded a 50 char name (including partition) ie:
"usb-SanDisk_Extreme_AA011219150433452149-0:0-part3"

So per disk we have eg:
'"partitions": ' 14 chars.
'{{}}'                 4 chars for outer and inner json denominators.
and per partition we have eg:
'"by-id name"'     52 chars from the above long example and including ""
[...]

PR #1716, opened 2017-05-22
support full disk luks. Fixes #550

### LINKED ISSUE #550: Support full disk LUKS

phillxnet:
Abstract:
Adds full disk LUKS UI components to perform initial LUKS format and set boot up configuration for the resulting container; with integral keyfile generation and registration where required. The consequent LUKS volumes also gain an ‘information only’ UI component to display their last recorded status which includes their backing container device. Further enhancements were required to the disk role improvements established in pr #1622 in order to meet the requirements of the UI components added. And to meet the expected pairing of LUKS data disk capability with encrypted system disk installs (via the “Encrypt my data” install option) the disk scan subsystem was improved to correctly propagate the otherwise incompatible LUKS in partition arrangement. Disk table view iconography was extended to indicate the open/closed status of LUKS containers.

Changes made:
- Add capability to appropriately indicate system disk use on encrypted installs.
- Improve iconic disk identification / tool tip info on disks page re LUKS.
- Add LUKS format tick to role/wipe page with relevant text and title changes.
- Add dual mode LUKS config/info page: Used for both containers and volumes.
- Report/configure /etc/crypttab config (boot up config) for LUKS containers/volumes.
- Partially account for existing udev sane LUKS config arrangements.
- Report last known status of LUKS volumes, including their container device.
- Enhance role and disk scan subsystems to accommodate the above.
- Add generic ‘mapped to pool’ icon to further clarify device state re btrfs.
- Add iconic indicator of pool member’s LUKS status on all main pool tables.
- Add in context explanatory text re LUKS passphrase, boot up config, and container / volume relationship.

Existing Pathologies:
- Removing a luks volume from an existing pool results in a detached device for up to 30mins after the end of a balance. Cause is as per commit: https://github.com/phillxnet/rockstor-core/commit/085e3871b4c51ccf16067f4ae19e5d9d7c0c67dc , however as a balance is often a long running process this artifact is less crucial and given the additional anticipated complexities of the same approach as per commit only applied directly after the end of a balance it was deemed worthy of it’s own issue / pr.
- When using a non keyfile boot up configuration the system has no way to authenticate reattachment of an auto detached LUKS volume (such as when a full volume fs is wiped). And so for the time being systemd-tty-ask-password-agent is a required at the console to enable reattachment prior to redeployment. Or a reboot as the name and user text presented against the relevant options suggests. To address this inelegance only the “Auto unlock via keyfile” option is recommended via user text. This could be addressed in the future when adding manual immediate container unlock UI components.
- Due to the extended use of the role field and it’s combined use to identify partitions and their filesystem, the existing 256 character db field limit can be exceeded upon encountering partitioned Open LUKS Volumes. However this is not a supported configuration. The existing issue #1709 will address this lack of robustness by way of quadrupling the disk.role maximum field length.

Configuration caveats:
- The only recognized way to identify which device is referenced in /etc/crypttab is via the containers uuid ie 2nd column must be of the form: UUID=3efb3830-fee1-4a9e-a5c6-ea456bfc269e however this is a commonly used and robust option.
- Keyfile location is non configurable via UI: all auto generated keyfiles will be placed in /root.
- No account is made for keyslot availability or configuration. However appropriate error messages should be surfaced.
- Only locked containers can be wiped and only if they also have no current boot up configuration (ie the “No auto unlock” selection). This is a safety measure and it is enforced by backend validation and front end function (ie the “Wipe locked LUKS container” link is otherwise hidden).

Noteworthy anticipated exclamations:
- It is assumed that existing LUKS configurations, identified as custom, will experience difficulties where the chosen mapped name for the resulting LUKS volume (first column in /etc/crypttab) does not begin with “luks-”. This is not recognized by udev in the same way and so does not receive a LUKS class of by-id naming: ie will not receive a “dm-name-” prefix in /dev/disk/by-id. In this case it is advised to proceed such names with “luks-” to attain on next reboot, a udev assigned dm-name-luks- by-id name. Or have Rockstor assert the luks-{uuid} type name by way of “Boot up configuration” reconfiguration.

Testing methods employed are to follow in additional comments to this pr.

Fixes #550
Please see #550 for development history.

Ready for review.

### DIFF
--- a/base-buildout.cfg
+++ b/base-buildout.cfg
@@ -61,7 +61,7 @@ command =
 	postgresql-server postgresql-devel kernel-ml btrfs-progs rsync \
 	nfs-utils avahi netatalk smartmontools net-tools sos hdparm \
 	postfix cyrus-sasl-plain yum-cron nano usbutils pciutils shellinabox \
-	epel-release
+	epel-release cryptsetup
 
 [rpm-deps-nut]
 recipe = plone.recipe.command
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -252,8 +252,10 @@ def mount_root(pool):
         raise Exception('Cannot mount Pool(%s) as it has no disks in it.'
                         % pool.name)
     last_device = pool.disk_set.last()
+    logger.info('Mount by label failed.')
     for device in pool.disk_set.all():
         mnt_device = ('/dev/disk/by-id/%s' % device.name)
+        logger.info('Attempting mount by device (%s).' % mnt_device)
         if (os.path.exists(mnt_device)):
             mnt_cmd = [MOUNT, mnt_device, root_pool_mnt, ]
             if (len(mnt_options) > 0):
@@ -265,11 +267,13 @@ def mount_root(pool):
                 if (device.name == last_device.name):
                     # exhausted mounting using all devices in the pool
                     raise e
-                logger.error('Error mouting: %s. '
+                logger.error('Error mounting: %s. '
                              'Will try using another device.' % mnt_cmd)
                 logger.exception(e)
-    raise Exception('Failed to mount Pool(%s) due to an unknown reason.'
-                    % pool.name)
+        else:
+            logger.error('Device (%s) was not found' % mnt_device)
+    raise Exception('Failed to mount Pool(%s) due to an unknown reason. '
+                    'Command used %s' % (pool.name, mnt_cmd))
 
 
 def umount_root(root_pool_mnt):
--- a/src/rockstor/storageadmin/static/storageadmin/js/router.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/router.js
@@ -39,6 +39,7 @@ var AppRouter = Backbone.Router.extend({
         'disks/smartcustom/:diskName': 'smartcustomDrive',
         'disks/spindown/:diskName': 'spindownDrive',
         'disks/role/:diskName': 'roleDrive',
+        'disks/luks/:diskName': 'luksDrive',
         'disks/:diskName': 'showDisk',
         'pools': 'showPools',
         'pools/:poolName': 'showPool',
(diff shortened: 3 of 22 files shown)


phillxnet:
All validation mechanisms and UI contextual behaviour was tested to behave as expected and as indicated in the pr description.

Pre-pr encrypted install system drive iconography:
3 8 16-16-luks-root

Post-pr encrypted install system drive display (here also showing a btrfs in partition virtio dev):
encrypted-install-plus-btrfs-in-partition-icon

In addition a real life scenario involving purposefully slow and constrained hardware (2G RAM single core dual thread P4) with 35 GB of data on an existing 3 disk raid 1 pool totalling 225 GB was taken through the following semi typical scenario for moving an existing pool from non LUKS formatted devices to LUKS formatted devices and their consequent mapped counterpart Open LUKS Volumes.

- Click the ? for role / wipe / LUKS format page on spare Samsung blank disk.
pre-luks-3drive-pool-blank-spare
luks-format-on-roles-wipe-page

- Disk page now shows a lock icon to indicate the container.
post-lus-format-spare

- Click on the lock to set “Boot up Configuration” to “Auto unlock via keyfile”
set-container-boot-up-config

- Revisiting the above page now indicates the keyfile’s existence and the new ‘active’ setting.
confirm-keyfile-config-and-no-wipe-link

- In this example we remove one of our pool disks via the resize function.
pool-now-has-2-members

- N.B. pending existing code issue to open here re no UI feedback during this removal.
- While this is happening background systemd processes enact our crypttab configuration and our first Open LUKS Volume appears. N.B. as this is also a boot up config, a reboot is required for test it’s effect, this is done later in this test scenario.
crypttab-config-enacted-via-reboot-or-wait

- LUKS format the freshly removed from pool device via ? icon as with our first LUKS format.
- And similarly set to “auto unlock via keyfile” with associated and required passphrase entry.
- Reboot to prove auto unlock configuration and the consequent opening of the second luks container.
- Now we have a 2 disk pool which we can then add our 2 Open LUKS Container to.
two-open-luks-containers-and-volumes

- Resize pool - add disks - select our 2 Open LUKS Containers.
add-2-luks-volumes-to-pool

- Let balance finish.
- Half luks half plain pool so remove the remaining non LUKS pool member via resize.
pool-disk-table-two-of-each-type
pools-table-two-of-each-type
half-luks-pool-members

- Pool - resize - remove non luks disks one at a time in this case due to size constraints leaving each process to finish.
- N.B. re above N.B. on pending existing code issue on UI feedback failure during this removal.
- A disk page refresh and we have our expected all luks pool:
pool-now-on-luks-formatted-disks

schakrava:
Amazing job @phillxnet. Nice to see a PR from you again. Thoroughly done as usual :)

phillxnet:
@schakrava Thanks and Cheers.

Issue #1547, opened 2016-11-27
insufficient use of btrfs device scan


phillxnet:
Currently we only execute this command on boot via bootstrap.py btrfs/device_scan(). In certain, currently rare, circumstance this can lead to failed mounts with multi-volume filesystems when mounting by label as we do: https://btrfs.wiki.kernel.org/index.php/Problem_FAQ#Filesystem_can.27t_be_mounted_by_label

    "... if one volume of a multi-volume filesystem fails when mounting, but the other succeeds:

```
# mount /dev/sda1 /mnt/fs
mount: wrong fs type, bad option, bad superblock on /dev/sdd2,
      missing codepage or helper program, or other error
      In some cases useful info is found in syslog - try
      dmesg | tail  or so
# mount /dev/sdb1 /mnt/fs
#

Then you need to ensure that you run a btrfs device scan first:
# btrfs device scan
```
Rockstor log excerpt of the same:
```
[27/Nov/2016 16:45:29] ERROR [storageadmin.views.command:87] Exception while refreshing state for Pool(luks-pool-on-bcache). Moving on: Error running a command. cmd = ['/bin/mount', '/dev/disk/by-label/luks-pool-on-bcache', '/mnt2/luks-pool-on-bcache', '-o', ',compress=no']. rc = 32. stdout = ['']. stderr = ['mount: wrong fs type, bad option, bad superblock on /dev/mapper/luks-3efb3830-fee1-4a9e-a5c6-ea456bfc269e,', '       missing codepage or helper program, or other error', '', '       In some cases useful info is found in syslog - try', '       dmesg | tail or so.', '']
```
In the above instance a 2 disk LUKS full disk encrypted pool was unable to mount as it's drives had only been unlocked after boot.
```
Label: 'luks-pool-on-bcache'  uuid: bb027f88-a3d9-445c-b1d1-ffb86a103c98
	Total devices 2 FS bytes used 272.00KiB
	devid    1 size 2.00GiB used 417.12MiB path /dev/mapper/luks-3efb3830-fee1-4a9e-a5c6-ea456bfc269e
	devid    2 size 2.00GiB used 417.12MiB path /dev/mapper/luks-a47f4950-3296-4504-b9a4-2dc75681a6ad
```
And our "by-label" points to one of them as expected:
```
ls -la /dev/disk/by-label/luks-pool-on-bcache 
[...]

PR #1729, opened 2017-06-13
add ro rw degraded and skip_balance mount options. Fixes #1728

### LINKED ISSUE #1728: support ro rw degraded and skip_balance mount options
The additions of the 'ro' (read only) and 'degraded' mount options will reduce (but not remove) the need to resource the command line in data recovery / drive failure scenarios. Ie the combination of these mount options can allow for a safe degraded pool mount which in turn would allow for Web-UI directed access to data on degraded pools. And the combination of rw and degraded options will allow mounting a pool that has missing devices in order that they might be replaced or deleted if btrfs raid levels allow, and if not then a raid level change could be initiated followed directly by a then legitimate device delete: given the btrfs raid level change.

It is understood that we currently do not report pool mount status #737 and #1531 and ideally this would come first however the ongoing enhancements to pool mount logic in #1547 and the planned enhancement of #1700 have these mount options are prerequisites. And situations such as those presented in #1416 and #1199 would also benefit.

But given the current state of btrfs rw,degraded mount having certain 'one shot' limitation in low drive count for raid level situations, we may, in adding this level of flexibility, allow for more foot shooting. However we have a chicken and egg scenario here and have to start somewhere. Plus our first responsibility is to report the state of the system and allow for mount options where they are useful. It is not proposed that we default to a degraded mount.

phillxnet:
All 4 additional mount options were verified as successfully validated and the resulting mounts indicated that they were successfully applied. Tool-tip modifications were also verified as displaying as intended on both the Create Pool page and the Pool detail page . No additional validation logic was added, only the initial syntactic filter was expanded.

All option lists were alphabetically arranged to assist with future mount option additions.

N.B. Although associated test code was updated to include the new settings the tests themselves were not verified.

Fixes #1728 

Please see issue #1728 text for the context of this pull request. Primarily it is a pre-requisite for proving mount improvements in development as part of current milestone issue #1547. Please note however that #1547 in not dependant on this pr.


phillxnet:
Ready for review.

schakrava:
Nicely done @phillxnet!

phillxnet:
@schakrava Thanks.

Issue #1341, opened 2016-06-09
Extra button adding storage to a Rock-On


AngeleToR:
I've found that when adding storage to a Rock-On (Plex one FWIW), the first time the 'Add Storage' dialog only shows a 'Next' button for you to click, but second, third, ... times you are offered two buttons to continue, the 'Next' one and an 'Add Storage' one, that looks like inherited from the previous screen or something. Well, if you click on 'Add Storage' that dialog will go blank and you cannot further continue adding storage and have to close, losing all the configuration made to that point.

No critical bug, just annoying as the Rock-on doesn't lose its config.

Regards, Angel

phillxnet:
@AngeleToR Thanks for creating this issue.
I'm linking this issue back to your forum post so that we can tie the two together. Just trying to avoid duplicate issues being created when others see only the forum post.
Thanks.

Linking to related forum post (by same author):-
https://forum.rockstor.com/t/small-bug-adding-storage-to-a-rock-on/1593

gkadillak:
I'll go ahead and pick this issue up :)

schakrava:
Thanks @gkadillak

schakrava:
Fixed in #1731 . Thanks @priyaganti

PR #1731, opened 2017-06-15
remove html text for prev-page and add show method. solves #1341

priyaganti:


schakrava:
LGTM. Thanks @priyaganti and welcome back!

PR #1736, opened 2017-06-20
display config modal when rockon is not configured but switch is turned on. fixes #1579

### LINKED ISSUE #1579: [Bug]Rock-on bad behaviour when starting with no configuration
When starting Rock-on service without configuration Rockstor raises an error (right), but Rock-on service switch turns ON anyway

priyaganti:


schakrava:
thanks @priyaganti

PR #1737, opened 2017-06-20
insufficient use of btrfs device scan. Fixes #1547 

### LINKED ISSUE #1547: insufficient use of btrfs device scan

phillxnet:
In some circumstances a multi-device pool can only be mounted by a sub set of it's member devices. This is a know issue referenced in 1: below. Please see #1547 for further details. This affects 'mount by label', the method we use, as this is essentially a symbolic link to a pool member, ie pool mount by a single member. The documented work around for this issue is to first execute a:

    btrfs device scan

However this will re-scan all block devices and was considered overkill by this pr's author. Consequently our existing 'btrfs device scan' wrapper was extended to accommodate single device scans. This wrapper was then called in it's single device mode for each pool member in turn just prior to pool mount from mount_root(pool). These changes required extending the Disk model to more easily account for devices with a redirect role.

While more closely examining the associated mount_share() / mount_snap() a bug was found and fixed, using the new model method, where bye these methods would fail if the first member of their respective pools also carried a redirect role. The fix was proved using a single redirect role device pool where share creation was the trigger. This was an over site of the authors in pull request #1622 .

To evaluate the effect of the device scan changes the reproducer was the same configuration as described in the issue text #1547. This involved a pool with 2 LUKS members, each backed by bcache. Prior to the changes within this pr, if the pool members were unlocked after Rockstor's initial mount attempts had finished, the pool would there after always fail to mount (unless a degraded mount option was specified) with the resulting logged error:

```
    mount: wrong fs type, bad option, bad superblock on /dev/mapper/luks-3efb3830-fee1-4a9e-a5c6-ea456bfc269e, missing codepage or helper program, or other error
```

The success of the degraded mount option in these scenarios confirmed that a member device was not 'registered' by the btrfs subsystems. A command line reproducer of this exact scenario was recorded in issue text in order to prove the effectiveness of a device specific 'btrfs device scan' on each pool member as a system wide scan had already been proved.

Post pull request changes resulted in 100% success in the above pool mount scenario.

These changes should assist with adopting a more dynamic disk capability going forward and help towards a more robust mount mechanism.

Fixes #1547 

1: Reference to btrfs wiki post sighted:
https://btrfs.wiki.kernel.org/index.php/Problem_FAQ#Filesystem_can.27t_be_mounted_by_label

A minor code tidy (reduced code duplication) enabled by the enhanced Disk model was also included as well as unit tests to prove the new btrfs device scan wrapper performed as expected:
```
test_device_scan_all (fs.tests.test_btrfs.BTRFSTests) ... ok
test_device_scan_parameter (fs.tests.test_btrfs.BTRFSTests) ... ok
```

Ready for review.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -351,7 +351,8 @@ def mount_share(share, mnt_pt):
     if (is_mounted(mnt_pt)):
         return
     mount_root(share.pool)
-    pool_device = ('/dev/disk/by-id/%s' % share.pool.disk_set.first().name)
+    pool_device = ('/dev/disk/by-id/%s' %
+                   share.pool.disk_set.first().target_name)
     subvol_str = 'subvol=%s' % share.subvol_name
     create_tmp_dir(mnt_pt)
     toggle_path_rw(mnt_pt, rw=False)
@@ -360,7 +361,8 @@ def mount_share(share, mnt_pt):
 
 
 def mount_snap(share, snap_name, snap_mnt=None):
-    pool_device = ('/dev/disk/by-id/%s' % share.pool.disk_set.first().name)
+    pool_device = ('/dev/disk/by-id/%s' %
+                   share.pool.disk_set.first().target_name)
     share_path = ('%s%s' % (DEFAULT_MNT_DIR, share.name))
     rel_snap_path = ('.snapshots/%s/%s' % (share.name, snap_name))
     snap_path = ('%s%s/%s' %
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -240,21 +240,27 @@ def mount_root(pool):
     if (pool.compression is not None):
         if (re.search('compress', mnt_options) is None):
             mnt_options = ('%s,compress=%s' % (mnt_options, pool.compression))
+    # Prior to a mount by label attempt we call btrfs device scan on all
+    # members of our pool. This call ensures btrfs has up-to-date info on
+    # the relevant devices and avoids the potential overkill of a system wide
+    # call such as is performed in the rockstor-bootstrap service on boot.
+    # Disk.target_name ensures we observe any redirect roles.
+    device_scan([dev.target_name for dev in pool.disk_set.all()])
     if (os.path.exists(mnt_device)):
         if (len(mnt_options) > 0):
             mnt_cmd.extend(['-o', mnt_options])
         run_command(mnt_cmd)
         return root_pool_mnt
-
     # If we cannot mount by-label, let's try mounting by device; one by one
-    # until we get our first success.
+    # until we get our first success. All devices known to our pool object
+    # have already been scanned prior to our mount by label attempt above.
     if (pool.disk_set.count() < 1):
         raise Exception('Cannot mount Pool(%s) as it has no disks in it.'
                         % pool.name)
     last_device = pool.disk_set.last()
     logger.info('Mount by label failed.')
     for device in pool.disk_set.all():
-        mnt_device = ('/dev/disk/by-id/%s' % device.name)
+        mnt_device = ('/dev/disk/by-id/%s' % device.target_name)
         logger.info('Attempting mount by device (%s).' % mnt_device)
         if (os.path.exists(mnt_device)):
             mnt_cmd = [MOUNT, mnt_device, root_pool_mnt, ]
@@ -1002,8 +1008,39 @@ def balance_status(pool):
     return stats
 
 
-def device_scan():
-    return run_command([BTRFS, 'device', 'scan'])
+def device_scan(dev_byid_list=['all']):
+    """
+    When called with no parameters a 'btrfs device scan' is executed, ie a
+    system wide scan of all /dev block devices to update their btrfs status.
+    Otherwise the list of devices is iterated and a 'btrfs device scan dev'
+    is executed for each item in the passed list. Detached device names and
+    path names that don't exist are ignored.
+    :param dev_byid_list: list of byid device names (without paths) to perform
+    a 'btrfs device scan' on. If not supplied then a single element list
+    ['all'] is substituted and this flags a system wide scan request.
+    :return: (out, err, rc) of the first rc !=0 run or the last rc = 0 run.
+    """
+    out = err = ['']
+    # default to successful return code unless we find otherwise.
+    rc = 0
+    if len(dev_byid_list) > 0:
+        if dev_byid_list[0] == 'all':
+            return run_command([BTRFS, 'device', 'scan'])
+        for dev_byid in dev_byid_list:
+            if re.match('detached-', dev_byid) is not None:
+                # Skip detached devices as we know they don't exist.
+                # Potential log point for early detached device discovery.
+                continue
+            dev_byid_withpath = ('/dev/disk/by-id/%s' % dev_byid)
+            if os.path.exists(dev_byid_withpath):  # only scan existing devices
+                out, err, rc = run_command(
+                    [BTRFS, 'device', 'scan', dev_byid_withpath])
+                if rc != 0:
+                    # Return on first non zero return code.
+                    # Note that a drive specific device scan on a non btrfs
+                    # device returns 'Invalid argument'!! and rc=1.
+                    return out, err, rc
+    return out, err, rc
 
 
 def btrfs_uuid(disk):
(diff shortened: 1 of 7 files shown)


src/rockstor/fs/btrfs.py
```diff
@@ -240,21 +240,27 @@ def mount_root(pool):
...
     if (os.path.exists(mnt_device)):
         if (len(mnt_options) > 0):
             mnt_cmd.extend(['-o', mnt_options])
         run_command(mnt_cmd)
         return root_pool_mnt
-
     # If we cannot mount by-label, let's try mounting by device; one by one
-    # until we get our first success.
+    # until we get our first success. All devices known to our pool object
+    # have already been scanned prior to our mount by label attempt above.
```
schakrava: nice!

src/rockstor/fs/btrfs.py
```diff
@@ -1002,8 +1010,39 @@ def balance_status(pool):
...
+            return run_command([BTRFS, 'device', 'scan'])
+        for dev_byid in dev_byid_list:
+            if re.match('detached-', dev_byid) is not None:
+                # Skip detached devices as we know they don't exist.
+                # Potential log point for early detached device discovery.
+                continue
+            dev_byid_withpath = ('/dev/disk/by-id/%s' % dev_byid)
+            if os.path.exists(dev_byid_withpath):  # only scan existing devices
+                out, err, rc = run_command(
+                    [BTRFS, 'device', 'scan', dev_byid_withpath])
```
schakrava: even if dev_byid_withpath exists, this could potentially raise an exception. Since we are interested in the return code in both failure and success scenarios, it's best to use
 
run_command([BTRFS, 'device', 'scan', dev_byid_withpath], throw=False)

not sure if I am missing something here, but hope you'll agree with me otherwise.

phillxnet: @schakrava Thanks for the review.

I think I see what you mean however what about the prior and current use of device_scan() in main() of bootstrap.py. if we silence the exception will it not cause us to fail to catch what is described there (in the exception clause) as "... a serious problem." where upon we sys.exit(1) from bootstrap. Like wise when in development I called a partitioned device (by it's root dev name) the error was surfaced to the user, via mount_root(pool). Hence the accompanying target_name property additions in this pr, ie the surfacing of the error led to identifying the need for target_name.

However I'm a bit vague on how the exceptions cascade actually, but it looked like we needed this one to not be silenced in order to identify/surface larger problems so I left it 'as was' with no throw=False, guided in part by the existing device_scan() prior to this pr and it's use in bootstrap.py and the behaviour as described below:

I.e if we revert the disk.target_name to disk.name in mount_root():

    device_scan([dev.target_name for dev in pool.disk_set.all()])

then 'as is' (without throw=False) we get the following when creating a pool with a single partition as the only device member:

dev-scan-partitioned-by-root

because we call btrfs dev scan virtio-serial-1 instead of the intended virtio-serial-1-part2. There by indicating/surfacing the failed call and in turn the incorrect device name use.

However if we do the same but with throw=False added as indicated then we have a silent failure of an incorrect "btrfs dev scan virtio-serial-1" system call.

Apologies if I have missed you point here.

phillxnet:
@schakrava Cheers.

Re pr discussion on throw=False and your later commit adding this in https://github.com/rockstor/rockstor-core/commit/c44a799f56497c3005428d5f4fabcf24cfdb8024: I see your point and this is more 'proper' going forward (re your commit comment) and may well avoid an easy over site re reliance on command exception as issue communication so once we have a proper alert system in place your change could help ensure greater separation of concerns. That is: neater to pass up to callers and inform / log / alert (once we have the system to do so (other than CommandException)) in the long run.

Oh well, bit by bit.

Linking to related "Alerts and notifications issue" #701 whose product, in time, can be used to properly communicate exceptions in contexts such as this. At that point we can more properly interpret and propagate device_scan() rc!=0 returns in all instances. As is and for the time being, with throw=False, such exceptions will be silent however. Potentially a safer play but good to know for future debugging until we interpret device_scan()'s return values, which we currently don't. Maybe we should consider enabling run_command()'s logging on this one in the near future?

schakrava:
Hi @phillxnet , I think we both agree that silent failure of device scanning is not good when done for all devices. But my comment was specific to the logic where you are running it per device in a list. If you want to rely upon rc != 0 after running a command, you should pass throw=False. Because if rc != 0, run_command will throw an exception and short circuit your code path.

Feel free to refactor again, but for now, I've added a comment, made the change and merged it.

PR #1745, opened 2017-06-28
detached disks used in mount command. Fixes #1422

### LINKED ISSUE #1422: detached disks used in mount command
Thanks to @g6094199 in the forum for reporting this. It may be that we should simply skip all attempts to mount a 'detached' device by simply parsing it's name first and using the existing fail over mechanism of trying each of the other devices in a pool in turn. In the reporters circumstance this occurred after a 'btrfs device delete' command and the resulting pool was reported healthy. An NFS share creation via Web-UI was then attempted.

Please update the following forum thread with significant development on this issue:
https://forum.rockstor.com/t/error-running-a-command-cmd/1928

phillxnet:
To avoid the unintended use of detached devices the Disk model was extended with a custom manager to return a query set that excluded Disk.name__startswith='detached-'. This mechanism was then used to replace Pool.disk_set.all() where appropriate. The manager was made default to facilitate it's use via the related fields/Disk.pool ForeignKey relationship.

Components of this pr:

1. Disk Model extension as above.
2. Use of the above model extension to replace Pool.disk_set.all() and Pool.disk_set.first() with Pool.disk_set.attached() and Pool.disk_set.attached().first() respectively.
3. Bootstrap command (called up to 17 times from src/rockstor/scripts/bootstrap.py via /etc/systemd/system/rockstor-bootstrap.service upon encountering any exception in bootstrap) was modified via DiskMixin inheritance to be able to update disk state prior to pool state. This was also added as an external command capability.
4. Bootstrap command, as in (3) above, was also enhanced via the model extension in (1) to avoid all attempts at refreshing Pool and Share info via detached disks, including skipping pool updates whose members are all detached.
5. An essentially unrelated but minor bug fix where partitioned devices could cause snapshot rollback to fail (a simple replacement of Disk.name with Disk.target_name) which was found on a line also altered in (2) above.
6. During final testing of the above it was noticed that the api calls 'try' block executed by the rockstor-bootstrap service would always fail on the first attempt “Exception while setting access_token”. Adding an additional 2 second delay directly after APIWrapper() setup resulted in no similar exceptions on the same configuration. A 1 second delay was trialled and found to be insufficient. Testing was carried out on a clean build which included this pr’s other patches.

Besides the:
Fixes #1422 
this pr also contributes towards degraded mount capability by avoiding the use of detached members of a pool in all device queries. It also improves the bootstrap behaviour re pools whose members are all detached. There by enabling greater flexibility re scenarios such as external drive use where a device is the only member of a pool and is not permanently attached.

Ready for review.

### DIFF
--- a/src/rockstor/storageadmin/models/disk.py
+++ b/src/rockstor/storageadmin/models/disk.py
@@ -23,6 +23,16 @@ from system.osi import get_disk_power_status, read_hdparm_setting, \
     get_disk_APM_level, get_dev_temp_name
 
 
+class AttachedManager(models.Manager):
+    """Manager subclass to return only attached disks"""
+    use_for_related_fields = True
+    def attached(self):
+        # Return default queryset after excluding name="detached-*" items.
+        # Alternative lookup type __regex=r'^detached-'
+        query_set = self.get_queryset()
+        return query_set.exclude(name__startswith='detached-')
+
+
 class Disk(models.Model):
     """Pool can be null for disks that are not part of any pool currently"""
     pool = models.ForeignKey(Pool, null=True, on_delete=models.SET_NULL)
@@ -58,6 +68,10 @@ class Disk(models.Model):
     role can be Null if no flags are in use.
     """
     role = models.CharField(max_length=1024, null=True)
+
+    # The default manager is the first defined.
+    attached = AttachedManager()  # Only return attached Disk objects.
+    objects = models.Manager()  # Ensure Object manager is still accessible.
 
     @property
     def pool_name(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


src/rockstor/storageadmin/models/disk.py
```diff
@@ -23,6 +23,16 @@
     get_disk_APM_level, get_dev_temp_name
 
 
+class AttachedManager(models.Manager):
```
schakrava: very nice!

schakrava:
Really nice job @phillxnet thanks for a beautiful contribution!

phillxnet:
@schakrava Cheers. And thanks for the generous comments.

PR #1761, opened 2017-07-16
mark whole disk LVM members as unusable. Fixes #1710

### LINKED ISSUE #1710: mark whole disk LVM  members as unusable
Thanks to forum user PhillVS in the following forum post for highlighting this issue. It appears that we fail to block the use of whole disk LVM2: ie FSTYPE="LVM2_member" (Physical Volume) and instead present the devices as blank. This is obviously not desired behaviour. It is proposed that we label these devices using the role system which should automatically exclude their use in pools. A handlebars helper could then be written for the Disks page to identify this fs type and inform the UI to explain that LVM is not compatible and that use will require the drive to be wiped: akin to a partitioned device. The roles/wipe/(future luks format) page should, as is, identify the whole disk fs type and give all the appropriate warnings re 'all data lost' etc so can be used as the link for the proposed icon. It may be that we can simply expand the use of the cog (setting/action required) icon to include this class of device.

Link to the associated forum thread:
https://forum.rockstor.com/t/import-btrfs-from-omv-no-option-to-import/3217

phillxnet:
Prior to this pull request whole disk LVM2 physical volumes are not recognised and are incorrectly identified as blank disks. By introducing an 'LVM2member' whole disk role, keyed from fstype, we automatically exclude these devices from pool participation selection by our existing white list on roles approach. Once assigned the role informs a handelbars helper and both front and back end logic to enhance UI feedback accordingly.

Fixes #1710 

Ready for review.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disks_table.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/disk/disks_table.jst
@@ -50,6 +50,9 @@
         {{else if (isBcacheCdev this.role)}}
             <a href="#" class="bcache_caching_drive" data-disk-id="{{this.id}}" title="Bcache Caching Drive (UI pending)." rel="tooltip">
             <i class="glyphicon glyphicon-link"></i><i class="glyphicon glyphicon-link"></i></a>
+        {{else if (isLVM2member this.role)}}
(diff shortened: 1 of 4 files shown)


schakrava:
Thank you @phillxnet

phillxnet:
@schakrava Cheers.

PR #1770, opened 2017-07-25
rogue share.pqgroup db field fails share mounts. Fixes #1769

### LINKED ISSUE #1769: rogue share.pqgroup db field fails share mounts
Originally reported by KarstenV in the following forum thread, with the trigger unidentified at the time; in very specific (and potentially pathological) db state instances all shares (subvols) fail to mount as share_helpers.py import_shares() throws a programmatic error when volume_usage() fails to behave as expected. Having now established at least one possible trigger (specific db state) for the indicated behaviour, this issue is opened to address the robustness of the share mounting mechanism in this scenario. An attempt should be made to track the cause of the db field 'pathology' however in any case all mounts should not fail in this scenario so that is the focus of this issue.

Observed error:
```
File "/opt/rockstor/src/rockstor/storageadmin/views/command.py", line 271, in post
import_shares(p, request)
File "/opt/rockstor/src/rockstor/storageadmin/views/share_helpers.py", line 83, in import_shares
volume_usage(pool, share.qgroup, share.pqgroup)
ValueError: need more than 2 values to unpack
```
Resulting in all shares failing to import. 

A trigger for this behaviour appears to be a share.pqgroup db value of "-1/-1" which results in volume_usage returning only 2 values when import_shares, as it employs volume_usage's 3 parameter personality, expects 4 values.

Issue authors intended approach to improve robustness in this scenario:

- [x] Extend fs/tests/test_btrfs.py test_volume_usage() unit test to reproduce the observed behaviour.
- [x] Modify volume_usage to guarantee 'actual parameter count' to 'return value count' mapping and verify this mapping via the extended unit test.
- [x] Establish if the trigger db value is 'self healing', ie overwritten, with sane values there after.

Please update to the following forum thread with this issues resolution:
https://forum.rockstor.com/t/after-upgrading-to-3-8-16-11-rock-on-wont-start/2842

phillxnet:
Under some circumstances the share.pqgroup db field can contain '-1/-1', when this value was passed to volume_usage() as it's 3rd parameter it did not behave as expected.  The result was a programmatic error where callers were anticipating 4 values and only receiving 2. All shares would then fail to mount on all subsequent boots. This pr contains:

- An extension to the existing unit test for volume_usage() pertaining to the -1/-1 scenario above.
- A modification to volume_usage() to ensure it returns the expected list element count.

Post pr the added and existing volume_usage() related unit tests pass.

It was also observed that post pr the previously observed programmatic errors were no longer logged and the trigger value was confirmed as having been overwritten on the subsequent reboot. This subsequent boot also successfully mounted all shares. The origin of the -1/-1 share.pqgroup value is beyond the scope of this issue/pr set but it's adverse effect as indicated is addressed.

Fixes #1769 

Ready for review.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -752,10 +752,19 @@ def update_quota(pool, qgroup, size_bytes):
 
 
 def volume_usage(pool, volume_id, pvolume_id=None):
-
     """
     New function to collect volumes rusage and eusage instead of share_usage
     plus parent rusage and eusage (2015/* qgroup)
+    N.B. this function has 2 personalities.
+    When called with 2 parameters (pool, volume_id) it returns 2 values.
+    But with 3 parameters (pool, volume_id, pvolume_id) it returns 4 values if
+    the last parameter is != None.
+    :param pool: Pool object
+    :param volume_id: qgroupid eg '0/261'
+    :param pvolume_id: qgroupid eg '2015/4'
+    :return: list of len 2 (when pvolul_id=None) or 4 elements. The first 2
+    pertain to the qgroupid=volume_id the second 2, if present, are for the
+    qgroupid=pvolume_id. I.e [rfer, excl, rfer, excl]
     """
     # Obtain path to share in pool, this preserved because
     # granting pool exists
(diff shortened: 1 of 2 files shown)


schakrava:
Thank you @phillxnet

phillxnet:
@schakrava Cheers.

PR #1772, opened 2017-07-26
fix typo of variabled to variable

jmhardison:
noticed this typo during debug of a custom Rockon.

Variabled should be variable.


schakrava:

schakrava:

PR #1777, opened 2017-07-29
fix scrub in progress message format regression. Fixes #1776

### LINKED ISSUE #1776: fix scrub in progress message format regression
Thanks to forum member Christian_Rost in the following forum thread for reporting this issue. When a scrub operation is, or is assumed to be, in progress, and another scrub task is requested via the UI. The message presented to the user should contain the pid of the pool. As this pid is presented in Unicode but is decoded via %d the user error message becomes a programmatic error message:

    File "/opt/rockstor/src/rockstor/storageadmin/views/pool_scrub.py", line 88, in post
    'and start a new scrub, use force option' % pid)
    TypeError: %d format: a number is required, not unicode

We can for the time being (prior to improving our general formatting of such things via .format) address this by casting to string rather than decimal, even though the pool id (pid) is a decimal. This at least will fix the observed formatting exception.

Please update the following forum thread with this issue resolution:

phillxnet:
As the pool id (pid) is presented in unicode we can't cast as %d in message string so use %s (string) instead.

Fixes #1776

Ready for review.

Please see issue text for user report details.

### DIFF
--- a/src/rockstor/storageadmin/views/pool_scrub.py
+++ b/src/rockstor/storageadmin/views/pool_scrub.py
@@ -84,7 +84,7 @@ class PoolScrubView(rfc.GenericView):
                     p.save()
                 else:
                     e_msg = ('A Scrub process is already running for '
-                             'pool(%d). If you really want to kill it '
+                             'pool(%s). If you really want to kill it '
                              'and start a new scrub, use force option' % pid)
                     handle_exception(Exception(e_msg), request)
 
--- a/src/rockstor/storageadmin/views/pool_scrub.py
+++ b/src/rockstor/storageadmin/views/pool_scrub.py
@@ -36,7 +36,7 @@ class PoolScrubView(rfc.GenericView):
         try:
             return Pool.objects.get(id=pid)
         except:
-            e_msg = ('Pool: %s does not exist' % pid)
+            e_msg = ('Pool with database id: ({}) does not exist.'.format(pid))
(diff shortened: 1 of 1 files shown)


phillxnet:
Test of fix:
%d formatting pre-pr:
d_formatting
%s formatting post-pr:
post-string-format-change

phillxnet:
@schakrava are you planning a 'more proper' fix for the issue that this pr addresses. I.e to return to displaying the pool name for instance. Otherwise it does, at least temporarily, address an know issue with a proven fix.

schakrava:
Hi @phillxnet have you tried using string.format()?
Something like this is better IMO:

`e_msg = ('A Scrub process is already running for '
                  'pool({}). If you really want to kill it '
                  'and start a new scrub, use force option'.format(pid))
`

phillxnet:
@schakrava "Something like this is better IMO: ..." I agree and have added a follow-up commit to address this. I have also returned full prior function by returning to pool name in the relevant error message, rather than id, and improved / normalise the other messages in this view.

All altered error messages tested with the following results:

Explain pid in validate:
pool-db-id-nonexistent

Command validation e_msg:
fake-command-message

Scrub returned to pool.name and minor surface formatting improvements:
scrub-in-progress-back-to-name

Ready for re-review.

schakrava:
Thank you @phillxnet

phillxnet:
@schakrava Cheers.

PR #1791, opened 2017-08-07
improve scrub status reporting resolution. Fixes #1786

### LINKED ISSUE #1640: SCRUB fails to end / fails to restart after shutdown
My box experienced an unexpected shutdown during the execution of a scrub, and I am now unable to schedule further scrubs or cancel the currently "running" one through the webui.
The box is running 3.8.16-8, but I believe the problem originated when it was running an older version. The problem has persisted across restarts and upgrades.

Link to origional post on rockstor forums [here](https://forum.rockstor.com/t/scrub-fails-to-end-fails-to-restart-after-shutdown/2769)

```bash
```

### LINKED ISSUE #1786: improve scrub status reporting resolution
Thanks to arbrick, Christian_Rost and Dragon2611 in the following forum threads for highlighting this issue. On closer inspection it appears that the existing scrub_status() mechanism can inadvertently report as running a prior interrupted scrub. It is thought that this may, in part, be responsible for the referenced reports.

- [x] fix portability bug in fs unit tests. Fixes #1787
- [x] Add unit tests to account for current expected behaviour of scrub_status().
- [x] Add unit test to indicate current suspected false positive for 'running' status on 'interrupted' scrubs.
- [x] improve scrub status reporting resolution to include 'interrupted'.

Forum threads respectively:
https://forum.rockstor.com/t/scrub-fails-to-end-fails-to-restart-after-shutdown/2769
https://forum.rockstor.com/t/scheduled-scrub-not-running/3608

### LINKED ISSUE #1787: fix portability bug in fs unit tests
An existing fs unit test relied on a locally accessible device name. This was due to a lack of adequate mocking. This issue was introduced as part of pr #1737 and discovered (and fixed) during development against issue #1786.

As the fix is trivial it will be rolled into the developing patch set against issue #1786.

- [x] make test_device_scan_parameter portable

phillxnet:
Prior to this pr scrub_status() reported only 'unknow', 'running', and 'finished'. The scrub halted (interrupted) state was misreported as running which lead to buggy behaviour. This pr improves scrub status reporting resolution to address the consequent buggy behaviour and improve user experience re scrub status.

- Fix portability bug in fs unit tests. Fixes #1787
- Add unit tests to account for current expected behaviour of scrub_status().
- Add unit test to indicate current suspected false positive for 'running' status on 'interrupted' scrubs.
- improve scrub status reporting resolution to include halted (interrupted), conn-reset (Connection reset by peer), and cancelled (aborted).
- Pool scrub job monitoring was also extended to include 'halted' and 'cancelled' as 'non running' jobs.

Fixes #1786

And as #1786 was essentially a duplicate of #1640 (with more specificity) we also have:
Fixes #1640

Note that this pr also includes a trivial fix for:
issue #1787 "fix portability bug in fs unit tests" so adding the also fixes:
Fixes #1787 

```
./bin/test --settings=test-settings -v 3 -p test_btrfs*
```
results in:
```
...
test_scrub_status_cancelled (fs.tests.test_btrfs.BTRFSTests) ... ok
test_scrub_status_conn_reset (fs.tests.test_btrfs.BTRFSTests) ... ok
test_scrub_status_finished (fs.tests.test_btrfs.BTRFSTests) ... ok
test_scrub_status_halted (fs.tests.test_btrfs.BTRFSTests) ... ok
test_scrub_status_running (fs.tests.test_btrfs.BTRFSTests) ... ok
...
```
All scrub states were reproduced on real hardware and their UI reflection confirmed. Note that to reproduce the halted ('interrupted') state the test machine was shutdown 'mid scrub'. Upon next boot the halted ('interrupted') state was in effect. And to reproduce 'conn_reset' / time out warning state a second scrub with the -f option was enacted over and existing running one. This caused interactivity issues and upon the first scrub finishing the conn_reset status was observed on an active (and waiting) btrfs scrub status command.

A trivial amount of code duplication was created but was highlighted with a "TODO:" and is very local in nature: within the same if-else block.

Ready for review.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -934,24 +934,57 @@ def scrub_start(pool, force=False):
 
 
 def scrub_status(pool):
+    """
+    Returns the raw statistics per-device (-R option) of the ongoing or last
+    known btrfs scrub. Works by parsing the output of the following command:
+    btrfs scrub status -R <mount-point>
+    :param pool: pool object
+    :return: dictionary indexed via 'status' and if a finished or halted, or
+    cancelld scrub is indicated then the duration of that scrub is added as
+    value to added index 'duration'. In all 'status' cases bar 'unknown',
+    data_bytes_scrubbed is passed as value to index 'kb_scrubbed' and all
+    other -R invoked details are returned as key value pairs.
+    """
     stats = {'status': 'unknown', }
     mnt_pt = mount_root(pool)
     out, err, rc = run_command([BTRFS, 'scrub', 'status', '-R', mnt_pt])
-    if (len(out) > 1):
-        if (re.search('running', out[1]) is not None):
+    if err != [''] and len(err) > 0:
+        if err[0] == "WARNING: failed to read status: Connection reset by " \
+                     "peer":
+            stats['status'] = 'conn-reset'
+            return stats
+    if len(out) > 1:
+        if re.search('interrupted', out[1]) is not None:
+            stats['status'] = 'halted'
+            # extract the duration from towards the end of the first line eg:
+            # "... 2017, interrupted after 00:00:09, not running"
+            dfields = out[1].split()[-3].strip(',').split(':')
+            stats['duration'] = ((int(dfields[0]) * 60 * 60) +
+                                 (int(dfields[1]) * 60) + int(dfields[2]))
+        elif re.search('running', out[1]) is not None:
             stats['status'] = 'running'
-        elif (re.search('finished', out[1]) is not None):
+        elif re.search('finished', out[1]) is not None:
             stats['status'] = 'finished'
+            # extract the duration from the end of the first line eg:
+            # "... 2017 and finished after 00:00:16"
+            dfields = out[1].split()[-1].split(':')
+            stats['duration'] = ((int(dfields[0]) * 60 * 60) +
+                                 (int(dfields[1]) * 60) + int(dfields[2]))
+        elif re.search('aborted', out[1]) is not None:
+            stats['status'] = 'cancelled'
+            # extract the duration from the end of the first line eg:
+            # "... 2017 and was aborted after 00:04:56"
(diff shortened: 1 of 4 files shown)


src/rockstor/fs/btrfs.py
```diff
@@ -934,24 +934,57 @@ def scrub_start(pool, force=False):
...
+    value to added index 'duration'. In all 'status' cases bar 'unknown',
+    data_bytes_scrubbed is passed as value to index 'kb_scrubbed' and all
+    other -R invoked details are returned as key value pairs.
+    """
     stats = {'status': 'unknown', }
     mnt_pt = mount_root(pool)
     out, err, rc = run_command([BTRFS, 'scrub', 'status', '-R', mnt_pt])
-    if (len(out) > 1):
-        if (re.search('running', out[1]) is not None):
+    if err != [''] and len(err) > 0:
```
schakrava: To me this line reads as `if len(err) != 0 and len(err) > 0`
can't this just be `if len(err) > 0:`?

schakrava:
Thank you @phillxnet , this is nice. I added a small comment, which is minor and you may address in a future pull request.

src/rockstor/fs/btrfs.py
phillxnet: Thanks, I'll revisit this soon as I have a few tidy/checkup's/todo's pending now and have in mind a suitable issue against which I can do them. But I think it's right as your analogy would hold if it were err != [] as len([]) = 0, however len(['']) = 1 and I think I was after:

if not an empty string in list (specifically) and at least a single element (to ensure a safe 0 index).

But yes given we then check the exact contents of err[0] directly there after the first if clause is redundant. However given it's the most common return (err=[''] is the norm) it's a rapid and cheap first check to avoid all others: I had imagined we would have occasion to add additional inner ifs than the one I added there.

I can at least clarify with comments on the why.

Cheers.

phillxnet:
@schakrava Thanks for the review and Cheers.

PR #1792, opened 2017-08-07
regression in 'force' option - scrub and balance. Fixes #1790

### LINKED ISSUE #1790: regression in 'force' option - scrub and balance
Thanks to forum members arbrick, Christian_Rost and Dragon2611 for helping to bring this bug to light.
The intention with the force option for scrubs and balances is two fold:

1. change the status of the last listed job (if currently 'started' or 'running') to 'terminated'.
2. pass the -f option to the respective btrfs commands and monitor this under a new job.

It is surmised that a recent Django or related update altered the requirement by which the existing 'force' option was passed from the Web-UI front end to the backend balance/scrub command wrappers. The result was that the force option was ineffectual in both 1. and 2. above. However in the case of a scrub there existed an additional bug that prevented 2. above. This is also fixed in the pr to be associated with this issue. Tasks addressed by the pending pr:

- [x] Fix 'force' UI option transition to backend - scrub/balance
- [x] Fix -f addition to scrub command with force option.

Please update the following forum threads with the resolution of this issue:
https://forum.rockstor.com/t/scrub-fails-to-end-fails-to-restart-after-shutdown/2769
https://forum.rockstor.com/t/scheduled-scrub-not-running/3608

phillxnet:
The intention with the force option for scrubs and balances is two fold:

1. change the status of the last listed job (if currently 'started' or 'running') to 'terminated'.
2. pass the -f option to the respective btrfs commands and monitor this under a new job.

It is surmised that a recent Django or related update altered the requirement by which the existing 'force' option was passed from the Web-UI front end to the backend balance/scrub command wrappers. The result was that the force option was ineffectual in both 1. and 2. above. However in the case of a scrub there existed an additional bug that prevented 2. as well. This is also addressed in this pr.

- Fix 'force' UI option transition to backend - scrub/balance
- Fix -f addition to scrub command with force option.

Logging was used to identify the failure of the force option to successfully propagate to the back-end as expected and the scrub and balance commands were confirmed to have received their '-f' options post pr.

Fixes #1790 

Ready for review.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -928,7 +928,7 @@ def usage_bound(disk_sizes, num_devices, raid_level):
 
 def scrub_start(pool, force=False):
     mnt_pt = mount_root(pool)
-    p = PoolScrub(mnt_pt)
+    p = PoolScrub(mnt_pt, force)
     p.start()
     return p.pid
 
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/pool_rebalance_table.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/pool_rebalance_table.js
@@ -83,6 +83,7 @@ PoolRebalanceTableModule = RockstorModuleView.extend({
                 $.ajax({
                     url: '/api/pools/' + _this.pool.get('id') + '/balance',
                     type: 'POST',
+                    contentType: 'application/json',
                     data: postdata,
                     success: function() {
                         _this.$('#pool-rebalance-form :input').tooltip('hide');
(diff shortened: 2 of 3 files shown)


src/rockstor/fs/btrfs.py
```diff
@@ -928,7 +928,7 @@ def usage_bound(disk_sizes, num_devices, raid_level):
 
 def scrub_start(pool, force=False):
     mnt_pt = mount_root(pool)
-    p = PoolScrub(mnt_pt)
+    p = PoolScrub(mnt_pt, force)
```
schakrava: really? good catch!

schakrava:
Special appreciation for this pr @phillxnet , those bugs are most likely mea culpa

phillxnet:
@schakrava Thanks and Cheers.

PR #1793, opened 2017-08-08
improve rebooted dialog message. Fixes #1584

### LINKED ISSUE #1584: improve rebooted dialog message
Occasionally when activating a reboot via the Web-UI we are for a while greeted with a message dialog which persists and has the message:
"Rockstor is now up and running"
I propose that we change this to:
"Rockstor is now up and running. Please refresh your browser window"
Small point as we eventually time out to first a self contradictory message and then a fresh login, but this change may help new users re-enter the Web-UI more smoothly on the occasions where this corner case is seen.

Some times the timing is such that a fresh login is offered directly which is of course fine.

phillxnet:
Simple pr to change the rebooted message that under certain circumstances reads:
"Rockstor is now up and running."
to
"Rockstor is now up and running. Please refresh your browser window"
This message is rarely displayed as it depends on quite specific system and browser timings and will often only appear transiently; but just for the instances when it does 'stick' this change makes it more user friendly.

Fixes #1584 

Ready for review.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst
@@ -117,7 +117,7 @@
 	  <div id="reboot-time-left"></div>
 	</div>
 	<div id="reboot-user-msg" style="display: none"><h4> System reboot still not complete. Connect to the console for current status of the system.</h4> </div>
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet that's an important reminder :)

phillxnet:
@schakrava Cheers.

PR #1794, opened 2017-08-09
Replace Chosen plugin with Select2 plugin. Fixes issue#1270

priyaganti:


schakrava:
Thank you @priyaganti nicely done!

PR #1801, opened 2017-08-21
scheduled scrub blocked by halted, conn-reset, and cancelled states. Fixes #1800

### LINKED ISSUE #1800: scheduled scrub blocked by halted, conn-reset, and cancelled states
In pr #1791 "improve scrub status reporting resolution." the additional scrub end states of 'halted' (interrupted), 'conn-reset' (connection reset by peer), and 'cancelled' (aborted) were introduced. Although the indicated pr addressed the mis-reporting of 'halted' (interrupted) as a running state it did not extend knowledge of the referenced additional states to the scheduled scrub manager. So although subsequent manual scrubs could be performed 'on top' of a now correctly identified 'halted' scrub a scheduled scrub was still not able to recognise this new scrub state as terminal: only 'error' and 'finished' were recognised as terminal (non running).

So from a scheduled scrub perspective any scrub performed manually (via Web-UI) or via a prior schedule (untested) that was then halted, cancelled, or resulted in a connection reset would block all future scheduled jobs on that same pool as although manual scrub awareness was extended to encompass the 'halted', 'conn-reset' and 'cancelled' states, the scheduled scrub mechanism was not.

The work around for this state is to enact a Web-UI initiated manual scrub that ends in state finished. There after the existing scheduled scrubs on the same pool would no longer be blocked as they are able to recognise 'finished' as a terminal scan state and initiate a new scrub task as a result. The force option is not required to enact this 'reset'.

Issue Tasks:

- [x] extend scheduled scrub manager's knowledge of the recently added terminal states.
- [x] test existing scheduled scrub tasks ability to recover / re-start post issue fix.

This issue was found while investigating the cause of #1798 but is not believed to be a mutual fix as it does not explain the list type specified in that issue.

phillxnet:
Please see issue text for context. Essential this pr extends the known terminal states of a scrub, from the scrub scheduler point of view, from the previous states of 'error' and 'finished' to include the additional non running states of 'halted', 'cancelled', and 'conn-reset'. Pre pr and pre scrub status enhancement pr #1791 a prior scheduled or manually initiated (via Web-UI) scrub that was in state 'halted' (interrupted eg by a reboot) would cause all future scheduled scrubs of the associated pool to not execute. This was as a result of the 'running' (pre pr #1791) or (post pr #1791) 'halted' state as both were considered as non-terminal, ie as not 'finished' or not 'errror' and hence all future scrub tasks would be blocked. By extending the terminal status to include the now correctly identified but equally non-running additional states existing scheduled scrub jobs are initiated successfully.

Tested by placing an ongoing Web-UI initiated scrub into both 'cancelled' (via 'btrfs scrub cancel') and 'halted' / interrupted (by rebooting mid scrub) and then ensuring that pre pr the existing scheduled scrub would fail:

    DEBUG [scripts.scheduled_tasks.pool_scrub:61] Non terminal state(cancelled) for task(2). Checking again.
    DEBUG [scripts.scheduled_tasks.pool_scrub:66] Non terminal state(cancelled) for task(2). A new task will not be run.

and then post pr that the same existing schedule scrub was initiated successfully.

Fixes #1800 

Ready for review.

### DIFF
--- a/src/rockstor/scripts/scheduled_tasks/pool_scrub.py
+++ b/src/rockstor/scripts/scheduled_tasks/pool_scrub.py
@@ -27,6 +27,10 @@ from django.utils.timezone import utc
 import logging
 logger = logging.getLogger(__name__)
 
+# List of scrub states considered terminal (non running):
+TERMINAL_SCRUB_STATES = ['error', 'finished', 'halted', 'cancelled',
+                         'conn-reset']
+
 
 def update_state(t, pool, aw):
     url = ('pools/%s/scrub/status' % pool)
@@ -56,11 +60,11 @@ def main():
 
         if (Task.objects.filter(task_def=tdo).exists()):
             ll = Task.objects.filter(task_def=tdo).order_by('-id')[0]
-            if (ll.state != 'error' and ll.state != 'finished'):
+            if ll.state not in TERMINAL_SCRUB_STATES:
                 logger.debug('Non terminal state(%s) for task(%d). Checking '
                              'again.' % (ll.state, tid))
                 cur_state = update_state(ll, meta['pool'], aw)
-                if (cur_state != 'error' and cur_state != 'finished'):
+                if cur_state not in TERMINAL_SCRUB_STATES:
                     return logger.debug('Non terminal state(%s) for task(%d). '
                                         'A new task will not be run.' %
                                         (cur_state, tid))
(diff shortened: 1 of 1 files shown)


schakrava:
Very nice! Thank you @phillxnet

phillxnet:
@schakrava Cheers.

PR #1803, opened 2017-08-22
scrub task end time incorrectly updated. Fixes #1802

### LINKED ISSUE #1802: scrub task end time incorrect
Thanks to forum member peter in the following forum post for highlighting this issue. The end time of pool scrub jobs is almost always around 2 seconds after the start time; indicating a bug in the update mechanism of this UI element. Note that the associated pool scrub end time, as presented on the relevant pool details page (scrub tab), is updated correctly.

A resolution to this bug has been found and a pull request is being prepared.

https://forum.rockstor.com/t/scheduled-scrub-not-running/3608/18

phillxnet:
Scrub task end times were incorrectly updated directly after scrub task initialization. They were not updated
there after. This resulted in premature end times of start time + 2 seconds.

The pull request moves the end time update and adds a db save to directly after current scrub state is assessed as terminal. This also results in an empty end time during scrub execution which is in keeping with the pool details scrub tab table.

Note that his patch is against master but was tested post pr #1801 which alters a line that lies between the moved line in this patch (within 5 lines each way). Please advise if this patch should be re-based on master once pr #1801 is merged.

Fixes #1802 

Ready for review.

### DIFF
--- a/src/rockstor/scripts/scheduled_tasks/pool_scrub.py
+++ b/src/rockstor/scripts/scheduled_tasks/pool_scrub.py
@@ -77,7 +77,6 @@ def main():
             t.state = 'error'
             logger.exception(e)
         finally:
-            t.end = datetime.utcnow().replace(tzinfo=utc)
             t.save()
 
         while True:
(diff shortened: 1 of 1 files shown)


schakrava:
Thank you @phillxnet . No rebasing necessary.

phillxnet:
@schakrava cool, thanks.

PR #1805, opened 2017-08-29
improve share mount re ro,degraded pool options. Fixes #1804

### LINKED ISSUE #1804: improve share mount re ro,degraded pool options
On boot ro,degraded pool mount options, in the presence of a degraded pool, can cause all of it's own sub-vols to fail to mount and, depending on db output order, all other non system pool sub-vols to also not be mounted and block UI reporting of system drive sub-vol mount status. This is caused by the current handling of a couple of quota related command wrappers that fail to deal elegantly with the command exception that results from attempting to first create and then update the respective shares' Share.pool relevant qgroup limit.

Note that another behavioural artefact of this bug is that when booting with default mount options when a pool is degraded (ie device missing), and the typical ro,degraded options are applied; the degraded pools sub-vols (shares) are not, in turn, mounted upon Pool or Share page refresh. This complicates the data retrieval that is associated with such a scenario and voids the UI's utility in this context.

phillxnet:
When mounting a ro,degraded pool for example, not catching the resulting exception on qgroup create/limit can break subvol(share) mounts. Except for the specific case of a Read-only file system this patch does not alter existing behaviour.

The manual test case for this pr was to establish a 2 dev test pool with multiple subvols (shares) on a system containing existing pools with their own subvols (shares). The test pool was then degraded by removing one of it's drives (while the system was down) and booting the system with this degraded pool.

- Pre pr behaviour:
On bootup the missing dev pool was indicated as unmounted and it's prior subvols were not listed. Adding ro,degraded as typical custom options in this scenario resulted in a pool(vol) remount as expected but no prior subvols were listed/mounted. Also note the occasional side effect (detailed in the issue text #1804) of all non system drive pools' subvols failing to mount on subsequent reboots; was also observed multiple times.

- Post pr behaviour:
As for "Pre pr behaviour" except that all prior test pool subvols were successfully mounted on all occasions and no other pools or subvols were observed to be affected in subsequent reboots.

Directly after subvols are mounted, both after ro,degraded options are applied to a degraded pool, and after a reboot there after, smb shares function as expected. This enables data retrieval via regular means, where possible, prior to addressing the cause of the ro,degraded mount requirement.

Fixes #1804 

Ready for review.

Note on curios mount option differences re subvol mounts directly after custom pool mount options were applied (and consequent page refresh) and with those same mount options in play on subsequent reboots:

_When booting with missing pool device and no custom options and then adding ro,degraded to the pool as custom options, the subvols are mounted as rw but upon attempting to create further shares (subvols) within the ro,degraded pool the expected 'ERROR: cannot create subvolume: Read-only file system' was presented. This appears to be default btrfs behaviour as Rockstor has no share level mount options but this behaviour could be addressed, if necessary, when share (subvol) level mount options are added, ie we could prioritise a pool ro flag if present to it's subvols. It is interesting to note that post a reboot with these same custom mount options in place (ro,degraded) all test pool subvols are then mounted ro,degraded._

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -687,7 +687,19 @@ def qgroup_create(pool):
     # mount pool
     mnt_pt = mount_root(pool)
     qid = ('%s/%d' % (QID, qgroup_max(mnt_pt) + 1))
-    o, e, rc = run_command([BTRFS, 'qgroup', 'create', qid, mnt_pt], log=True)
+    try:
+        out, err, rc = run_command([BTRFS, 'qgroup', 'create', qid, mnt_pt],
+                                   log=True)
+    except CommandException as e:
+        # ro mount options will result in o= [''], rc = 1 and e[0] =
+        emsg = 'ERROR: unable to create quota group: Read-only file system'
+        # this is non fatal so we catch this specific error and info log it.
+        if e.rc == 1 and e.err[0] == emsg:
+            logger.info('Pool: {} is Read-only, skipping qgroup '
+                        'create.'.format(pool.name))
+            return qid
+        # raise an exception as usual otherwise
+        raise
     return qid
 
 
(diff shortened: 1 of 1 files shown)


schakrava:
This is really nice @phillxnet, much needed fix. Thanks a bunch!

phillxnet:
@schakrava Cheers.

Issue #1759, opened 2017-07-13
Scheduled scrub throws error


kbogert:
Per: https://forum.rockstor.com/t/scheduled-scrub-throws-an-error/3448

The scheduled scrub tasks are failing due to use of the pool's name field instead of the ID.

phillxnet:
Had a quick look into this earlier and my intention is to have another look later today. All progress will be noted in this issue.

phillxnet:
Each scheduled task contains a json_meta field which is used by:
src/rockstor/scripts/scheduled_tasks/pool_scrub.py
to construct the API url.
Example contents of this json_meta field:
```
json_meta={u'pool': u'rockstor_rockstor'}
```
or via eg: https://rockdev.local/api/sm/tasks/ excerpt:
```
json_meta": "{\"pool\": \"rockstor_rockstor\"}
```
So we need to populate this json_meta field with pool id's rather than names.

phillxnet:
@schakrava I have identified and linked to a duplicate report of this issue (#1762) and commented there on it's duplicate nature: indicating the 'awaiting review' status of the associated pull request (#1760).

PR #1760, opened 2017-07-14
fix scheduled scrub regression re pool api changes. Fixes #1759

phillxnet:
Thanks to @kbogert, @MFlyer, and KarstenV on the forum for helping to diagnose the associated issue.
Update UI input to pass pool id not name as per new api. Also adds pool_name to meta field and backbone model of scrub tasks to enable scheduled task interfaces to report the associated pool name.

Some minor cosmetic spacing adjustments were also included in UI pool name display elements.

Fixes #1759 

@schakrava Ready for review

Please see the associated issue text and the following forum thread for historical context:
https://forum.rockstor.com/t/scheduled-scrub-throws-an-error/3448

PR #1808, opened 2017-09-08
change share and snapshot endpoints to by id and not by name. Fixes #…

schakrava:
…1807

ids don't change during the lifecycle of the resource and let's us rename
shares easily.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/models/models.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/models/models.js
@@ -130,7 +130,7 @@ var PoolCollection = RockStorPaginatedCollection.extend({
 
 var Share = Backbone.Model.extend({
     url: function() {
-        return '/api/shares/' + this.get('shareName');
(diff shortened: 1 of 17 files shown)


phillxnet:
@schakrava On a quick look at these changes I think we will have the same issue as we had earlier with the API change for pools (re:scrubs) where smart_manager_taskdefinition contains (within field json_meta) now defunct references ie "share": "luks-share" for snapshot tasks. This will presumably self correct for all new tasks (as with pool scrub tasks) but we are then left with no snapshot task json_meta name reference and will have to resolve to a name (might be tricky in places). Otherwise task tables will display the id rather than the name.

Requesting that this be debugged and fixed prior to the next stable release.

For pools I stamped the current name within the json_meta (not ideal but allowed for a work around with existing task mechanisms with only cosmetic artefacts there after) ie see:
pr #1760 "fix scheduled scrub regression re pool api changes. Fixes #1759"

PR #1812, opened 2017-09-12
fix snapshot scheduled task regression re share api change #1809

### LINKED ISSUE #1809: snapshot scheduled task regression re 3.9.1-9 API change
As of Rockstor version 3.9.1-9 there is a regression in scheduled snapshot tasks. This is due to a change in share API from name to id which needs to be accompanied by a pending change in task definition. Previously we used the the 'share' field within json_meta in smart_manager_taskdefinition to store ie: "share": "luks-share" as the name of the share to be snapshot-ed however given the API change we can no longer use this same element to effect API creation.

A parallel scheduled pool scrub regression was indicated in issue:
#1759 "Scheduled scrub throws error"
and consequently fixed in pr:
#1760 "fix scheduled scrub regression re pool api changes. Fixes #1759"

The fix in that instance was to move existing use of json_meta .pool to reference id and add json_meta.pool_name (during task creation) so that scrub tasks held pool name info for display purposes. This is not an ideal fix in light of the pending pool / share re-name capability but did serve to perpetuate pool name display in task related tables.

It is suggested that we enable a similar quick, if less than ideal, fix in this scenario and look to enabling a fix that honours the pending share name change feature when that feature is introduced.

Current consequence is that all existing and freshly created snapshot scheduled tasks will fail to execute with the indication of such being an "error" task status upon scheduled task execution time.

Manual snapshot creation is not affected.

phillxnet:
Update UI input to pass share id not name as per new api. Also adds share_name to meta field and backbone model of snapshot tasks to enable scheduled task interfaces to report the share name.

Update cron activated snapshot script to use the new share id based api, including the removal of > max snapshot removal elements.

Includes some minor tab char removal.

Fixes #1809 

@schakrava  Ready for review

Tested by establishing a share snapshot task to run every 5 mins with a 4 snapshot limit. All tasks ran and a max of 4 snapshots were observed on the share's details page. Email notifications indicated no errors with the associated snapshot cron jobs.

### DIFF
--- a/src/rockstor/scripts/scheduled_tasks/snapshot.py
+++ b/src/rockstor/scripts/scheduled_tasks/snapshot.py
@@ -68,7 +68,7 @@ def delete(aw, share, snap_type, prefix, max_count):
         name__startswith=prefix).order_by('-id')
     for snap in snapshots[max_count:]:
         try:
-            url = ('shares/%s/snapshots/%s' % (share.id, snap.name))
+            url = ('shares/{}/snapshots/{}'.format(share.id, snap.name))
             aw.api_call(url, data=None, calltype='delete', save_error=False)
         except Exception as e:
             logger.error('Failed to delete old snapshots exceeding the '
@@ -105,7 +105,7 @@ def main():
             name = ('%s_%s'
                     % (meta['prefix'],
                        datetime.now().strftime(settings.SNAP_TS_FORMAT)))
-            url = ('shares/%s/snapshots/%s' % (share.id, name))
+            url = ('shares/{}/snapshots/{}'.format(share.id, name))
             # only create a new snap if there's no overflow situation. This
             # prevents runaway snapshot creation beyond max_count+1.
             if(delete(aw, share, stype, prefix, max_count)):
--- a/src/rockstor/scripts/scheduled_tasks/snapshot.py
+++ b/src/rockstor/scripts/scheduled_tasks/snapshot.py
@@ -34,10 +34,15 @@ def validate_snap_meta(meta):
         raise Exception('meta must be a dictionary, not %s' % type(meta))
     if ('prefix' not in meta):
         raise Exception('prefix missing from meta. %s' % meta)
+    if ('share_name' not in meta):
+        raise Exception('share_name missing from meta. {}'.format(meta))
     if ('share' not in meta):
         raise Exception('share missing from meta. %s' % meta)
-    if (not Share.objects.filter(name=meta['share']).exists()):
-        raise Exception('Non-existent Share(%s) in meta. %s' %
+    if not meta['share'].isdigit():
+        raise Exception('Non-digit share element ({}) in meta {}'
+                        .format(meta['share'], meta))
+    if (not Share.objects.filter(id=meta['share']).exists()):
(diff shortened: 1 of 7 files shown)


phillxnet:
Note that this pr is based on the prior merged:
#1760 "fix scheduled scrub regression re pool api changes. Fixes #1759"
which concerned a fix for a similar pool api change.

The approach taken in both this pr and the referenced  pool scrub api fix is not ideal but is consistent with existing code and should only result in cosmetic issues upon pool/share name change features being introduced at a later date, with an available work around of re-creating the scheduled task.

N.B. as with the pool api scheduled scrub task fix, it will be necessary for users to delete and re-create all their existing scheduled snapshot tasks (post this pr) in order that their prior function (pre share api change) be restored.

src/rockstor/scripts/scheduled_tasks/snapshot.py
```diff
@@ -63,7 +68,7 @@ def delete(aw, share, snap_type, prefix, max_count):
         name__startswith=prefix).order_by('-id')
     for snap in snapshots[max_count:]:
         try:
-            url = ('shares/%s/snapshots/%s' % (share.name, snap.name))
+            url = ('shares/%s/snapshots/%s' % (share.id, snap.name))
```
schakrava: since share.id is an integer,  it should be %d. But even better to just use `'shares/{}/snapshots/{}'.format(share.id, snap.name)` so we don't have to deal with types ourselves.

src/rockstor/scripts/scheduled_tasks/snapshot.py
```diff
@@ -100,7 +105,7 @@ def main():
             name = ('%s_%s'
                     % (meta['prefix'],
                        datetime.now().strftime(settings.SNAP_TS_FORMAT)))
-            url = ('shares/%s/snapshots/%s' % (share.name, name))
+            url = ('shares/%s/snapshots/%s' % (share.id, name))
```
schakrava: same as my previous comment.

schakrava:
Thanks for having my back @phillxnet ! Please address my review comments and I'll take another look and merge.

src/rockstor/scripts/scheduled_tasks/snapshot.py
phillxnet: Agreed, I had thought we would run into the same unicode issue as in #1776 with %d use but as you say in this context that is not the case. However as advised I have move to str.format for the url string formats at least; much nicer.

src/rockstor/scripts/scheduled_tasks/snapshot.py
phillxnet: Now sorted as per my other comment. Cheers.

phillxnet:
@schakrava Thanks for the review. I have now moved the indicated lines to str.format as advised and re-tested successfully.

Haven't touched the other string formats for the time being as intending to do a project wide issue dedicated to moving to str.format and can then catch the remaining old style string formats we have remaining in this file.

Ready for re-review.

schakrava:
thank you sir!

phillxnet:
Cheers.

PR #1820, opened 2017-09-17
Add travis file for CI

gkadillak:
Adds continuous integration builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ python:
 
 # command to install dependencies
 install:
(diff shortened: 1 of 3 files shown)


barnhill:
very good idea for an open source project

phillxnet:
Closing in preference for a GitHub actions equivalent. We do also now have a core team Buildbot arrangement but an actions equivalent would be nice for in-GitHub results made easy.

Also closing as part of a general pr tidy as we approach our first v4 stable.

PR #1835, opened 2017-09-30
share link regression on snapshots page. Fixes #1829

### LINKED ISSUE #1829: share link regression on snapshots page
The share details links within the Share column of the main snapshots page have not been updated to use the new share API. This results in the following error when they are accessed:

    Unknown client error doing a GET to /api/shares/luks-share/snapshots?page=1&format=json&page_size=5000&count=&snap_type= 

Rather than the intended behaviour of linking to the Share's detail page.

phillxnet:
Move share link in main snapshots page over to using the newly changed shares API. This is done by explicitly adding a share_name element to the existing UI snapshots array. This then 'uncovers' the existing share id value allowing for additional future enhancements / code simplifications. Todo's were added to this latter end.

Tested by applying included changes and ensuring share links within the snapshots table 'Shares' column then worked. Cloning and deleting snapshots via the first column icons was also tested to still work. The Create button function was also tested: though should not have been affected.

Fixes #1829 

Ready for review.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/share/snapshots.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/share/snapshots.jst
@@ -56,17 +56,17 @@
               </td>
               <td><i class="glyphicon glyphicon-camera"></i>&nbsp;{{this.name}}
                {{#if this.writable}}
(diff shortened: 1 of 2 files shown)


schakrava:
Thank you @phillxnet

phillxnet:
@schakrava Cheers.

PR #1838, opened 2017-10-07
failure in share import prep re duplicate name and usage. Fixes 1828

### LINKED ISSUE #1828: failure in share import prep re duplicate name and usage
Thanks to forum member @magicalyak for reporting and helping to diagnose this issue.  Under certain circumstances when a share on disk is not currently reflected in the db an attempt is made to import and mount each share.  This effort includes a check to establish if there are duplicate share names which is unfortunately circumvented due to a bug which causes this check to never be executed, along with various share calculations that would otherwise also be executed.

Please see the following forum thread for reference of this issues origin:
https://forum.rockstor.com/t/subvolumes-not-mounted-but-pools-mount/3797

phillxnet:
When attempting to import a pool that has a subvol by the same name as an existing subvol on another pool known to the system the intended warning explaining the lack of support for this arrangement (ie unique subvol names across an install) failed to log as expected. This was down to a prior exception in the same try clause preventing the intended log action: a call to share_info() using it's prior parameter type.

A caveat of the related fix (share_info() parameter fix) was that an intended call to volume_usage(), introduced during the existence of the share_info() parameter bug, was also blocked from ever running. And this code, in turn, had a syntactic anomaly.

1. Update misleading docstring on share_info().
2. Refactor import_shares() via var name changes to aid maintenance / readability.
3. Update the call to share_info() to use the new parameter type.
4. Improve 'duplicate name' during import log message wording and update format method of same.
5. Correct syntactic anomaly to the now live code concerning the intended volume_usage() call.
6. Tighten db retrieval filter to more exactly represent our required search: ie specify the desired pool rather than the existing system wide subvol search.
7. Confine explicit mount to the intended setup - avoiding a silent variable scope issue that would prevent this mount's execution anyway.
8. Add 'bare bones' debug level logging to aid in debugging future import issues.
9. Abstract db update pertaining to smart_manager_shareusage db table, aids readability / maintenance.
10. Use newly abstracted smart_manager_shareusage update mechanism to do the same on import for prior unknown, or moved (between pools) subvols. This was previously neglected.
11. Ensure that when we create a new share db entry we specify rusage, eusage, pqgroup_rusage, and pqgroup_eusage.

Please see pr issue #1828 text for a reproduction of the reported issue and a consequent proof of fix re duplicate subvol name element of this issue.

Fixes #1828 

Ready for review.

This pr does not change the essential mechanism used by share_import() but does, by way of bug fixes, extend the work done. Especially where the db in concerned: specifically smart_manager_shareusage and the additional calls to volume_usage. In the comments I will detail a basic test procedure used to establish existing function remained intact.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -406,8 +406,15 @@ def snapshot_list(mnt_pt):
 
 
 def shares_info(pool):
-    # return a list of share names under this mount_point.
-    # useful to gather names of all shares in a pool
+    """
+    Returns a list of share/subvol names by retrieving the mount point of the
+    passed pool and using this to run "btrfs subvol list -s mnt_point".
+    N.B. Child snapshots and subvolumes are ignored but writable snapshots that
+    are immediate children of a pool (vol) are not ignored and regarded as
+    shares in their own right.
+    :param pool: Pool object
+    :return: list of share/subvol names found under Pool.name
+    """
     try:
         mnt_pt = mount_root(pool)
     except CommandException as e:
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -407,13 +407,16 @@ def snapshot_list(mnt_pt):
 
 def shares_info(pool):
     """
-    Returns a list of share/subvol names by retrieving the mount point of the
-    passed pool and using this to run "btrfs subvol list -s mnt_point".
+    Returns a dictionary of share/subvol names via passed pool mount point
+    lookup and using this to run "btrfs subvol list -s mnt_point" for snapshots
+    and "btrfs subvol list -p mnt_point" for all subvols including parent id.
     N.B. Child snapshots and subvolumes are ignored but writable snapshots that
     are immediate children of a pool (vol) are not ignored and regarded as
-    shares in their own right.
+    shares in their own right (a Share 'clone' in Rockstor parlance).
     :param pool: Pool object
-    :return: list of share/subvol names found under Pool.name
+    :return: dictionary indexed by share/subvol names found directly under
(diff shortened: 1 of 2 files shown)


phillxnet:
The basic import mechanism was tested, post pr, via the following script:

```
#!/bin/bash

# assumes a blank virtio disk with serial "serial-6"
# thus /dev/disk/by-id/virtio-serial-6

# create our pool/volume ie import-test-pool (it-pool)
mkfs.btrfs -L it-pool /dev/disk/by-id/virtio-serial-6
mkdir -p /mnt2/it-pool
mount /dev/disk/by-id/virtio-serial-6 /mnt2/it-pool
mkdir -p /mnt2/it-pool/.snapshots


# proc to create multiple subvols (shares), each with multiple rw and ro snapshots.
for sb in $(seq 1 10)
do
    btrfs subvol create /mnt2/it-pool/it-subvol-"$sb"
    for snap in $(seq 1 10)
    do
      	mkdir -p /mnt2/it-pool/.snapshots/it-subvol-"$sb"
        btrfs subvol snapshot /mnt2/it-pool/it-subvol-"$sb" /mnt2/it-pool/.snapshots/it-subvol-"$sb"/it-subvol-"$sb"-snapshot-"$snap"
        btrfs subvol snapshot -r /mnt2/it-pool/it-subvol-"$sb" /mnt2/it-pool/.snapshots/it-subvol-"$sb"/it-subvol-"$sb"-snapshot-ro-"$snap"
    done
done
```

Essentially creating, on a single blank virtio drive with serial "serial-6", 10 subvols (shares) each with 10 ro and 10 rw snapshots structured as if they had been created by a prior Rockstor instance. Multiple import / disk wipe cycles were carried out with no relevant errors / logging observed.

schakrava:
Thank you @phillxnet

phillxnet:
@schakrava Cheers.

PR #1839, opened 2017-10-11
abstract by-id device path in a function

KaiRo-at:
As discussed in PR #1704 this is an updated version of the patch there, reworked on top of current master.
This patch abstracts away the actual /dev/disk/by-id/ path substitution into a function, grouped with similar function in osi.py.
This would help us at Minebox to stay closer to upstream.


schakrava:

schakrava:

phillxnet:
@KaiRo-at hello again.
@schakrava I can take a look at this, but not directly unfortunately. Hopefully soon though.

schakrava:
LGTM. Thanks @KaiRo-at

PR #1862, opened 2017-12-04
return realtime info in task defition via properties. Fixes #1857

### LINKED ISSUE #1857: suspected 3.9.2-1 scheduled snapshot UI regression
Thanks to forum members ScarabMonkey and Maddan in the following thread for highlighting this issue. It appears that we have a regression introduced by stable hotfix pr: #1855 which was intended to aid migration from scheduled task meta created from old share api. From Maddan's account this appears not to have worked as intended (in the field) and has also blocked / broken the prior confirmed, by ScarabMonkey, function of creating a fresh schedule. Ie no migration and broken freshly created scheduled snapshots. A quick fix for this issue (once confirmed / clarified by devs) would be to revert hot fix #1855 and return at least the function of a freshly created scheduled snapshot as then users have at least a work around (recreation of scheduled snapshot tasks), and fresh installs function as intended. Currently scheduled snapshots in 3.9.2-1 are suspected broken via migration or fresh creation (dev confirmation required).

@schakrava I suggest we revert #1855 in stable and publish release notes for the stable channel indicating the need to recreate scheduled tasks. Then as we approach the pool/share renaming capability we improve on the redundant storage of share name within meta for both scrub and snapshot scheduled tasks which we both agree is sub-optimal (my partial fix to at least return function to freshly created scheduled tasks). Alternatively a 'proper fix' is required that may not be timely enough, depending on resources, that includes the migration capability which is obviously preferred.

Please see the following forum thread for context:
https://forum.rockstor.com/t/issue-schedules-broken-after-update-to-3-9-2-0/3999

As of 3.9.2-0 (stable) freshly created scheduled tasks of snapshot or scrub have been confirmed as functioning and displaying as intended with the unfortunate but necessary caveat that we have, currently harmless, redundant meta date of share_name in the scheduled task info. I suggest that this be addressed in the next development release by properly retrieving share name from shareid for the UI components: a task I purposefully postponed and noted in the related fixes for broken scheduled scrubs and task; respective reviewed prs are:
"fix scheduled scrub regression re pool api changes. Fixes #1759" #1760
"fix snapshot scheduled task regression re share api change #1809" #1812
with the latter noting the following:

"N.B. as with the pool api scheduled scrub task fix, it will be necessary for users to delete and re-create all their existing scheduled snapshot tasks (post this pr) in order that their prior function (pre share api change) be restored."

Please note that Maddan's report has yet to be confirmed as it may be that we only have a UI regression here. Creating issue to address either way.

schakrava:
information like share_name and pool_name can be supplied when the model is
serialized during a GET, by constructing it in real time. This pattern is
already there in a few other models and works nicely for any information that
shouldn't be stored in db or at least not duplicated. The other way of solving
this problem is to have UI/client make explicit calls to appropriate end points
to get share_name etc.. which is more RESTful, but introduces unnecessary
complication in our case.

### DIFF
--- a/src/rockstor/smart_manager/models/task_def.py
+++ b/src/rockstor/smart_manager/models/task_def.py
@@ -16,7 +16,10 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
+import json
+
 from django.db import models
+from storageadmin.models import (Pool, Share)
 
 
 class TaskDefinition(models.Model):
@@ -39,3 +42,25 @@ class TaskDefinition(models.Model):
 
     class Meta:
         app_label = 'smart_manager'
+
+    @property
+    def share_name(self, *args, **kwargs):
+        sn = None
+        if self.task_type == 'snapshot':
+            task_metadata = json.loads(self.json_meta)
+            try:
+                sn = Share.objects.get(id=task_metadata['share']).name
+            except Share.DoesNotExist:
+                sn = 'N/A'
+        return sn
+
+    @property
+    def pool_name(self, *args, **kwargs):
+        pn = None
+        if self.task_type == 'scrub':
+            task_metadata = json.loads(self.json_meta)
+            try:
+                pn = Pool.objects.get(id=task_metadata['pool']).name
+            except Pool.DoesNotExist:
+                pn = 'N/A'
+        return pn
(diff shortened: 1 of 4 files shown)


phillxnet:
Thanks for the explanation. I rather dropped the ball on this one last time with the slightly rushed scheduled task regression fixes (scrub and snapshot).

PR #1864, opened 2017-12-04
share name blank in edit schedule snapshot page. Fixes #1863

### LINKED ISSUE #1857: suspected 3.9.2-1 scheduled snapshot UI regression

### LINKED ISSUE #1863: share name blank in edit schedule snapshot page
When editing an existing snapshot created with revised meta format (ver 3.9.2-1, ie no share_name element) we fail to retrieve the associated share name for this element via the recently added model extensions detailed in pr:

"return realtime info in task defition via properties. Fixes #1857" #1862 

The same modifications applied for js UI elements within the referenced pr may well apply to this remaining UI component.

phillxnet:
As we now have a share/pool name property components in the taskDef model we can remove the remaining UI dependencies on the previously redundant name storage within sched task meta info. Although currently this only affected scheduled tasks it would, once we also no longer store pool_name in sched scrub meta, also arise similarly for scheduled scrubs. Proof of this effect and the fix's function was enacted by hand editing scrub meta db info to simulate future removal of the pool_name element on a freshly created pool scrub scheduled task.

Also contains a fix for subtitle in task history showing 'undefined' where no share/pool_name element was found.

Fixes #1863 

Note that his pr depends on the changes applied in now merged pr #1862 :
"return realtime info in task defition via properties. Fixes #1857"

@schakrava Ready for review.


phillxnet:
Manual test data for scrub json_meta within smart_manager_taskdefinition was as follows:

"{"pool": "6"}" ie no pool_name element (my previous work around)

and both task history subtitle and existing scrub edit pages were able to successfully retrieve and display the pool name.

Scheduled snapshots that were created post 3.9.2-1 were affected as per issue text, where as post this pr the share name entries detailed in issue text functioned as expected.

phillxnet:
@schakrava Cheers.

PR #1874, opened 2017-12-18
improve quotas not enabled behaviour. Fixes #1869

### LINKED ISSUE #1869: improve quotas not enabled behaviour
Under certain as yet undetermined circumstances quotas can become disabled. One consequence of this was a failure to mount the related shares upon the following boot / reboot (addressed in issue #1867 and related pr #1868). However there are additional consequences for running with a pool / subvol that has quotas not enabled. These include:

- [x] Inability to create new shares in that pool.
- [x] Inability to resize existing shares.
- [x] Apparent Inability to create a snapshot (error on trying but it appears)
- [x] Apparent Inability to delete a snapshot (error on trying but it get deleted)
~~[ ] Size reporting of the affected subvols is reported as 0 bytes.~~
- [x] Deal with quotas being re-enabled once shares etc have been created with default pqgroup -1/-1
- [x] Test pool import for ability to recreate/reset pqgroups
- [x] Test bootstrap command on quota enabled pool after having been used with quotas disabled ie -1/-1 entries in db.
- [x] Ensure all pqgroup's are unique while being robust to quote enable disable re-enabled cycle.
- [x] Add user feedback on Pool Quotas Enabled / Disabled 
- [x] Share created when quotas disabled get fresh pqgroup during re-size re-submission if done when quotas are enabled (no change needed).
- [x] Share with non existent pqgroup (due to quota disable removing all pqgroups from pool) will also be re-assigned a fresh pqgroup once quotas are re-enabled and a resize is performed.
- [x] remove all non existent pqgroups and re-assign fresh ones during boot if quotas enabled, already done for prior -1/-1 pqgroup shares.

phillxnet:
Uses the existing -1/-1 default for pqgoups to represent an unset state for the share.pqgroup and adds an active return to this value whenever quotas are deemed to be disabled. Pqgroup setup is moved from the bootstrap process into the import_shares / shares refresh section. This allows for live setting / resetting of pqgroups via the regular share re-fresh process. Pool.quotas_enabled and share.pqgroup_exist properties are added and all used low level quota actions are adjusted to catch, log, and ignore a quota disabled state. Additionally active pqgroup assignment is added to share
resize and share refresh procedures which aid in returning to existence the expected native 2015/n pqgroups and their relationship to the auto generated 0/n qgroups as whenever quotas are disabled all this info (var the 0/n) is lost. UI elements are added in pool and share focused pages to indicate a live status for the associated pool quotas enabled status.

Fixes #1869 

@schakrava ready for review

Please see linked issue for images of the resulting UI components within this pr.

This pr aims to improve robustness to a quotas disabled state and, with some limitations, is able to re-establish full quota group relationships upon quotas becoming re-enabled. Although a share re-scan is required for this to take place.

Know limitations:
- When importing a pool via the disk page there is a failure to assert pqgroup ggroup relationships, this is intentional as adding the additional potential quota scans is currently an unknown load on what is already a fairly intense process. This short fall can be addressed separately in another issue/pr.
- when making a clone that clone's pqgroup is not immediately established, however there also exist a potentially related bug where clones are also not mounted: again this can be addressed separately and is in time resolved by future share scans anyway.
- Long term pqgroup to share identity is not maintained over quota disabled enabled cycles. This does not look to affect any current function, but given all pqgroups are deleted whenever quotas are disabled it would require additional complexity whenever reconstructing, ie split brain between db and pool: hence the defaulting to -1/-1 in unknown / quota disabled states for all shares associated with the pool in question.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -26,6 +26,7 @@ from system.osi import run_command, create_tmp_dir, is_share_mounted, \
 from system.exceptions import (CommandException)
 from pool_scrub import PoolScrub
 from django_ztask.decorators import task
+from django.conf import settings
 import logging
 
 """
@@ -41,7 +42,8 @@ UMOUNT = '/bin/umount'
 DEFAULT_MNT_DIR = '/mnt2/'
 RMDIR = '/bin/rmdir'
 QID = '2015'
-
+# The following model/db default setting is also used when quotas are disabled.
+PQGROUP_DEFAULT = settings.MODEL_DEFS['pqgroup']
 
 def add_pool(pool, disks):
     """
@@ -678,13 +680,71 @@ def disable_quota(pool_name):
     return switch_quota(pool_name, flag='disable')
 
 
+def are_quotas_enabled(mnt_pt):
+    """
+    Simple wrapper around 'btrfs qgroup show -f --raw mnt_pt' intended
+    as a fast determiner of True / False status of quotas enabled
+    :param mnt_pt: Mount point of btrfs filesystem
+    :return: True on rc = 0 False otherwise.
+    """
+    o, e, rc = run_command([BTRFS, 'qgroup', 'show', '-f', '--raw', mnt_pt])
+    if rc == 0:
+        return True
+    return False
+
+
+def qgroup_exists(mnt_pt, qgroup):
+    """
+    Simple wrapper around 'btrfs qgroup show --raw mnt_pt' intended to
+    establish if a specific qgroup exists on a btrfs filesystem.
+    :param mnt_pt: btrfs filesystem mount point, usually the pool.
+    :param qgroup: qgroup of the form 2015/n (intended for use with pqgroup)
+    :return: True is given qgroup exists in command output, False otherwise.
+    """
+    o, e, rc = run_command([BTRFS, 'qgroup', 'show', '--raw', mnt_pt])
+    # example output:
+    # 'qgroupid         rfer         excl '
+    # '-------         ----         ---- '
+    # '0/5             16384        16384 '
+    # ...
+    # '2015/12             0            0 '
+    if rc == 0 and len(o) > 2:
+        # index from 2 to miss header lines and -1 to skip end blank line = []
+        qgroup_list = [line.split()[0] for line in o[2:-1]]
+        # eg from rockstor_rockstor pool we get:
+        # qgroup_list=['0/5', '0/257', '0/258', '0/260', '2015/1', '2015/2']
+        if qgroup in qgroup_list:
+            return True
+    return False
+
+
 def qgroup_id(pool, share_name):
     sid = share_id(pool, share_name)
     return '0/' + sid
 
 
 def qgroup_max(mnt_pt):
-    o, e, rc = run_command([BTRFS, 'qgroup', 'show', mnt_pt], log=True)
+    """
+    Parses the output of "btrfs qgroup show mnt_pt" to find the highest qgroup
+    matching QID/* if non is found then 0 will be returned.
+    Quotas not enabled is flagged by a -1 return value.
+    :param mnt_pt: A given btrfs mount point.
+    :return: -1 if quotas not enabled, else highest 2015/* qgroup found or 0
+    """
+    try:
+        o, e, rc = run_command([BTRFS, 'qgroup', 'show', mnt_pt], log=True)
+    except CommandException as e:
+        # disabled quotas will result in o = [''], rc = 1 and e[0] =
+        emsg = "ERROR: can't list qgroups: quotas not enabled"
+        # this is non fatal so we catch this specific error and info log it.
+        if e.rc == 1 and e.err[0] == emsg:
+            logger.info('Mount Point: {} has Quotas disabled, skipping qgroup '
+                        'show.'.format(mnt_pt))
+            # and return our default res
+            return -1
+        # otherwise we raise an exception as normal.
+        raise
+    # if no exception was raised find the max 2015/qgroup
     res = 0
     for l in o:
         if (re.match('%s/' % QID, l) is not None):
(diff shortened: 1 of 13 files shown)


phillxnet:
Note that in some situations the rfer and excl columns of 'btrfs qgroup show pool_mnt_pt' can be all zeros and consequently all share entries for that pool on the shares page show 0 for both Usage and Btrfs Usage.
A further execution of:
```
btrfs quota enable /mnt2/pool-name
```
can correct this situation. This appears to be related to the interplay docker-ce has with quotas.

src/rockstor/fs/btrfs.py
```diff
@@ -715,7 +798,18 @@ def qgroup_create(pool):
...
 
 def qgroup_destroy(qid, mnt_pt):
-    o, e, rc = run_command([BTRFS, 'qgroup', 'show', mnt_pt])
+    cmd = [BTRFS, 'qgroup', 'show', mnt_pt]
+    try:
+        o, e, rc = run_command(cmd, log=True)
+    except CommandException as e:
+        # we may have quotas disabled so catch and deal.
+        emsg = "ERROR: can't list qgroups: quotas not enabled"
+        if e.rc == 1 and e.err[0] == emsg:
```
schakrava: Just a minor concern. Do we need to check e.rc == 1 here? I don't know if btrfs-progs has standardized error codes, so my concern is if the message stays the same in the future where as code is changed for some reason. I'd think that the message is reliable and reliable enough.

src/rockstor/fs/btrfs.py
```diff
@@ -726,7 +820,19 @@ def qgroup_destroy(qid, mnt_pt):
...
     # Returns true if the given qgroup qid is already assigned to pqid for the
     # path(mnt_pt)
-    o, e, rc = run_command([BTRFS, 'qgroup', 'show', '-pc', mnt_pt])
+    cmd = [BTRFS, 'qgroup', 'show', '-pc', mnt_pt]
+    try:
+        o, e, rc = run_command(cmd, log=True)
+    except CommandException as e:
+        # we may have quotas disabled so catch and deal.
+        emsg = "ERROR: can't list qgroups: quotas not enabled"
+        if e.rc == 1 and e.err[0] == emsg:
```
schakrava: same comment as the previous one.

src/rockstor/fs/btrfs.py
```diff
@@ -736,15 +842,34 @@ def qgroup_is_assigned(qid, pqid, mnt_pt):
...
 
     # since btrfs-progs 4.2, qgroup assign succeeds but throws a warning:
     # "WARNING: # quotas may be inconsistent, rescan needed" and returns with
     # exit code 1.
     try:
-        run_command([BTRFS, 'qgroup', 'assign', qid, pqid, mnt_pt])
+        run_command([BTRFS, 'qgroup', 'assign', qid, pqid, mnt_pt], log=True)
     except CommandException as e:
         wmsg = 'WARNING: quotas may be inconsistent, rescan needed'
         if (e.rc == 1 and e.err[0] == wmsg):
```
schakrava: same comment as the previous one.

src/rockstor/fs/btrfs.py
```diff
@@ -785,6 +917,25 @@ def update_quota(pool, qgroup, size_bytes):
...
                         'limit.'.format(pool.name))
             return out, err, rc
+        # quotas disabled results in o = [''], rc = 1 and e[0] =
+        emsg2 = 'ERROR: unable to limit requested quota group: ' \
+                'Invalid argument'
+        # quotas disabled is not a fatal failure but here we key from what
+        # is a non specific error: 'Invalid argument'.
+        # TODO: improve this clause as currently too broad.
+        # TODO: we could for example use if qgroup_max(mnt) == -1
+        if e.rc == 1 and e.err[0] == emsg2:
```
schakrava: see previous comment

src/rockstor/fs/btrfs.py
```diff
@@ -785,6 +917,25 @@ def update_quota(pool, qgroup, size_bytes):
...
+            logger.info('Pool: {} has encountered a qgroup limit issue, '
+                        'skipping qgroup limit. Disabled quotas can cause '
+                        'this error'.format(pool.name))
+            return out, err, rc
+        emsg3 = 'ERROR: unable to limit requested quota group: ' \
+                'No such file or directory'
+        if e.rc == 1 and e.err[0] == emsg3:
```
schakrava: see previous comment

src/rockstor/storageadmin/views/share_helpers.py
```diff
@@ -78,20 +80,48 @@ def import_shares(pool, request):
...
+                    logger.debug('#### replacing void, non-existent, or '
+                                 'duplicate pqgroup')
+                    pqgroup = qgroup_create(pool)
+                    update_quota(pool, pqgroup, share.size * 1024)
+                    share_pqgroup_assign(pqgroup, share)
+                else:
+                    # Our share's pqgroup looks OK so use it.
+                    pqgroup = share.pqgroup
+                # Record our use of this pqgroup to spot duplicates later.
+                share_pqgroups_used.append(deepcopy(share.pqgroup))
```
schakrava: I am unable to figure out why we need a deepcopy here. Could you add a comment explaining why?

phillxnet: I don't have one worth of code comment but I can comment here that I was just playing safe: ie avoiding any accidental same object dual reference type thing and never revisited it. Your call on if absolutely required. I know they are expensive but just didn't want to risk accidental dual reference and have our taken copy still referencing the source, ie proper separation. I did all my testing of expected behaviour with the deepcopy in place.

Hope that helps on the why.

src/rockstor/fs/btrfs.py
phillxnet: The marked code is on the addition of the log=True entry, I found this useful when working on the issue so left it in, especially given that I don't thing we are going to be done with quota rework for a bit.

src/rockstor/fs/btrfs.py
phillxnet: Agreed. We could just check the message. I was just catching what I saw and checking both to be thorough/precise but as you say it may just trip us up later.

phillxnet:
Thanks for the review.

schakrava:
Thank you @phillxnet !

PR #1881, opened 2018-01-16
fix clone auto mount on creation. Fixes #1836

### LINKED ISSUE #1836: shares via clone of writable snapshot not mounted on creation
Shares, upon creation, are normally mounted. But in the case of a writable snapshot then 'turned' share via the clone icon on the snapshots page, the resulting share is not immediately mounted as per a normal share creation. After a number of minutes the share via clone mechanism is, eventually mounted; but refreshing of disks, pools, or shares page does not ensure this. A system reboot does however ensure this share from clone is mounted.

Rockstor version 3.9.1-13

phillxnet:
Previously when creating a new share via the clone mechanism from either an existing share or a writable snapshot the resulting share would not be auto mounted. This is not consistent will all other 'new' share scenarios. Also includes a qgroup / pqgroup fix where the resulting share was not immediately assigned a rockstor native 2015/x pqgroup and update_quota was incorrectly called on the auto created 0/x qgroup rather than the pqgroup.

abstract:
- add call to mount_share()
- add missing creation of pqgroup (Rockstor native 2015/x quota group)
- call update_quota() on Rockstor native 2015/x not on auto created 0/x quota group
- use now generated pqgroup in new share creation

Fixes #1836 

It was deemed appropriate that a newly created share, via the clone mechanism, be mounted as this is in keeping with existing new share creation scenarios. Please see issue text for an exposition of these existing counterparts.

@schakrava Ready for review.

Testing involved creating a clone share and observing no auto mount pre pr. Post pr the expected auto mount of the newly created share (clone) is confirmed.

Existing reservations: we may still have an auto assigned to native (2015/x) qgroup assignment issue here. However there are other suspected qgroup assignment issues and it is suggested that this issues reservation and it's suspected kin be addressed as a whole (in a focused issue): this should help with normalising and confirming and qgroup / pqgroup treatment project wide.

### DIFF
--- a/src/rockstor/storageadmin/views/clone_helpers.py
+++ b/src/rockstor/storageadmin/views/clone_helpers.py
@@ -18,7 +18,8 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 from storageadmin.models import (Share, Snapshot)
 from storageadmin.util import handle_exception
-from fs.btrfs import (add_clone, share_id, update_quota)
+from fs.btrfs import (add_clone, share_id, update_quota, mount_share,
+                      qgroup_create)
 from rest_framework.response import Response
 from storageadmin.serializers import ShareSerializer
 import re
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet

phillxnet:
@schakrava Cheers.

Issue #1853, opened 2017-11-09
suspected replication regression re share api change


phillxnet:
Thanks to forum member clawes in the following thread for highlighting this issue. It would appear that we have a share api change related regression in the replication code as indicated by the submitted error message:
```
[08/Nov/2017 14:00:36] ERROR [smart_manager.replication.sender:73] Id: 423B4A9B-3B81-99CA-55E5-1A7620713001-2. Failed to create snapshot: aim_2_replication_1. Aborting… Exception: [u’Invalid api end point: http://127.0.0.1:8000/api/shares/aim/snapshots/aim_2_replication_1’]
```
Linking to the potentially related api change pr for context:
#1808 "change share and snapshot endpoints to by id and not by name."

Please update the following forum thread with this issues resolution:
https://forum.rockstor.com/t/replication-is-failing-on-v3-9-1-16/4005/1

phillxnet:
@schakrava As this affects our latest stable I suggest that it be prioritised. I haven't looked at the replication code just yet but could do so, just not directly unfortunately. I'll note here if/when that changes.

phillxnet:
Please also update the following forum thread with this issues resolution; forum member Prabu_Raj may well be affected similarly:
https://forum.rockstor.com/t/replication-on-a-remote-server/1176/3

phillxnet:
Currently taking a quick look at this issue. I have reproduced and can confirmed that it is a share api change regression.

phillxnet:
Should have an initial pr ready for this one soon as have narrowed down the issue to be a now non valid assumption of similarity re share url references between sender and receiver. Previous to our share api change from name based to db id based (pr #1808) a share could be successfully validated remotely via it's name within the relevant url. This is what the replication receiver did and so even after translating the required url's to shareId there was still a failure to validate the senders share from the receiver. My initial pr against this issue is expected to simply remove, or simply, this validation element and update the required urls until further discussion can be had on how best to proceed, ie if we need to send more info to allow for this function to be re-established for example.

![successful-rep-task](https://user-images.githubusercontent.com/2521585/34653536-fa0d1706-f3e4-11e7-84de-928d13a3be76.png)

phillxnet:
Correct replication history table formatting to be in line with most other project table formats, ie with search and sort (data-table) and ensured history is most recent first via ID desc order sort.

Before:
[...]

PR #1883, opened 2018-01-22
remove immutable flag prior to share delete. Fixes #1882

### LINKED ISSUE #1882: remove immutable flag prior to share delete
Thanks to forum members Haioken and Rene_Castberg for highlighting this issue (and it's current workaround). Under as yet undetermined circumstances a share (btrfs subvol) can have it's immutable flag set. The consequence of this is a failure to be able to delete the affected share. Given that Rockstor currently has no immutable flag observance and that the cause of this flag having been set is unknown it seems prudent for the time being to be robust in this scenario and simply remove the flag prior to a share's  delete (which is itself accompanied by a GUI warning of relevant data loss).

This issue was also observed to randomly break replication whilst working on #1853 as the receiver periodically removes and supplants a share with a received snapshot. As and when the true cause of this flag's having been set is identified we can more correctly observe it's use project wide (along with appropriate user messaging), but currently our base ability to manage shares and have more robust replication is judged as taking priority. This element is a key factor in breaking out this change from #1853 as it is important that it be revisited in due cause in it's own right.

Pull request of tested fix to follow shortly.

https://forum.rockstor.com/t/solved-btrfs-subvolume-create-delete-fail/3943

phillxnet:
As an immutable flag can break share delete and given we already provide a UI warning, this preemption and circumvention of an existing immutable flag is deemed appropriate. Especially in the context of potentially breaking replication: during snap-to-share supplanting. It is however advised to re-address this approach once the sporadic nature of immutable flags appearing without manual intervention is resolved.

Fixes #1882 

Please see issue text for more context.

@schakrava Ready for review.

Testing involved manually applying the immutable flag to a share ie via:

    chattr +i /mnt2/rock-pool/pppp-test1

Where pool.name = 'rock-pool' and share.name = 'pppp-test1'

and then attempting to delete that share via the UI. Pre pr an "Operation not permitted" was surfaced in the UI and the share remained but was left unmounted. Post pr the share (subvol) was deleted as expected.

Note that this pr is considered as a desirable pre-requisite of the pr to be associated with issue #1853 .

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -39,6 +39,7 @@ MKFS_BTRFS = '/sbin/mkfs.btrfs'
 BTRFS = '/sbin/btrfs'
 MOUNT = '/bin/mount'
 UMOUNT = '/bin/umount'
+CHATTR = '/usr/bin/chattr'
 DEFAULT_MNT_DIR = '/mnt2/'
 RMDIR = '/bin/rmdir'
 QID = '2015'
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet

phillxnet:
@schakrava Cheers.

PR #1885, opened 2018-01-24
Fix replication regression re share api change. Fixes #1853

### LINKED ISSUE #1853: suspected replication regression re share api change

### LINKED ISSUE #1882: remove immutable flag prior to share delete

phillxnet:
Prior function was restore by updating the relevant api urls in the replication system. But in one, non critical check this was not managed. As a result the non critical but desirable internal sanity check was, for the time being, remarked out and TODOs added to signify required future attention. Also includes a number of replication UI fix ups including table formatting, sort capability, appropriate ordering, and consistency with the rest of the UI.

A core mechanic of the replication system was also abstracted such that it could benefit from the existing share / snapshot management system. This alteration made more explicit, in code, a quirk involving the initial subvol transferred and how it requires a special treatment given it's unique nature in the context of the existing share / snapshot / clone / import structures.

Summary:
- Update internal replication api urls to fit new share by id scheme.
- Remove problematic non critical internal check (TODO added for later).
- Create new repclone command to improve replication share / snap handling, ie more transactional by using existing structures via api.
- Minor improvements were added in the new repclone system, ie sys refresh qgroup transfer, auto mounting etc.
- Ensure we import initial btrfs receive quirk subvol with replica flag and add logging to indicate it's quirk nature.
- Improve replication tables formatting, text, sort ability / order.
- Reduce and mirror (between send and receive) rep-snap-count.
- Fix bugs in receive status update, includes utilising the pending state.
- Improve user communication re data transferred / rate table cell text.
- Remove broken pool link in replication overview table.
- Improve debug logging re share / snap import.
- Minor steps / TODOs towards future share name/subvol_name seperation.
- TODOs towards centralising mount path creation.
- Various TODO added for future consideration.

N.B. this pr assumes the prior application of the following (currently pending) pull request:
"remove immutable flag prior to share delete. Fixes #1882" pr #1883 
As that fix also addresses an observed breakage during replication runs and was used (pre-applied) during all testing of this pr's code.

Fixes #1853 (if #1883 in pre-applied)

@schakrava Ready for review.

Testing included several hundred individual replication events (btrfs send receive hosts) with runs up to 150 ish. A real hardware arrangement was also tested where the source machine's (sender) share had more data than could be replicated in the chosen scheduled replication interval (10 mins). Additional send receive pairs were not activated and all data was successfully transferred.

Note that with the in pr rep-snap-count change to 2, 3 snapshots are stored both on the sender and against the receivers share counterpart. And that for a replication cycle to reach it's final state, which will in turn there after be maintained by rotation, 5 replication events must be completed: the initial full-send followed by 4 incremental sends.

There are still known limitations and bugs in the replication code post pr but these can be addressed more specifically in pending issues to be opened in due course: and depend on review outcome of this pr.

### DIFF
--- a/conf/settings.conf.in
+++ b/conf/settings.conf.in
@@ -314,7 +314,7 @@ SCHEDULER = ('127.0.0.1', ${django-settings-conf:schedulerport})
 REPLICATION = {
 	    'ipc_socket': '/var/run/replication.sock',
 	    'max_send_attempts': 10,
-	    'max_snap_retain': 5,
+	    'max_snap_retain': 2,
 	    'listener_port': 10002,
 }
 
--- a/conf/test-settings.conf.in
+++ b/conf/test-settings.conf.in
@@ -301,7 +301,7 @@ SCHEDULER = ('127.0.0.1', ${django-settings-conf:schedulerport})
 REPLICATION = {
     'ipc_socket': '/var/run/replication.sock',
     'max_send_attempts': 10,
-    'max_snap_retain': 5,
+    'max_snap_retain': 2,
     'listener_port': 10002,
 }
 
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -351,6 +351,9 @@ def add_share(pool, share_name, qid):
 
 
 def mount_share(share, mnt_pt):
+    # TODO: Consider making mnt_pt optional as per helper_mount_share() as then
+    # TODO: we could remove almost system wide many duplicates of temp mnt_pt
+    # TODO: created just prior and only for this methods call.
     if (is_mounted(mnt_pt)):
         return
     mount_root(share.pool)
@@ -560,6 +563,7 @@ def remove_share(pool, share_name, pqgroup, force=False):
     mount root pool
     btrfs subvolume delete root_mnt/vol_name
     umount root pool
+    :param pool: pool object
     """
     if (is_share_mounted(share_name)):
         mnt_pt = ('%s%s' % (DEFAULT_MNT_DIR, share_name))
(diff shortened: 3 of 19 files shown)


schakrava:
Nice work, patiently going through my implementation, making much awaited changes, making things better and more explicit! Thank you @phillxnet

phillxnet:
@schakrava Thanks, and apologies for the merge conflict. Seems like I dropped the ball a little there.

Issue #1347, opened 2016-06-13
Issues with afp share name


djevo1:
It would be useful to be able to change the share name so it doesn't contain any space. This is the case when it adds on rockstor at the end of the share name. It is currently only fixable if the netatalk conf file is edited to remove the name of the share.

timorjim:
I agree, having share names with "on rockstor" attached is awful IMO.

phillxnet:
Linking to forum member fva.taimyr recent post on a potential solution to this long standing issue:
https://forum.rockstor.com/t/chnage-share-name-on-afp/1597/3

FroggyFlox:
Closing as we no longer use AFP so this issue is now irrelevant.

PR #1895, opened 2018-02-13
Issues with afp share name #1347

fva-dev:


schakrava:

schakrava:

phillxnet:
Having a quick look at this now.

src/rockstor/storageadmin/views/netatalk.py
```diff
@@ -122,8 +122,9 @@ def post(self, request):
         try:
             for share in shares:
                 mnt_pt = ('%s%s' % (settings.MNT_PT, share.name))
-                cur_description = '%s %s' % (share.name, description)
```
phillxnet: From having a look at the original code and your proposed change, Your changes work. So that's a plus. But they do change the original logic which looks like it was intended to allow for the addition of a user input description, and if not found would default to this Share.name + " on Rockstor" which has proved to be contentious. That is share.name is never == '' or if it is we have other problems :).
I think for the time being a better approach to this issue would be to only apply your line 125 changes and additionally:
change line 95 from:

    def_description = 'on Rockstor'

to

    def_description = ''

I.e take out the contents between the 2 single inverted commas. That way the rest of the optional description handling stuff can await further attention on the UI front but we still end up with dropping the " on Rockstor" bit. It's a bit hacky but should do for now to force change our default AFP export name to that of the share.

phillxnet:
On my Inline comment re "RackMac" to "Rack Mac (G)" my finding on El Capitan was as follows (pre export name changes):
RackMac:
pre-pr-backups-on-rockstor

and RackMac (G)
rackmac g -change

Note however that this icon is pretty sticky in finder, I had to Force close Finder to get the update to RackMac (G) but was unable to get the icon to revert to the previous imac image.

And a pic to prove export name changed as intended for me (Share name is Backups)
afp-export-as-share-name
:

src/rockstor/system/services.py
```diff
@@ -223,7 +223,9 @@ def refresh_afp_config(afpl):
         rockstor_section = False
         tfo.write('; Netatalk 3.x configuration file\n\n')
         tfo.write('[Global]\n')
-        tfo.write('mimic model = RackMac\n\n')
+        tfo.write('mimic model = RackMac\n')
```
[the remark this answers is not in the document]
fva-dev: Without this line:
"uam list = uams_dhx.so,uams_dhx2.so"
i have a problem with connection to afp in osx Yosemite and El Capitan.

src/rockstor/storageadmin/views/netatalk.py
fva-dev: Good idea is to add description field for every share name.

src/rockstor/system/services.py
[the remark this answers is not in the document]
phillxnet: @fva-dev OK and thank for the follow up, I've just rechecked here on my El Capitan and Fedora 26 clients (AFP protocol within Nautilus) and both worked fine from fresh boot without that line. But I only have a single El Capitan machine fully updated as of yesterday. I tested with this share only exported via AFP. Note that SMB is now favoured for share access by recent OSX so I wonder if there is some interplay if a share is dual exported. Could you confirm this lines requirement at your end, ie remove the line and see if you can still gain access to the share - fresh boot on both client and server. I'm just using the share access as per "Connect As" top right of finder.

If this line does help then great, just a bit concerned as per my comment previously.

src/rockstor/storageadmin/views/netatalk.py
phillxnet: Yes, it would be good to add this feature to the UI in the future, especially given the backing code is already done.

schakrava:
Thank you @fva-dev and @phillxnet Sorry it took me a while to verify this.

PR #1899, opened 2018-02-16
empty compression setting on imported pools blocks extra mount options. Fixes #1896

### LINKED ISSUE #1861: inconsistent root pool mount edit options
Thanks to forum member razaborg in the following thread for helping to highlight this issue. In pool overview we have compression and extra mount options disabled for this pool, but in the same pool's details page these options are active. However any attempt to use them fails as we have backend preventative logic to catch these currently unsupported actions.

It is proposed  that we also disable these currently dysfunctional options in the system pool details page to gain consistency and then address, at a later date, the 'special case' of changing this system pool's mount options.

https://forum.rockstor.com/t/unable-to-change-mount-options-for-the-rockstor-rockstor-pool/4128

### LINKED ISSUE #1896: empty compression setting on imported pools blocks extra mount options
Thanks to forum member g6094199 in the following thread for highlighting this issue. When a pool is imported it's default "Compression" is displayed as "Empty". While this default setting persists it is not possible to add "Extra mount options" as the following input validation error of the unchanged "Empty" compression setting is thrown:

Unsupported compression algorithm(). Use one of ('lzo', 'zlib', 'no')

This does not affect newly created pools that default to 'no' for compression and once an imported pool has it's compression set, subsequent attempts to add/edit "Extra mount options" succeeds: given associated validation checks pass.

The issue author suggests that "None" or "OFF" (as per non editable system pool indicator) would be a better default setting indicator than “no”. And that these indicator should be consistent named.

Please update the following forum thread with this issue's resolution:
https://forum.rockstor.com/t/unsupported-compression-algorithm-use-one-of-lzo-zlib-no/4381

phillxnet:
Newly created pools have Pool.compression set to 'no' however imported pools, and the system pool, defaulted to Null / None for this field. Consequently "Empty" was indicated by x-editable inline compression edit on pool overview and detail pages. This entry and it's consequent value failed the dual submission of compression and mount options.

Summary:

- Set imported pools to 'no' compression by default.
- Handle "Empty" output from inline editable of compression setting.
- Put "Empty" inline editable colour in line with links as not an error condition (to allow for legacy imported pools).
- Change "Empty" legacy pool imports compression text to "Unset".
- Improve and normalise compression tooltip text.
- Add 'compression' and 'extra mount options' tooltips to Pool overview.
- Remove inconsistent root pool mount edit options see issue #1861 where compression and mount options were offered on system drive details page.
- Display system Pool.compression and Pool.mnt_options as per other pools but without edit options: i.e. 'no', if set, rather than 'OFF' as a hard wired value.

Fixes #1896 
and additionally:
Fixes #1861 
Please see both issues texts for context.

@schakrava Ready for review.

Test for 'proof of fix':

Imported pools on fresh build and ensured that their 'extra mount options' could be inline edited prior to their 'Empty' (post pr 'Unset') compression entries having a selection made. Also edited system pool's compression and mnt_options entries in storageadmin_pool db to ensure future configuration on these elements will display as intended. Tested a completely fresh build to ensure system pool had it's Pool.compression set as intended.

Outstanding caveat: When editing both compression and mount options in Pool details page without a page refresh between them, stale settings can be re-applied. This is due to the Pool info held by that page not being refreshed and the js using this outdated info on the second settings submission.  This bug can be addressed separately against an issue opened for this purpose.


schakrava:
I like this pr a lot. Thanks @phillxnet !

phillxnet:
@schakrava Thanks and Cheers.

Issue #1700, opened 2017-04-26
Implement a delete missing disk in pool UI


phillxnet:
Thanks to zoombiel in the forum and GitHub users @therealprof and @iFloris (#737) for highlighting this issue. At time of issue creation we have no UI means of addressing the removal of a missing device from a pool. I proposed that we extend our existing 'bin' icon to link to a specialised page to explain the detached status, ie this could be a backup drive that is not permanently attached (likely then the only member of a pool if it is a btrfs pool member at all) or it could be a member of a multi disk pool, in which case there is a high likely hood that the observed situation is a fail condition. We could in turn link to the associated pool details page where it is proposed we implement a "Remove Missing Devices" option which in turn runs the following:
```
btrfs device delete missing mountpoint
```
Reference: https://btrfs.wiki.kernel.org/index.php/Using_Btrfs_with_Multiple_Devices
Section: Using add and delete:
"... if a device is missing or the super block has been corrupted, the filesystem will need to be mounted in degraded mode:"
"btrfs device delete missing /mnt"
"btrfs device delete missing tells btrfs to remove the first device that is described by the filesystem metadata but not present when the FS was mounted."

N.B. As this option depends upon a degraded mount, it could also be presented only under those circumstance.

It is favoured that this option be located in the pool UI as it exists at the pool level of analysis: hence the link from disk to pool to emphasise the affect level.

A pre-requisite of this issue is to address pool health reports which in turn will require developing to be able to recognise the 'degraded' pool state and additionally, in a missing drive scenario, offer the option to mount degraded with all the associated warnings, ie the current one shot deal, if the drive count is at minimum, including the missing drive, for the given raid level.
Linking to related issues on pool health reporting: #1531 , #1199 , 

Please update the following forum thread with this issues resolution:
https://forum.rockstor.com/t/remove-failed-disk-from-pool-no-replacement/3151
Also note that this issue is potentially a duplicate (but with additional implementation details) of GitHub user @therealprof;s #737 with additional contributions from @iFloris .

phillxnet:
Please also update the following forum thread with this issues resolution:
https://forum.rockstor.com/t/cant-remove-failed-drive-from-pool-rebalance-in-progress/3319

phillxnet:
[...]

PR #1902, opened 2018-03-06
surface pool missing disk info in UI. Fixes #1897

### LINKED ISSUE #1897: surface pool missing disk info in UI
The current 'detached-..." disk info pertains to prior attached disks that are, on subsequent scan / boot, no longer attached, and although these disk level entries maintain their prior pool association info, if previously known, and all pool level views, ie pool overview / details pages, reflect these associations in their respective table "Disks" columns, we do not resource / surface the canonical pool level info on missing devices.

It is proposed, as a prerequisite of:
"Implement a delete missing disk in pool UI #1700",
that we also surface the pool level missing device info, as per 'btrfs fi show' for example, such that all imported pools, irrespective of mount state,  feature this pool level missing device info. This is particularly pertinent in situations such as when a pool is first imported in a degraded state, where no prior attached and subsequent 'detached..." disks can be indicated: as they have never been attached.

phillxnet:
By adding a lighweight 'has_missing_dev' property to the Pool object we can conveniently surface pool degraded status within the UI. The pools overview and detail pages can then be informed by this property and consequently display red "Some Missing" indicators within the pool's respective disks section. In addition another mechanism linked to a new gevent is used to inform system wide pool degraded status to the Web-UI header, so that no page refresh is required. This second mechanism can also report on the number of non managed / pre-imported degraded pools. The extraction of pool labels for un-managed pools is purposefully avoided in order to keep the initial detection algorithm as light as possible. On systems with no degraded pools the detection is down to a single system call. Pool Object creation for the header gevent notifications is avoided until at least one degraded pool is detected.

Summary:

- Add serialized 'has_missing_dev' boolean property to Pool Object.
- Add lightweight pool degraded detection to inform above.
- Add red "SOME MISSING" UI element to Disks column in Pool overview.
- Add red "Some Missing" UI element to Disks title in subsection of Pool details view.
- Add system wide pool degraded detection to inform Web-UI header, via a gevent, so that even on a freshly installed system, unimported pools can be indicated. This system does not require a Web-UI refresh and is modeled after the existing 'shutdown-status' indicator.
- Add a unit test to prove the function of the above system wide pool degraded sensor.
- Extend the existing back-end pool-info system to encompass degraded state along with expected total disk count and missing disk count. These additional facilities are not employed in this pr but will serve as a double check on Pool object info to inform future mechanisms before enacting on such operations as delete missing.
- Add TODO on informing system pool's status.

Fixes #1897 
Please see comment section for context and UI element images.

@schakrava Ready for review.

Tested in KVM by removing select virtio disks to 'degrade' one or more system pools. Both imported and non imported. The system wide, rather than individual pool, degrade detection mechanism was also developed along side a unit test included with this pr:

test_degraded_pools_found (fs.tests.test_btrfs.BTRFSTests) ... ok

All 3 degraded indicators were tested via development logging to reflect the observed pool degraded status as per 'btrfs fi show'.

Caveat: In some specific instances 'btrfs fi show' would indicate a degraded state that would then not be evident when a normal mount was attempted (as it succeeded without the degrade option). The system wide degrade detection algorithm does not report this instance (test case included for this caveat with relevant comments) but the pool specific degrade detection does indicate this. It was thought that this diversification was healthy and would help to inform future development on this issue. The 3rd pool degrade detection mechanism, added to get_pool_info(), is as yet unused but is intended to acts as a low level validation mechanism for future pool actions such as are indicated in:
"Implement a delete missing disk in pool UI #1700"
for which this pr is considered a prerequisite.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -85,46 +85,110 @@ def add_pool(pool, disks):
     return out, err, rc
 
 
+def is_pool_missing_dev(label):
+    """
+    Simple and fast wrapper around 'btrfs fi show --raw label' to return True /
+    False depending on if a device is reported missing from the given pool by
+    label. Works by matching the end of output lines for the string 'missing',
+    which is either, post remount reboot (first and last populated line) or
+    the last line (dev removed and no remount).
+    Label is used as this is preserved in our Pool db so will work if the pool
+    fails to mount, and there by allows surfacing this as a potential reason
+    for the mount failure.
+    :param label: Pool label.
+    :return: True if at least one device was found to be missing, False if not.
+    """
+    if label is None:
+        return False
+    # --raw used to minimise pre-processing of irrelevant 'used' info (units).
+    cmd = [BTRFS, 'fi', 'show', '--raw', label]
+    o, e, rc = run_command(cmd)
+    if o[-3].endswith('missing') or o[0].endswith('missing'):
+        return True
+    return False
+
+
+def degraded_pools_found():
+    """
+    Primarily intended to indicate the existence of any degraded pools, managed
+    or otherwise. Originally used by data_collector to feed real time Web-UI
+    indicators. Non-managed pool coverage allows for the indication of a
+    degraded mount requirement pre-import or on fresh disaster recovery
+    installs.
+    :return: Number of degraded pools as indicated by any line ending in
+    "missing" following an associated "Label" line.
+    """
+    # --raw used to minimise pre-processing of irrelevant 'used' info (units).
+    cmd = [BTRFS, 'fi', 'show', '--raw']
+    o, e, rc = run_command(cmd)
+    degraded_pool_count = 0
+    in_pool = False
+    for line in o:
+        if not in_pool and line[0:3] == 'Lab':
+            in_pool = True
+            continue
+        if in_pool and line.endswith('missing'):
+            # we are in pool details and have found a missing device
+            degraded_pool_count += 1
+            # use in_pool switch to avoid counting this pool twice if it has
+            # multiple missing as at least 1 missing dev is degraded.
+            in_pool = False
+        elif line == '':
+            # pool listings delimited by blank lines
+            in_pool = False
+    return degraded_pool_count
+
+
 def get_pool_info(disk):
     """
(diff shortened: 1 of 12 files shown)


src/rockstor/fs/btrfs.py
```diff
@@ -85,46 +85,110 @@ def add_pool(pool, disks):
...
+    for the mount failure.
+    :param label: Pool label.
+    :return: True if at least one device was found to be missing, False if not.
+    """
+    if label is None:
+        return False
+    # --raw used to minimise pre-processing of irrelevant 'used' info (units).
+    cmd = [BTRFS, 'fi', 'show', '--raw', label]
+    o, e, rc = run_command(cmd)
+    if o[-3].endswith('missing') or o[0].endswith('missing'):
```
schakrava: this seems fuzzy. If btrfs is inconsistent, may be better to iterete over all output lines and search for the matching substring. Also, can the substring be a bit more specific than just the word "missing"?

schakrava:
Very nice @phillxnet ! I added a minor comment for your consideration going forward. This is a cool feature to have!

src/rockstor/fs/btrfs.py
phillxnet: "... may be better to iterete over all output lines and search for the matching substring."
My concern here was to be as light/simple and fast as possible as we are a model property at this point.

I know what you mean though as I had originally just searched all lines for string.endswith('missing') but took it down to last (non empty) line and doubled checked against the first line just in case as this was the inconsistent one. We can later move to libbtrfsutil for this kind of stuff, once btrfs-progs is rebased on it and it becomes canonical. Should help to remove all this less than ideal parsing. Prior to that we can use the soon to be added json output switch which will be nice.

As for "missing" that seemed to cover all that I saw which was really only:
"*** Some devices missing" on last non empty line.
and the likes of:
"warning, device 5 is missing" which obviously varied (by id) and was also sometime repeated verbatim !!
Hence "missing" seemed to cover it.

Looking forward to json output options and the new libtrfsutil :)

phillxnet:
@schakrava Thanks for the review and glad you liked it. Agreed on parsing comments and will have to see how it goes. This capability should help to inform the pool recovery tools going forward.

PR #1903, opened 2018-03-07
Add option to disable BTRFS quota-qgroups. Fixes #1592

### LINKED ISSUE #1592: [New feature] Add option to disable BTRFS quota/qgroups
Qgroups seem to cause a lot of trouble / performance issues. Personally I get errors like `BTRFS warning (device sda): orphan qgroup relation 0x563->0x7df00000000000a`.
Picking up an idea from a [forum post by Matthew](https://forum.rockstor.com/t/somethings-broken-the-mounts/2241/8) it would be nice to be able to completely disable quota related functions in Rockstor.

### LINKED ISSUE #1785: Feature: Quota rebuild script?
Could we have a script/CLI that causes rockstor to re-enable quotas and re-create any qgroups it needs then do a rescan?

The reason for disabling quotas is when re-balancing or deleting a large amount of snapshots it's considerably faster to disable them perform the operation and then re-enable them and rescan.

By considerably I mean with quotas off a re-balance takes hours to days depending on the array, with it on we're talking weeks to months.

Last time I re-balanced it took 2 weeks to do the first 50% then I turned off quotas and the remainder was done in half a day.

### LINKED ISSUE #1869: improve quotas not enabled behaviour

phillxnet:
Adds pool api command to enable / disable quotas. Upon enabling a quota rescan is initiated. In some instances this scan is redundant but is robust to those instances and simply info logs the prior existence of an ongoing quota rescan and skips the rescan requested. Similarly provision is made to deal elegantly with quota rescan requests on read only filesystems. UI is by way of bootstrap inline editable as per current compression and custom mount options settings.

Summary:
- Add backend pool quota enable/disable command.
- Change existing quota status displays for inline editable elements.
- Move pool details page quota element from "Details" section to "Compression and extra mount options" section as then all inline edit elements are grouped.
- Rename compression / extra mount options header to indicate quotas.
- Allowed quotas enable/disable on the system pool.
- Initiate a quota rescan when re enabling via the inline edit UI element / api command.
- Improve user messaging (via tool tips) re 0 usage indication on Shares as a consequence of quotas disabled (currently).
- Indicate that the default quote state is enabled but provide the performance related argument for them being disabled (via tool tips).

Fixes #1592 
Additionally as we initiate a rescan upon enabling quotas on a given pool this pr also:
Fixes #1785 

@schakrava Ready for review.

This pr represents the enable/disable api and front end to the work done for issue:
"improve quotas not enabled behaviour #1869"
and it's pr:
"improve quotas not enabled behaviour. Fixes #1869" #1874
re quota groups management / recreation through quota disabled / enabled cycles.

Tested in KVM systems and on a number of small real hw installs in the low TB range of pool size.

Caveats / future enhancements: we currently give no indication of ongoing rescans, this could properly be addressed against it's own issue. The consequence of not surfacing an ongoing quota rescan is potential user confusion re share size reporting which will periodically adjust for a number of minutes after re enabling quotas (with quotas disabled shares are reported as taking 0 space (currently)).

### DIFF
--- a/src/rockstor/storageadmin/views/pool.py
+++ b/src/rockstor/storageadmin/views/pool.py
@@ -28,7 +28,7 @@ from storageadmin.serializers import PoolInfoSerializer
 from storageadmin.models import (Disk, Pool, Share, PoolBalance)
 from fs.btrfs import (add_pool, pool_usage, resize_pool, umount_root,
                       btrfs_uuid, mount_root, start_balance, usage_bound,
-                      remove_share)
+                      remove_share, enable_quota, disable_quota, rescan_quotas)
 from system.osi import remount, trigger_udev_update
 from storageadmin.util import handle_exception
 from django.conf import settings
@@ -83,6 +83,20 @@ class PoolMixin(object):
         except:
             e_msg = ('Problem with role filter of disks' % disks)
             handle_exception(Exception(e_msg), request)
+
+    @staticmethod
+    def _validate_new_quota_state(request):
+        logger.debug('#### validate_new_quota_state received new_state '
+                     '=({}).'.format(request.data.get('quotas')))
+        new_val = request.data.get('quotas', 'Enabled')
+        if new_val is None:
+            # We default to Quotas enabled if input is in doubt.
+            new_val = 'Enabled'
+        if new_val != 'Enabled' and new_val != 'Disabled':
+            e_msg = ('Unsupported quotas request ({}). '
+                     'Expecting "Enabled" or "Disabled"'.format(new_val))
+            handle_exception(Exception(e_msg), request)
+        return new_val
 
     @staticmethod
     def _validate_compression(request):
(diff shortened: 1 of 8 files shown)


src/rockstor/storageadmin/views/pool.py
```diff
@@ -211,6 +225,39 @@ def _remount(cls, request, pool):
...
+            return Response()
+        try:
+            if new_quota_state == 'Enabled':
+                # Current issue with requiring enable to be executed twice !!!
+                # As of 4.12.4-1.el7.elrepo.x86_64
+                # this avoids "ERROR: quota rescan failed: Invalid argument"
+                # when attempting a rescan.
+                # Look similar to https://patchwork.kernel.org/patch/9928635/
+                enable_quota(pool)
+                enable_quota(pool)
```
schakrava: It may be useful to open an issue to keep track of future refactoring of this logic.

schakrava:
This is really nice! I am eager to try disabled quotas on our staging boxes and see the performance differences.

src/rockstor/storageadmin/views/pool.py
phillxnet: Agreed. I have just created what I think should cover it:
"re-address double enable_quota() call requirement" #1905

phillxnet:
@schakrava Thanks. I'll try and revisit the sub optimal double quota thing on the next kernel / btrfs-progs update. I was a little surprised by that behaviour and I'm keen to re-address it as soon as we can.
Cheers for the review.

PR #1912, opened 2018-04-01
Normalise storageadmin exception msg format and update unit tests. Fixes #1847

### LINKED ISSUE #1847: root pool edit message formatting type error
Thanks to forum member Spiceworld for reporting this issue. We have a regression in the pool edit limitation exception message which prevents the full message being displayed.
```
"/opt/rockstor/src/rockstor/storageadmin/views/pool.py", line 351, in put 'as it contains the operating system.' % pid) TypeError: %d format: a number is required, not unicode
```
In prior examples of this type of regression, caused by a recent pool api change from name to db id in this case, we have resolved the unicode format issue by moving to the string.fomat method.

https://forum.rockstor.com/t/trying-to-remove-disks-from-pool-and-getting-following-errors/3962

phillxnet:
Normalise the construction and syntax of exception messages within storageadmin and its associated unit tests so that they adhere to the following rules:

- Exception message strings and their consumer counterpart comparisons (eg: unit tests and smart_manager’s replication) are to be formatted using string.format.
- Each error message starts with a capital letter and ends in a full stop.
- Where brackets are used to for example display a pool / share name we surround the brackets with spaces. Ie not: "Pool(leon) already exists." but: "Pool (leon) already exists.": eases readability.
- Numbers are expressed in their numeric variant where appropriate: ie '4' not 'Four': eases translation.
- Where a space is to exist in a multi part string we put this on the trailing end of the prior element.
- Where a standard unit designation is used, ie 'GB', we precede it with a space ie "6.7 GB".
- When quoting a system message at the end of our error message string we surround that message in brackets "()" so as to discern our full stop from any it may have.
- 'Pool', 'share', 'snapshot', 'rock-on', and 'disk' are not proper nouns so are not to be capitalise unless used to reference Web-UI entries ie "A Disks page Rescan may help." where both 'Disks' and 'Rescan' reference UI elements that are also Capitalised.

Begin updating existing storageadmin unit test fixtures re mocking and exception message interpretation where necessary as they should help to prevent the observed regression type going forward.

Unit test restoration informed additional model.save() additions in views share.py and snapshot.py plus the addition of a conditional delete() in nfs_export.py.

Specified CommandException (in osi.py) to prevent PyCharm’s "No exception to reraise" error which hampers that IDE’s function re code analysis. The resulting code was tested to ensure it’s behaviour was as expected.

Remove build blocking gulp-eslint step.

Fixes #1847 

@schakrava Ready for review.

There is more work to be done in both of the formatting and unit test areas but this pr represents a start on the indicated goals.

Testing:
Pre pr the indicated issue re exception message formatting is reproduced by attempting to add a disk to the system pool, ie rockstor_rockstor. This UI option should not currently be available but remains as the only trigger of this exception message post the closing of issue:
“inconsistent root pool mount edit options” #1861.
Post pr the message is formatted without error and displayed as expected.

Pre pr Unit test results via:

    ./bin/test -v 2 -p test_*

    Ran 29 tests in 0.857s 
    FAILED (errors=25)

Post pr unit test results:

    Ran 158 tests in 23.549s
    FAILED (failures=19, errors=4)

Current Unit test output along with a successful issue message format is to follow in the comments section.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -36,7 +36,6 @@ parts =
       js-sync
       collectstatic
       gulp-install
-      gulp-eslint
       supervisor
       supervisord-conf
       create-cert
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -1341,7 +1341,17 @@ def balance_status(pool):
     ie indexed by 'status' and 'percent_done'.
     """
     stats = {'status': 'unknown', }
-    mnt_pt = mount_root(pool)
+    # The balance status of an umounted pool is undetermined / unknown, ie it
+    # could still be mid balance: our balance status command requires a
+    # relevant active mount path.
+    # Note that if we silently fail through the mount confirmation then our
+    # balance status will reflect the system pool balance status.
+    try:
+        mnt_pt = mount_root(pool)
+    except Exception as e:
+        logger.error('Exception while refreshing balance status for Pool({}). '
+                     'Returning "unknown": {}'.format(pool.name, e.__str__()))
+        return stats
     out, err, rc = run_command([BTRFS, 'balance', 'status', mnt_pt],
                                throw=False)
     if (len(out) > 0):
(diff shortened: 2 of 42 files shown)


phillxnet:
@schakrava  I may have made a final squash error here as the final test output is now different. Best if you hold off a merge for a bit as I investigate.

PR #1916, opened 2018-04-04
On GUI only 5000 user show up. Fixes #1915

### LINKED ISSUE #1915: On GUI, only 5000 user show up
Hello there,

In this forum, we wrote the problem:

https://forum.rockstor.com/t/rockstor-dont-show-all-users-of-my-domain-active-directory/2898

In my AD is more than 5000 user, but now i only see users start with "y*" but not see the user start with like "z*"

It is possible to fix this issue, that we able to see all user? (if i list it on CLI, i see all of them)

phillxnet:
Increase maxPageSize from 5000 to 9000 predominantly to deal with larger installs, ie > 5000 users / groups. System wide change so should also accommodate all other datatable displays within the Web-UI.

Includes the following 2 unrelated minor dev / build related fixes:

- Remove gulp-eslint stage: currently blocks builds.
- Use specific CommandException to avoid conformance (ide) errors. This is to prevent PyCharm’s "No exception to reraise" error which hampers that IDE’s function re code analysis. The resulting code was tested to ensure it’s behaviour was as expected. Further comments re known fake / non unique motherboard uuids were added.

Fixes #1915 

@schakrava Ready for review

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -36,7 +36,6 @@ parts =
       js-sync
       collectstatic
       gulp-install
-      gulp-eslint
       supervisor
       supervisord-conf
       create-cert
--- a/src/rockstor/storageadmin/static/storageadmin/js/rockstor.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/rockstor.js
@@ -426,7 +426,7 @@ RockStorGlobals = {
     navbarLoaded: false,
     applianceNameSet: false,
     currentAppliance: null,
-    maxPageSize: 5000,
+    maxPageSize: 9000,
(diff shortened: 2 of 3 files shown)


phillxnet:
@schakrava I missed an existing Flake8 issue in osi.py on this pr (line 65). Will try and include on the next visit to this file. I don't think it's worth re-doing the pr especially since it was not introduces here. Let me know if you would rather I fix it this time around.

phillxnet:
@schakrava Cheers.

phillxnet:
Thanks to @vampywiz17 (of this pr's issue #1915) for confirming this fix.

PR #1924, opened 2018-05-13
improve subvol mount code robustness. Fixes #1923

### LINKED ISSUE #1880: check on quota updates re share rollback
In storageadmin/views/share_command.py we call update_quota using the 0/x auto created group name. This is not consistent will all other calls of update_quota where the Rockstor native 2015/x pqgroup is used: bar a single other instance that is about to be corrected in a pending pr associated with an otherwise unrelated issue concerning clone creation.  As this change will require some consideration and testing this issue has been opened to address it. Note that there may also be some confusion of snap and share in the associated code. Again further focused consideration would be appropriate as the rollback method may have unique requirements re the quota groups and their transfer. We may also be missing a snap.save() here.

### LINKED ISSUE #1923: improve subvol mount code robustness
Currently we mount subvols via '-o subvol'; it is proposed that a number of improvements to mount code clarity / robustness / flexibility can be had by moving to subvolid: which we already track in both the share and snapshot models via the qgroup field, ie '0/x' where x is the subvolid.  This in turn leads to a number of code simplifications where we are able to avoid redundant system subvol id discovery calls by passing the already retrieved qgroup from the model. And by not relying on the subvol name we avoid potential ambiguity re subvol path elements.

A proof of concept pr for this issue is currently in preparation and should follow shortly.

phillxnet:
Mount subvols via subvolid derived from share/snap model qgroup fields.

A significant element of this commit/pr is the deprecation / removal of rollback_snap() in favour of re-using the newer create-repclone(). This helps to remove the number of areas within code that a subvol mount is enacted, whilst simultaneously addressing a quota update issue associated with rollback_snap()'s caller in share_command.py

- Includes re-use of create-repclone() code for share rollback, replacing the prior, and older, rollback_snap().
- Above helps to reduce the number of disparate locations a mount is enacted by the removal of the older rollback_snap.
- Removes older rollback_snap()'s caller's incorrect use of update_quota().
- Reduces number of low level calls in higher level management code.
- Adds replication abort prior to share delete if no on-disk snap found: taken from the prior rollback_snap() code.
- Move informative snap-to-share message from debug to info.
- Passes qgroup (and hence subvol id) from upper level management code to lower level mount code to avoid redundant subvolid system queries.
- Updates test_share_commands.py unit test mocks accordingly.

Fixes #1923 
and also:
Fixes #1880 
by way of problem code removal.

@schakrava Ready for review.

Testing:
The unit test modifications were tested via:
```
./bin/test -v 2 -p test_share_commands*
```

test_clone_command (storageadmin.tests.test_share_commands.ShareCommandTests) ... ok
test_rollback_command (storageadmin.tests.test_share_commands.ShareCommandTests) ... ok

Manual share/snap creation with the various available options was as expected along with the promotion of a snap to share via the clone mechanism. All expected mount states / behaviours were as expected and unchanged from pre pr behaviour.

A greater than 5 cycle replication was also tested in both directions using fresh builds of master with this pr applied to test both send and receive ends.

N.B. we have an existing bug re share rollback where prior snapshots are lost to the UI. This has yet to be issued but is unaffected / identically reproduced by this pr's changes to the rollback code. I have a reproducer sequence and will issue accordingly soon.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -19,7 +19,6 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 import re
 import time
 import os
-import shutil
 from system.osi import run_command, create_tmp_dir, is_share_mounted, \
     is_mounted, get_dev_byid_name, convert_to_kib, toggle_path_rw, \
     get_device_path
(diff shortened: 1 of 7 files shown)


schakrava:
Thank you @phillxnet Nicely done!

phillxnet:
@schakrava Thanks.

PR #1928, opened 2018-05-24
remove --stdin option use for passwd. Fixes #1927

### LINKED ISSUE #1927: remove --stdin option use for passwd
Some more progressive implementations of passwd have the --stdin option deprecated/removed, this is to avoid the passing of plain text passwords via standard in. Accommodate for this implementation detail ahead of it's effects.

It is proposed that we instead make use of the python crypt module to pre-encrypt an entered system user's password and pass this via the -p option to usermod: circumventing the use of 'passwd --stdin'.

phillxnet:
By substituting 'usermod -p' for 'passwd --stdin' when setting user pass
we avoid potential future issues on passwd's deprication / removal of
the --stdin option: as has been observed in more defensive / progressive
implementations of passwd.

Elements of commit/pr:

- Drop 'passwd --stdin' in favour of 'usermod -p'.
- Use crypt module for password pre-encrypt as required for above change, ensuring sha-512 (observed current system default).
- Use random full length (16 chars) salt in above.
- Add comments for post python3 move enhancements re crypt's use.

Fixes #1927 

@schakrava Ready for review.

Testing:

- Post pr created users were able to login via  local console.
- Visual inspection of /etc/shadow confirmed $6$ (sha-512) encryption along with 16 char (max) salt length.
- A fresh build of current master with this pr applied was able to successfully register the initial admin user as expected, additional users (both without and with web admin ability) were created and worked as expected.

### DIFF
--- a/src/rockstor/system/users.py
+++ b/src/rockstor/system/users.py
@@ -28,6 +28,9 @@ import grp
 from shutil import move
 from tempfile import mkstemp
 import chardet
+import random
+import string
+import crypt
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


phillxnet:
@schakrava Thanks.

Issue #1656, opened 2017-02-23
[BUGS] Default ulimit setting too low


PhilippeAB:
Hi,
For production environment with about 20 shares with snapshot we ran into "too many open files"
problem.
Snapshot, shadow copy and exportfs fail then the whole webgui crash with internal server error 500

Created a file in /etc/security/limits.d/
cat 30-nofile.conf 
root	   soft    nofile    8192
root       hard    nofile    16364

(default is 1024)

PhilippeAB:
With systemd the configuration through security/limits don't work.
You need to update de rockstor.service with the needed file limits

[Service]
LimitNOFILE=10000

PhilippeAB:
The php frontend crash even with 10000 files. It just take longer.
There is a FD leak on /tmp/exports
lsof give this
gunicorn   13354 13367                     root  615u      REG               0,40          592    9875127 /tmp/exports (deleted)

3700+ files after few days

phillxnet:
Linking to relevant forum threads where forum members are experiencing this same issue:
nfriedly in https://forum.rockstor.com/t/web-ui-errors-mostly-too-many-files-open-while-runing-a-balance/3689
and:
peter in https://forum.rockstor.com/t/error-24-too-many-open-files-causes-scheduled-scrub-job-to-fail-and-prevents-login-to-web-ui/3649
Please update these forum threads with this issues resolution.

nfriedly:
Yea, I ran into this during a rebalancing; (https://forum.rockstor.com/t/web-ui-errors-too-many-open-files-while-runing-a-balance/3689/2), here's a couple of comments I can add to follow up on [my forum thread]:

* I have a number of rockons, but I turned off the rockons service before I noticed much of this issue
* I have a scheduled scrub, but it did not occur during the balance (so far)
* I have scheduled nightly, weekly, and monthly snapshots of my main share (~4TB of data). Since starting the balance it's attempted 3 nightly and 1 weekly snapshot, all have failed.
* Grabbing files over SMB, or even just navigating folders, frequently failed. It would often time out several times, and then go very quickly when it did eventually work. 
* I installed `lsof` today, and running `lsof | wc -l` right now reports 10925.  I'll see if I can come up with some bash foo to narrow down which commands make up the bulk of thos| 

[...]

PR #1934, opened 2018-06-08
Possible fix for https://github.com/rockstor/rockstor-core/issues/1656 ("Too many open files" / umlimit issue)

ericrisler:
Typical problem looks like this:
https://forum.rockstor.com/t/exception-while-running-command-usr-bin-hostnamectl-static-errno-24-too-many-open-files/2272/5
...
fix inspired by this post: http://carsonip.me/posts/gevent-pywsgi-http-keep-alive-fd-leak

NOTE: this fix needs review as I don't know if this will cause issues elsewhere. I've update my local rockstor and left the WebUI open and open files from gunicorn remains stable at < 1000 open files.

### DIFF
--- a/src/rockstor/cli/api_wrapper.py
+++ b/src/rockstor/cli/api_wrapper.py
@@ -34,7 +34,7 @@ class APIWrapper(object):
         self.client_secret = client_secret
         # directly connect to gunicorn, bypassing nginx as we are on the same
         # host.
-        self.url = 'http://127.0.0.1:8000'
+        self.url = 'http://127.0.0.1:443'
         if (url is not None):
             # for remote urls.
             self.url = url


schakrava:

schakrava:

MFlyer:
Hi @phillxnet , pitchining in on this: @ericrisler why moving from tcp socket 8000 to tcp socket 443 should solve this?!? (your PR doing only this, maybe you missed something / had wrong PR) ?!?

By the way, Rockstor gevent part has been replaced using python.socketio instead of gevent, so we probably not using gevent anymore

M.

P.S.: @phillxnet & @schakrava , missing you all, getting mad with code refactoring on work projects :(

ericrisler:
@MFlyer - that change doesn't move the service to another port...all I did was change the port that the API calls go through from 8000 (direct to gunicorn) to 443 (via nginx). Reading these threads...

benoitc/gunicorn#1428
benoitc/gunicorn#1375
benoitc/gunicorn#1327
https://bugs.python.org/issue10099

....hints that py wsgi can keep file descriptors open and not close them causing a file descriptor leak. I then found this post http://carsonip.me/posts/gevent-pywsgi-http-keep-alive-fd-leak that says if you use a reverse proxy such a nginx it'll take care of the http connection in such a way that pywsgi will properly close the FDs. :)

As I said, it probably should be tested more and reviewed by Rockstor developers to make sure it doesn't break something else. Assuming that gunicorn is there to server the webui only I figure this change will be ok.

As for gevent not being in use: perhaps there is some other underlying issue that affected the users in those posts and Rockstor that is not specific to gevent? 

My PR is probably better labelled as a workaround instead of a fix.

ericrisler:
Hey @MFlyer - I'm certainly not arguing with you on your points - I saw all that before I made the change. The api_wrapper.py has a comment "# directly connect to gunicorn, bypassing nginx as we are on the same host" so I assume that this wrapper is used internally someplace to make API calls directly to gunicorn and not through nginx. I'm sure you know this of course - the change I make forces all api calls to nginx, which proxy-passes to gunicorn as you stated above. 

Perhaps this shouldn't be a pull request but a discussion elsewhere? I just did it this way to draw attention to the point that this fixes the issue locally for me and may for others but you're the pro on this project not me so I'll defer to your expertise. Hopefully this discussion will lead to the real underlying cause of the issue.

This change certainly solves the issue on my local rockstor machine. If I change the line back to port 8000, restart the rockstor service and open the webui then watch the open files I get a steady growing count form gunicorn. Listing them shows repeated open file descriptions to "/tmp/afp.conf (deleted)". This grows and grows until the error. Changing the port to 443 and hence forcing the calls through nginx and watch the open files the symptoms are gone.

Your point on the fact that nginx has a default timeout may be a factor that causes problems if there are long running gunicorn tasks that don't respond to nginx.

MFlyer:
@ericrisler : I believe you and your tests, so if we have gunicorn directly exposed that's not expected Rockstor bahaviour and you've got a nice catch! I'll find time to catch it, that's a promise! ;)

M.

ericrisler:
Cool - it's really a great system, it's easy to setup, configure and manage. Thanks for all the efforts!

phillxnet:
Linking back for context and to make it easier to find this pr from our main issue for what this pr is intended to address:
"[BUGS] Default ulimit setting too low" #1656

@ericrisler and @MFlyer any more progress on where we are to go with this one?

@ericrisler This pr has not been forgotten by the way. I'm just in 2 minds as to it's appropriateness. Hopefully we will get more input / validation and I try and find the time to look into this more myself when I can. Thanks for sharing your fix by the way, much appreciated.

phillxnet:
Closing as a general tidy-up of now old pull requests and the suspected initiator of this issue's root cause has now been removed entirely as part of our removing AFP and all associated code:
https://github.com/rockstor/rockstor-core/releases/tag/3.9.2-56
released on 1st April 2020.

@ericrisler Apologies for never enacting this but we've not had a similar report since our above code removal.

Please take a look at our more recent v4 "Built on openSUSE" offering and see if you still experience this. We also now have a DIY or pre-build installer option.

PR #1939, opened 2018-06-27
Fix Rockon unknown state due to multi-line docker inspect

### LINKED ISSUE #1933: Rockon state cannot be determined
Hi team,

When running some rockons I get "unknown_error" as state. This is due to the docker inspect returning more than one line (e.g. if you have Health status) so the State fields are not parsed:
https://github.com/rockstor/rockstor-core/blob/907f5c2b28e8307c75154d56d0a10bec1ec98933/src/rockstor/storageadmin/views/rockon_utils.py#L30-L33

Can I open a PR to change the query to be "Error:{{.State.Error}},ExitCode:{{.State.ExitCode}}.Running:{{.State.Running}}"? These three fields are the only ones we care about in the code and changing the filter won't required more modifications, plus remains flexible if we want to add more fields in the future.

sgissi:
Fixes #1933. The proposal is to only gather the 3 fields we care about (Running, Error and ExitCode). Could be further improved by leaving the labels out and arbitrarily order the fields but would be less flexible.

Signed-off-by: Silvio Gissi <silvio@gissilabs.com>

### DIFF
--- a/src/rockstor/storageadmin/views/rockon_utils.py
+++ b/src/rockstor/storageadmin/views/rockon_utils.py
@@ -28,7 +28,7 @@ def container_status(name):
     state = 'unknown_error'
     try:
         o, e, rc = run_command([DOCKER, 'inspect', '-f',
-                                '{{range $key, $value := .State}}{{$key}}:{{$value}},{{ end }}', name])  # noqa E501
+                                'Error:{{.State.Error}},ExitCode:{{.State.ExitCode}},Running:{{.State.Running}}', name])  # noqa E501
         state_d = {}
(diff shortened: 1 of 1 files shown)


schakrava:

schakrava:

sgissi:
Friendly ping @MFlyer @schakrava.

This fixes a problem with containers that report Health status (e.g. Unifi Rockon) showing unknown state and not allowing the user click the button to reach the container UI.

schakrava:
Nice fix @sgissi. Thanks for your contribution!

PR #1943, opened 2018-07-01
Added security headers

hashtaginfosec:
See rockstor#1940 for more info. While the issue only states cache-control, I added following security headers in total to improve client/browser side security:

-  Cache-Control no-store;
-  X-XSS-Protection "1; mode=block";
-  X-Frame-Options SAMEORIGIN;

Tested locally and this has not affected functionality.

Thanks @phillxnet for your encouragement.

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -41,6 +41,10 @@ http {
 	keepalive_timeout	75 20;
 	ignore_invalid_headers	on;
 	index index.html;
+	
+	add_header Cache-Control no-store;
+	add_header X-XSS-Protection "1; mode=block";
+	add_header X-Frame-Options SAMEORIGIN;
 
 	server {
 	       listen 80;
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -41,10 +41,6 @@ http {
 	keepalive_timeout	75 20;
 	ignore_invalid_headers	on;
 	index index.html;
-	
-	add_header Cache-Control no-store;
-	add_header X-XSS-Protection "1; mode=block";
-	add_header X-Frame-Options SAMEORIGIN;
 
 	server {
 	       listen 80;


schakrava:

schakrava:

phillxnet:
@hashtaginfosec First off, well done on finding the options and getting to the pr stage.

A couple of comments.

If I understand correctly the file you have changed is not actually used anymore, we should clear this out if so. Take a look at it's git history:
https://github.com/rockstor/rockstor-core/commits/master/conf/nginx.conf
where as my understanding is that the 'source' nginx template file we use during a build is:
https://github.com/rockstor/rockstor-core/commits/master/conf/nginx.conf.in
where we see @MFlyer has been the latest contributor.

@MFlyer if you have time could you confirm my understanding here, thanks.

See my comment earlier on the referenced issue on how this template nginx.conf.in file is used to generate the actual used file of for example "/opt/rockstor-dev/etc/nginx/nginx.conf" where /opt/rockstor-dev is used as the build dir, usually /opt/rockstor for an rpm install. Take a look at the contents of your local rpm installed /opt/rockstor/etc/nginx/nginx.conf.

Second minor issue is that as per the "Developers" doc entry:
http://rockstor.com/docs/contribute.html#developers
it is much preferred, and in the end helps all around, if you create a specifically named branch for each issue and do your pull request against that branch. Otherwise if you try and work on 2 issues as you appear now to have the solutions for (hitting the ground running there I see) you can keep their respective code bases divided / separated. Otherwise, if your pr is from your master branch, as in this pr, then it is far easier to accidentally change that branch mid pull request review and end up having unintended code associated with one or more pull requests.

Hope that helps and it would be great to have these changes in so please do persist with this potentially tedious nit picky element of the submission as then we can have these changes properly attributed, and your hopefully future pr's should be easier also. If need be you can just close this pr and open a new one against the new branch and file changes.

Oh and apologies if I've gotten the wrong end of the stick here. @MFlyer should put us straight on the above I suspect.

hashtaginfosec:
Thanks. I'll get on it tomorrow.

phillxnet:
@hashtaginfosec I'm closing this pr given our previous discussion re issue specific branches and that it now represents no change. Feel free to open a new pr as per the developer documentation and with our prior discussion here in mind if you fancy. Please don't be discouraged, this is just part of getting our ducks in a row with each pr to help us with maintenance and code review as we go along.

Issue #1342, opened 2016-06-09
Importing an unlabeled btrfs RAID5 pool fails


AngeleToR:
Yesterday I found a minor bug when trying to import a btrfs RAID5 5 disk array into my new system. Another similar array with a label in it worked flawlessly, but the unlabeled one got succesfully imported, mounted and even displayed for a seconds, then dissapeared from the GUI but remained mounted under /mnt2/none (none is how it got temporaly labeled).

Just going to a shell, manually unmounting it, setting a label and re-importing from the GUI did the trick and made it work. As I said, minor annoyance and we all should label our filesystems after all, don't we??? 😆 

It happened on latest stable, 3.8.13 fresh install

phillxnet:
@AngeleToR This is a great find; thanks for reporting.
Linking to your forum thread to keep the two tied together.

Please update the following forum thread (by the same author) with any significant development on this issue:- https://forum.rockstor.com/t/importing-an-unlabeled-btrfs-raid5-volume-pool-fails/1590

This issue is caused by the import mechanism doing it's job as expected, the momentary appearance of the pool as evidence, but because the background periodic _update_disk_state() mechanism depends on labels to do it's thing it then fails to deal with the 'None' label condition.

My thoughts on this are that we have some pending changes in this area so I think it would be best to get these out of the way first before looking to this issue, especially now you have provided the work around (Thanks) but maybe after the commits associated with #1320 have been finalised, reviewed, and committed (if appropriate), we could fix this by enhancing the import mechanism to either auto label a pool if it is found to be un-labelled or prompt the use for a label when this circumstance is found. Just my quick thoughts on the matter.

AngeleToR:
Sorry for not updating the thread, I though I already did... Anyway, as Phillxnet just pointed out, naming the pool just fixed the issue. Tha fail came from a previous Ubuntu setup, where I could create and use the array unnamed and so I tried to import it that way... triggering the error...

phillxnet:
@AngeleToR Thanks for the update, but I think this issue could be re-opened as it still exists, I was simply referencing your fine find on the work around you presented here. Not strictly an issue with intra-Rockstor pools and a niggle as you say but still it's not elegant behaviour and imports of whole disk btrfs should fall within our remit. Your call.

AngeleToR:
[...]

PR #1949, opened 2018-08-05
add system pool auto label and enhance associated db mechanisms. Fixes #1947

### LINKED ISSUE #1947: add system pool auto label and enhance associated db mechanisms
It is proposed that we elegantly deal with unlabelled system pools, and as we know in advance their purpose we can deploy the default label of 'system' in this scenario. This is further to the enhancements introduced in:
"pool name should default to uuid if label is none. Fixes #1342" pr #1683 which addressed non system pools only (due to requiring no existing mount).

It is also proposed that we normalize our device identification of existing pool affiliation via 'btrfs fi show'; ie via the existing get_pool_info() wrapper which is currently resourced for this purpose by:
_refresh_pool_state()
and
_btrfs_disk_import()
however _update_disk_state(), for the same purpose, uses the disks labels only. This is inconsistent and non canonical as directly after a btrfs vol / pool label change the associated disks labels are not updated. This results in a non tenable db entry of ‘none’ for the special case of system pool creation that takes place within _update_disk_state() only and ends up blocking all future db actions and updates.

phillxnet:
Normalise use of 'btrfs fi show' for label info and when no label is found, initiate an auto label; given the existence of a by-id (device serial). If no serial and no label on the system drive / pool then skip system pool / vol db counterpart creation. Also adds better logging and earlier fail points concerning no by-id (no serial) by avoiding subsequent mount attempts of associated vols and subvols.

Summary:

- Make ‘btrfs fi show’ via get_pool_info() canonical for volume label retrieval, retaining a fail over to disk label for robustness.
- Extend existing auto volume (pool) labeling, within get_pool_info(), to account specifically for the system pool: see previous issue #1342, pr #1683, for non system disk.
- Improve system volume database management code to be more robust to no system volume label scenarios; including in concert with no by-id name being available (ie no serial on system disk).
- Add debug logging re system pool database counterpart creation or otherwise (ie no by-id in combination with no serial).
- Dynamically retrieve and assign / update system volume (pool) raid; where a by-id name is available.
- Move previous multiple hardcoded ‘system’ label references to settings.conf.in / test-settings.conf.in, see caveats section of pr #1935.
- Avoid all pool mounts when the referenced disk has no by-id name, and error log accordingly: avoiding later complex fail messages where downstream code requires by-id naming.
- Avoid share import / status update when the associated pool is not mounted, error log accordingly.
- Avoid mount attempts of prior known shares whose associated pool is not mounted, error log accordingly.

Fixes #1947
Please see issue text for further context.

@schakrava Ready for review (really this time - hopefully).

Note that the default ‘system’ label was chosen to differentiate between legacy ‘rockstor_rockstor’ labelled pools and non legacy system pools. It is not advised that we maintain the rockstor_rockstor system label when we relabel as then there is no outward indication that this system has been relabelled and all testing has concentrated on a separation between legacy (existing label of ‘rockstor_rockstor’) and proposed new system bases which by default have no existing system vol / pool label and very different vol / subvol layouts.

Testing:
Legacy, and non legacy systems were tested, in the latter case with and without snapper root config. All configs were confirmed to work as expected with either no system drive label only or in combination with no system drive serial, where with the latter it was also confirmed that post serial addition and reboot, the system pool was successfully labelled and surfaced. All configs were also tested for their ability to re-label the system drive were it to have it’s existing label removed via:
btrfs fi label / ''

Unit test results (no additional test added):
./bin/test --settings=test-settings -v 3 -p test_btrfs*
All ok
and
./bin/test  --settings=test-settings -v 3 -p test_osi*
All ok

Caveats:
In this pr we introduce additional double conversions of device names, ie from temp (canonical) type of sda to by-id and then back again. This is inelegant and slow (though not significantly) and could be addressed in future refactoring but is left for the time being to help reduce the scope of these code changes.

### DIFF
--- a/conf/settings.conf.in
+++ b/conf/settings.conf.in
@@ -276,6 +276,10 @@ MNT_PT = '/mnt2/'
 NFS_EXPORT_ROOT = '/export/'
 SFTP_MNT_ROOT = '/mnt3/'
 
+# System volume label when no btrfs volume label is set as per default openSUSE
+# install: ie 'btrfs fi show' gives 'Label: none'
+SYS_VOL_LABEL = 'system'
+
 TAP_DIR = '${django-settings-conf:taplib}'
 TAP_SERVER = ('127.0.0.1', ${django-settings-conf:tapport})
 MAX_TAP_WORKERS = 10
--- a/conf/test-settings.conf.in
+++ b/conf/test-settings.conf.in
@@ -263,6 +263,10 @@ MNT_PT = '/mnt2/'
 NFS_EXPORT_ROOT = '/export/'
 SFTP_MNT_ROOT = '/mnt3/'
 
+# System volume label when no btrfs volume label is set as per default openSUSE
+# install: ie 'btrfs fi show' gives 'Label: none'
+SYS_VOL_LABEL = 'system'
+
 TAP_DIR = '${django-settings-conf:taplib}'
 TAP_SERVER = ('127.0.0.1', ${django-settings-conf:tapport})
 MAX_TAP_WORKERS = 10
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -149,7 +149,7 @@ def degraded_pools_found():
     return degraded_pool_count
 
 
-def get_pool_info(disk):
+def get_pool_info(disk, root_pool=False):
     """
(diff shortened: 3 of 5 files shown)


schakrava:
Thank you @phillxnet !

phillxnet:
@schakrava Cheers.

PR #1946, opened 2018-07-11
disk serial=none or fake-serial re sda[a-z] dev names. Fixes #1925

phillxnet:
On systems with 27 or more disks named sd* where the system/root is also installed on sda ie sda3 and where the 27th and subsequent disks are named sdaa, sdab etc and are also btrfs formatted; a regex based serial propagation bug (from base device to partition) in scan_disks() resulted in the first listed (by lsblk) sda[a-z] device receiving a serial=none attribution and the second and all subsequent listed sda[a-z] devices receiving an erroneous fake-serial attribution. All affected devices would also have empty (or incorrectly inherited from sda) model, transport, vendor, and hctl attributions. This resulted in an inability for the system to update the disk state as the abstracted disk info from the corrupt parsing of attached disk state caused an exception: "TypeError: expected string or buffer".

Note that all affected disks (sda[a-z]) were also miss attributed as root=True in the erroneous abstract data produced by scan_disks()

Summary

1. Add scan_disks() test mode to enable repeatable fake-serial- tests.
2. Establish unit tests to reproduce the above issue, as reported (36 disk system) and as a minimum artificial drive subset.
3. Establish unit tests to identify current correct behaviour for system disk on LUKS, mdraid, and bios mdraid; as well as btrfs in partition and data disk on bios mdraid behaviour.
4. Fix overly broad regex to identify system disk partitions, if any: ie improve system disk partition match mechanism.
5. Fix system disk LUKS, mdraid, bios mdraid regressions as a result of (4).
6. Address pre-existing TODO: re mdraid info via model info on non system disks; as part of the fix for (5).

Fixes #1925 
and, by way of duplication:
Fixes #1834 ("disk serial is null")

@schakrava Ready for review.

Testing:
The reported issue was reproduced via unit testing based on the reporter's 36 disk system. An abstract of the minimum drive count thought to be able to reproduce the issue was also produced. Further unit tests were constructed for regression testing during the fix development.
```
./bin/test --settings=test-settings -v 3 -p test_osi*
```
with the relevant scan_disks() related results, post pr, as follows:
```
...
test_scan_disks_27_plus_disks_regression_issue (system.tests.test_osi.OSITests) ... ok
test_scan_disks_btrfs_in_partition (system.tests.test_osi.OSITests) ... ok
test_scan_disks_dell_perk_h710_md1220_36_disks (system.tests.test_osi.OSITests) ... ok
test_scan_disks_intel_bios_raid_data_disk (system.tests.test_osi.OSITests) ... ok
[...]

PR #1952, opened 2018-08-14
regression - nvme system disk not identified. Fixes #1951

### LINKED ISSUE #1951: regression - nvme system disk not identified
Thanks to forum members Jorma_Tuomainen and Flox in the following thread for highlighting this regression. As from 3.9.2-31 pr #1946 nvme system disks were no longer identified as attached: resulting in, post update, the system disk showing as detached and consequently it's default surfaced share 'home', and any created post install, would fail to be available.

Associated log entry:
```
ERROR [storageadmin.views.command:70] Skipping Pool (rockstor_rockstor00) mount as there are no attached devices. Moving on.
```
Where the Pool name may vary from install to install.

Non system pools / shares were not affected.

https://forum.rockstor.com/t/cant-start-rock-ons/5065

phillxnet:
Address nvme system disk regression, not recognised as attached, by improving the partition regex mechanism initially enhanced in commit acbd6a5 (pr #1946) which created the indicated regression.

Summary:

- Unit test added to instantiate observed nvme system disk regression and expected behaviour.
- Simplify regex match conditional, for root partition identification, so that it again incorporates nvme device names: with root in partition.
- Establish conditional partition regex prior to above match.
- Enhance existing btrfs-in-partition unit test to catch a regression observed prior to establishing the above conditional regex that affected the 42nd scsi type drive 'sdap' if that drive had a btrfs partition concurrent with an 'sda' root drive name (and the prior regex's were 'or' combined).
- Add debug logging of sys partition info inheritance from base device.
- Add debug logging of btrfs in partition device skip: we only surface the base device.
- Minor cosmetic bracket changes and comment updates.
- Added TODO: on future option re base device identification from partition device name.

Fixes #1951 
Please see issue text for further context.

@schakrava Ready for review.

Testing:
All existing osi and btrfs tests return OK:
extended / modified tests:
test_scan_disks_btrfs_in_partition (system.tests.test_osi.OSITests) ... ok
new tests:
test_scan_disks_nvme_sys_disk (system.tests.test_osi.OSITests) ... ok

Caveats:
This pr and  #1946 introduce a white list approach to dev name regex re system partition identification but this was deemed necessary given the related issues observed. Ideally in time we can remove the 'special treatment' required for the system disk by moving it's associated mechanisms over to using the redirect role which is currently only associated with non system disk btrfs in partition disks. This move should help to simplify the code and normalise the partitions element of the Disk collection across all drives, ie move to surfacing the system disk as per non system disks: via the base device.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -365,7 +365,7 @@ def scan_disks(min_size, test_mode=False):
             # This dict will be populated when we find our partitions and back
             # port their names and fstype (as values).
         if ((not is_root_disk and not is_partition) or
-                (is_btrfs)):
+                is_btrfs):
             # We have a non system disk that is not a partition
             # or
             # We have a device that is btrfs formatted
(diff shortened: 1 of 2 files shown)


schakrava:
Thanks for resolving this @phillxnet

phillxnet:
@schakrava Cheers. It's a shame I let it through in the first place. Oh well at least we have a regression test in place going forward.

PR #1956, opened 2018-08-25
remove grubby dependency for kernel version reporting. Fixes #1955

### LINKED ISSUE #1955: remove grubby dependency for kernel version reporting
On systems where grubby is not installed, the existing mechanism fails over to not displaying the kernel version in the Web-UI header. Modify this behaviour so that in this instance we circumvent the failure point, that of attempting to re-assert a recognised kernel (via a non existent grubby), and simply return the already retrieved uname kernel version element.

This caters for non legacy systems where we no longer have to enforce a btrfs appropriate kernel, and as such it is suggested that we move to relying on our upstream maintainers (non legacy) to maintain an appropriate kernel. We can therefore, going forward, drop the kernel version check entirely. This issue is the initial step to maintain backward compatibility while in transition.

phillxnet:
Ensure that grubby exists prior to embarking on it's use. This circumvents an otherwise unintended failure to return the running kernel version for Web-UI display. For legacy systems (grubby pre-installed) this should not represent a functional change.

Fixes #1955 
Please see issue text for context.

@schakrava Ready for review.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -517,7 +517,7 @@ def def_kernel():
 
 def kernel_info(supported_version):
     uname = os.uname()
-    if (uname[2] != supported_version):
+    if uname[2] != supported_version and os.path.isfile(GRUBBY):
         e_msg = ('You are running an unsupported kernel(%s). Some features '
                  'may not work properly.' % uname[2])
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @phillxnet

Issue #1722, opened 2017-06-01
pool resize disk removal unknown internal error and no UI counterpart


phillxnet:
Thanks to forum member Noggin for highlighting this behaviour. Occasionally when removing a disk from a pool there can be a UI time out directly after the last dialog entitled "Resize Pool / Change RAID level for ..." which acts as last confirmation of the configured operation:

![harmless-put-timeout-on-dev-remove](https://cloud.githubusercontent.com/assets/2521585/26693598/a6a7f03a-46fc-11e7-80d2-f45df1f32e28.png)

There is then no UI 'balance' indicated while the removal is in progress, yet the UI indicates that a balance is in progress when a balance is attempted (only attempted by Noggin as I did not attempt to execute a balance whilst the removal was in progress).

```
btrfs balance status /mnt2/time_machine_pool/
No balance found on '/mnt2/time_machine_pool/'
```

The pool resize is however indicated by the requested disk's having their size 'demoted' to zero and showing a reduced usage with subsequent executions of **btrfs fi show**:

```
Label: 'time_machine_pool'  uuid: 8f363c7d-2546-4655-b81b-744e06336b07
	Total devices 4 FS bytes used 31.57GiB
	devid    3 size 149.05GiB used 17.03GiB path /dev/sdd
	devid    4 size 0.00B used 5.00GiB path /dev/sda
	devid    5 size 149.05GiB used 23.03GiB path /dev/mapper/luks-d36d39ea-c0b3-4355-b0c5-bd3248e6bbfe
	devid    6 size 149.05GiB used 23.00GiB path /dev/mapper/luks-d7524e90-4d9e-4772-932f-d1407b6b5fe7
```

and then later on:
```
Label: 'time_machine_pool'  uuid: 8f363c7d-2546-4655-b81b-744e06336b07
	Total devices 4 FS bytes used 32.57GiB
	devid    3 size 149.05GiB used 18.03GiB path /dev/sdd
	devid    4 size 0.00B used 2.00GiB path /dev/sda
	devid    5 size 149.05GiB used 24.03GiB path /dev/mapper/luks-d36d39ea-c0b3-4355-b0c5-bd3248e6bbfe
	devid    6 size 149.05GiB used 24.00GiB path /dev/mapper/luks-d7524e90-4d9e-4772-932f-d1407b6b5fe7
```
[...]

PR #1971, opened 2018-10-02
Implement a delete missing disk in pool UI #1700

### LINKED ISSUE #737: As a user I'd like to replace failed drives in a RAID with new ones
For some reason unknown to me there does not seem to be an obvious way to replace a failed drive which is part of a RAID pool by a new one and resync the pool. 

Steps to reproduce: 

1) Set up a RAID-1 pool
2) Shut down the system
3) Pull a drive
4) Reboot the system
5) Have a look around and discover that there's no evidence of a degraded RAID or any possibility to remove the failed drive or add a new one in its place

### LINKED ISSUE #1199: Rockstor does not handle drive/pool errors with Grace or Recognition
I Did another break fix test.
vmware 5.5, unique UUID for vmdk's.

I shut down Rockstor and removed one of the drives in vmware, then booted it back up. When viewing the near empty raid5 pool I now see 89.37% usage and get an error:

```
Error! 
Error running a command. cmd = ['/bin/mount', u'/dev/disk/by-label/test2', u'/mnt2/test2', '-o', u',compress=no']. rc = 32. stdout = ['']. stderr = ['mount: wrong fs type, bad option, bad superblock on /dev/sdd,', ' missing codepage or helper program, or other error', '', ' In some cases useful info is found in syslog - try', ' dmesg | tail or so.', ''] 
If you need more help on this issue, email us at support@rockstor.com with the following information: 
• The error message: "Error running a command. cmd = ['/bin/mount', u'/dev/disk/by-label/test2', u'/mnt2/test2', '-o', u',compress=no']. rc = 32. stdout = ['']. stderr = ['mount: wrong fs type, bad option, bad superblock on /dev/sdd,', ' missing codepage or helper program, or other error', '', ' In some cases useful info is found in syslog - try', ' dmesg | tail or so.', '']"
• A step by step description of the actions leading up to the error.
• Download the tarred and zipped server logs by clicking here, and attach it to the email.

```

I can’t click into the degraded pool at all without that error happening so I have no way to view/repair/investigate/research. I see no degraded indicator, red highlight, or other reason looking at pools or share lists to think I might be having issues.

The removed drive even still shows up in the Disks list. I can click error logs/self-test logs and it shows “there are no errors”! Shouldn't it at least say the drive is missing or present some kind of error?

I see no other external indicator to tell me that a drive dropped out/failed. That kind of problem should be jumping out at me on the dashboard. I should probably get a pop up requiring acknowledgement before continuing. Console output is loaded with errors, but nothing in Rockstor webUI.

Rockstor should probably be actively polling drive status to report on these kinda of things.

### LINKED ISSUE #1700: Implement a delete missing disk in pool UI

phillxnet:
Extend the existing pool details page and it's Resize/ReRaid wizard to facilitate the removal of detached / missing disks. Includes a number of bug fixes re Pool edits, mostly concerned with disk add / remove when a pool has one or more detached members. Advise and link on Disks page against detached members of managed pools. Note that detached disks with no managed pool association are unaffected by these changes. I.e. they can be, as before, removed (from db) via a bin icon.

Summary:
- Ensure pool is mounted prior to add / remove / raid change actions and present a user error if this prerequisite is not met.
- Add 'btrfs device delete missing' function to managed pool detached member via the existing 'Remove disks' option within Pool details page 'Resize/ReRaid Pool'.
- Web-UI Disk's page - add map icon to identify "... detached member of a Rockstor managed Pool." tooltip quote.
- Web-UI Disk's page - add exclamation mark "Use the associated Pool page Resize/ReRaid 'Remove disks' option if it is not to be reattached." tooltip quote.
- Web-UI Pool details page - add contextually aware text / link giving advice and optional 'dev delete missing' work around for when no detached disks are known: i.e importing a degraded pool on a fresh
install for disaster recovery.
- Change remaining / resulting disk count validation to account for detached disks.
- Remove redundant balance after disk delete (pool member removal) operation.
- Change disk model temp_name property to return canonical variant of target_name, ie make it redirect role aware.
- Add pool details page Disk table "Temp Name" column (first and currently only use of temp_name disk model property).
- Fix bug where io_error_stats failed for btrfs-in-partition (ie redirect role).
- Add "Replace existing disk (planned option)"" placeholder Web-UI text.
- Minor rewording of user facing messages re pool edits.
- Minor variable name refactoring / renaming.
- Manage disk selection within Resize/ReRaid wizard and consequent validation via disk id not disk name, fixes buggy behaviour re detached disk management.
- Add TODO: re running resize_pool() as async task, relevant mostly to disk delete as a time out is generated on all but trivially small pools.
- Add TODO: pool_info() - move 'disks' to dictionary with (devid, size, used) tuple value.
- Add TODO: _update_disk_state() - how to fix a recent btrfs in partition regression (my fault).

Fixes #1700 
See issue comments for context and images of added Web-UI elements.
And as the last remaining sub-issue linked in the following super set issues:
Fixes #737
(6 other dependant and now closed issues).
Fixes #1199 
(5 other dependant and now closed issues).

@schakrava Ready for Review

Testing:
All affected pool edit options were tested to behave as expected (given known caveats below) on both KVM systems and real hardware.

Caveats:
- When removing a drive live "btrfs fi show" and consequently Rockstor indicate detached / missing pool members. But prior to a reboot any attempt to 'btrfs device delete missing' results in a '... no missing devices..' type error which we do surface. Post reboot delete missing works as intended.
- As "btrfs device delete (missing or otherwise)" initiates an internal balance that can take several hours we experience a time out that break the Web-UI feedback and associated db updates for this operation. However this is a related but independent issue which is detailed in existing issue:
"pool resize disk removal unknown internal error and no UI counterpart" #1722
The work around for the time being is to execute the same disk removal operation again where the second attempt (if post internal balance completion) executes as expected due to the initial disk removal having now succeeded and so is skipped, avoiding the cause of the timeout that breaks the otherwise functional (on trivially small pools) btrfs device delete (missing or otherwise) pool operation.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -230,6 +230,7 @@ def get_pool_info(disk, root_pool=False):
     o, e, rc = run_command(cmd)
     pool_info = {'disks': [], 'hasMissingDev': False, 'fullDevCount': 0,
                  'missingDevCount': 0}
+    # TODO: Move 'disks' to dictionary with (devid, size, used) tuple values
     full_dev_count = 0  # Number of devices in non degraded state.
     attached_dev_count = 0  # Number of currently attached devices.
     for l in o:
@@ -321,7 +322,7 @@ def resize_pool(pool, dev_list_byid, add=True):
     (when adding) or are already members of the pool (when deleting).
     If any device in the supplied dev_list fails this test then no command is
     executed and None is returned.
-    :param pool: btrfs pool name
+    :param pool: btrfs pool object
     :param dev_list_byid: list of devices to add/delete in by-id (without
         path).
     :param add: when true (default) or not specified then attempt to add
@@ -330,7 +331,17 @@ def resize_pool(pool, dev_list_byid, add=True):
     :return: Tuple of results from run_command(generated command) or None if
         the device member/pool sanity check fails.
     """
-    dev_list_byid = [get_device_path(d) for d in dev_list_byid]
+    if pool.has_missing_dev and not add:
+        if dev_list_byid == []:
+            dev_list_byid = ['missing']
+        else:
+            # list has at least a single element
+            # substiture 'missing' for any member matching 'detached-'
+            dev_list_byid = [
+                'missing' if re.match('detached-', dev) is not None else
+                get_device_path(dev) for dev in dev_list_byid]
+    else:
+        dev_list_byid = [get_device_path(dev) for dev in dev_list_byid]
     root_mnt_pt = mount_root(pool)
     cur_dev = cur_devices(root_mnt_pt)
     resize_flag = 'add'
@@ -338,15 +349,17 @@ def resize_pool(pool, dev_list_byid, add=True):
         resize_flag = 'delete'
     resize_cmd = [BTRFS, 'device', resize_flag, ]
     # Until we verify that all devices are or are not already members of the
-    # given pools depending on if we are adding (default) or removing
-    # (add=False) those devices we set our resize flag to false.
+    # given pool, depending on if we are adding (default) or removing
+    # (add=False), we set our resize flag to false.
     resize = False
     for d in dev_list_byid:
-        if (((resize_flag == 'add' and (d not in cur_dev)) or
-                (resize_flag == 'delete' and (d in cur_dev)))):
+        if (resize_flag == 'add' and (d not in cur_dev)) or \
+                (resize_flag == 'delete' and ((d in cur_dev) or
+                                              d == 'missing')):
             resize = True  # Basic disk member of pool sanity check passed.
             resize_cmd.append(d)
     if (not resize):
+        logger.debug('Note: resize_pool() taking no action.')
         return None
     resize_cmd.append(root_mnt_pt)
     return run_command(resize_cmd)
--- a/src/rockstor/storageadmin/models/disk.py
+++ b/src/rockstor/storageadmin/models/disk.py
@@ -107,14 +107,14 @@ class Disk(models.Model):
     def io_error_stats(self, *args, **kwargs):
         # json charfield format
         try:
-            return get_dev_io_error_stats(str(self.name))
+            return get_dev_io_error_stats(str(self.target_name))
         except:
             return None
 
     @property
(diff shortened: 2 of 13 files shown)


schakrava:
These are thorny issues you addressed in this PR @phillxnet . Very nice, thank you!

phillxnet:
@schakrava Cheers and thanks.

PR #1979, opened 2018-10-28
regression in dashboard disk activity widget. Fixes #1978

### LINKED ISSUE #1978: regression in dashboard disk activity widget
Thanks to forum member juchong for highlighting this issue. For devices that have only same length by-id names we have irregular or non functional activity in the dashboard disk activity widget. This is due to a disparity between the temp canonical to by-id name mapping employed by the widget and that used by the rest of the system. The widget, for the time being, uses a particularly light weight construct which simply uses the longest (or first if equal name lengths found) by-id name for a given canonical name in the output of 'ls -l /dev/disk/by-id'. However the rest of the system uses the algorithm instantiated within get_dev_byid_name() which is essentially the same but with additional needed safeguards logging / debug info etc. But as from pr/issue:
"improve by-id device name retrieval. Fixes #1936" #1937
applied in stable channel release version 3.9.2-28, which addressed an instability in the by-id name selection for devices that had only multiple same length by-id names, the name selection algorithms diverged. Hence prior to the referenced release the non deterministic bug present only in the rest of the system (only cosmetic side effects known) could cause an irregular mismatch with the mapping used only by the dashboard disk activity widget. And post the indicated fix the by-id name selection, in certain circumstances, would cause no activity to be registered at all for the given device: our regression.

The fix proposed by the author is to normalise the reverse lexicographical sort of by-id options employed in the improved algorithm referenced in the above fix that was used to add determinism (stability) to by-id name selection in the rest of the system to the dashboard widgets by-id map creation as well: there by re-converging the by-id name selection between the two.

https://forum.rockstor.com/t/disk-activity-plot-not-working-with-sas-drives/5351

Note that this issue only affects a device if it has both:
1. Multiple by-id names (usual).
2. All by-id names are of equal length.

phillxnet:
Recent improvements / bug fixes in the non dashboard by-id name selection has lead to a deviation in algorithms for by-id name resolution leading to a failure in the disk activity widget due to unknown by-id name selection within that widget. This only happens if all by-id names for a given device are of the same length.

Summary:

- Move dashboard 'temp name -> by-id' map generation to reverse lexicographical to normalise on by-id name selection across the system.
- Add TODO on potential dashboard code simplification in this area: has performance test pre-requisite.
- Add unit tests for both prior and current behaviour of get_byid_name_map() to ensure nature of change is understood.

Fixes #1978 
See issue text for further context.

@schakrava Ready for review

Testing:
The included additional unit tests indicate prior and proposed behaviour of the changed code with prior behaviour consistent with the observed/reported issue.
```
./bin/test --settings=test-settings -v 3 -p test_osi*
test_get_byid_name_map (system.tests.test_osi.OSITests) ... ok                                                                                                              
test_get_byid_name_map_prior_command_mock (system.tests.test_osi.OSITests) ... ok  
...
```

Caveat:
There is currently a failure to build issue in the master branch, db migration / creation related, that was worked around for the development of this issue. It may be pertinent to address this db initialisation / build issue prior to this pr's merge (I believe them to be unrelated). I will open another issue detailing my findings to date and further diagnose it's cause: looks to be related to recent db additions in master that break or surface prior hidden issues associated with clean db creation, as in a fresh build from source.

### DIFF
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -569,6 +569,9 @@ class DisksWidgetNamespace(RockstorIO):
             stats_file_path = '/proc/diskstats'
             cur_stats = {}
             interval = 1
+            # TODO: Consider refactoring the following to use Disk.temp_name or
+            # TODO: building byid_disk_map from the same. Ideally we would have
+            # TODO: performance testing in place prior to this move.
             # Build a list of our db's disk names, now in by-id type format.
             disks = [d.name for d in Disk.objects.all()]
             # /proc/diskstats has lines of the following form:
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -1620,7 +1620,7 @@ def get_dev_byid_name(device_name, remove_path=False):
 
 
 def get_byid_name_map():
-    """Simple wrapper around 'ls -l /dev/disk/by-id' which returns a current
+    """Simple wrapper around 'ls -lr /dev/disk/by-id' which returns a current
     mapping of all attached by-id device names to their sdX counterparts. When
     multiple by-id names are found for the same sdX device then the longest is
     preferred, or when equal in length then the first listed is used. Intended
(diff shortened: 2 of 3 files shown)


schakrava:
Thanks @phillxnet

phillxnet:
@schakrava Cheers.

Issue #1906, opened 2018-03-12
docker-ce dictating pool quota enabled disabled status


phillxnet:
After moving to docer-ce in stable 3.9.2-3 (pr #1865) it was found that docker-ce, upon starting (ie on boot or when it's service was switched on) would disable quotas. This at the time was very disruptive to many basic Rockstor functions so a work around was found in:
"rock-ons-root host pool quota disabled by docker-ce" #1872
and implemented in:
"rock-ons-root host pool quota disabled by docker-ce. Fixes #1872" #1873
essentially adding " '--storage-opt', 'btrfs.min_space=1G' " to dockerd's initialization.
This 'work around' effected a re-enabling of quotas by dockerd almost directly after it had initially disabled them, which it still did. This quota cycling of the affected pool on every boot was alleviated some what by improving Rockstor's quota disabled/cycling capabilities under issue:
"improve quotas not enabled behaviour" #1869
and pr:
"improve quotas not enabled behaviour. Fixes #1869" #1874
Given the above recent improvement in quota disabled behaviour a Web-UI selector for quotas / pool was introduced via issues:
"[New feature] Add option to disable BTRFS quota/qgroups" #1592 and "Feature: Quota rebuild script?" #1785 
and implemented in pr:
"Add option to disable BTRFS quota-qgroups. Fixes #1592" #1903
So we now entertain the user selectable option of Quotas Disabled. But given that a pool's quota state is remembered by the pool itself and all our existing mechanism observe and are informed by the behaviour we have again surfaced an issue with docker-ce dictating pool quota status.

This, to the issue author, looks to be the same upstream issue as was observed in the first referenced issue above (re-referenced here for ease: https://github.com/docker/for-linux/issues/78).

It is not clear how we are to proceed as currently docker-ce is now dictating the final quota state for the pool that hosts it's rock-ons-root share (subvol). This is inappropriate and not in keeping with btrfs defaults of adopting the last quota stats set, which the rest of Rockstor observes.

A workaround for those wishing to adopt a quota disabled state for a pool currently hosting the rock-ons-root is to re-create their rock-ons-root on another pool (where enabled quotas are acceptable) or potentially to revert the changes in #1873; but the latter would simply have docker-ce dictate that quotas were disabled rather than enabled shortly after boot and again circumvent the user setting (added in #1903) and the btrfs default behaviour of maintaining the last setting requested.

[...]

PR #1988, opened 2018-11-13
improve pool import with disabled or indeterminate quota state. Fixes #1987

### LINKED ISSUE #1987: improve pool import with disabled or indeterminate quota state
When attempting to import a pool which has quotas disabled or is in an indeterminate quota state the following error can result:
```
'ERROR: unable to assign quota group: Invalid argument'
```
This error can also arise when creating a new share on a pool that has an indeterminate quota state.

This is due to a mostly transient state where some quota commands work as expected, ie: 'btrfs qgroup show mnt_pt', which we use as an indicator for if quotas are enabled and to find our highest current qgroup, while others, which we normally only execute after first establishing if quotas are enabled, will fail; an example being 'btrfs qgroup assign child parent mnt_pt': hence the referenced error potentially occurring during pool import and share creation.

The proposed fix, to be presented shortly via an associated pr, is to add sensitivity / heed warnings re a quota disabled state (often only displayed for a short period and accompanied by otherwise normal quota enabled behaviour) and additionally make our 'btrfs qgroup assign' wrapper robust to this type of failure.

The latter proposed modification will avoid hard failure on import / share creation during these interim / indeterminate quota states but will result in our db assigned qgroup for the given share being unassigned. This however can self correct over subsequent pool/share updates (usually a few seconds later) or a quota disabled / enabled cycle and is non critical; unlike a pool import / share creation failure. We cannot currently avoid the rather heavy handed proposed latter modification as it has been observed where a pool's prior btrfs quota commands gave no indication; error, warning, or exception, yet an assign quota still returned 'Invalid argument'.

phillxnet:
Add sensitivity / heed warnings re a quota disabled state (often only displayed for a short period and accompanied by otherwise normal quota enabled behaviour) and additionally make our 'btrfs qgroup assign' wrapper fail elegantly in this interim / indeterminate quota state.

Addressed issue could in rare situations also block share creation.

Summary:

- Enhance qgroup_max() re qgroup disabled warnings.
- Enhance are_quotas_enabled() re qgroup disabled warnings.
- Don’t fail import / share creation for failure to quota group assign: log as error and continue.

Fixes #1987 
See issue text for more context.

@schakrava Ready for review.

Testing:

During unrelated Rockstor development on docker-ce a larger number of docker service enable / disable cycles were performed. This is suspected as the cause for a then persistent indeterminate quota state that blocked further share creation. See issue:
"docker-ce dictating pool quota enabled disabled status" #1906
for buggy docker-ce btrfs quota interaction.
It was then also noticed that importing certain prior quota disabled pools failed in a related manner.
Post pr the remaining reproducer examples behaved as expected, ie share creation was fixed and imports were successful in the respective instances.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -932,11 +932,24 @@ def are_quotas_enabled(mnt_pt):
     Simple wrapper around 'btrfs qgroup show -f --raw mnt_pt' intended
     as a fast determiner of True / False status of quotas enabled
     :param mnt_pt: Mount point of btrfs filesystem
-    :return: True on rc = 0 False otherwise.
+    :return: True on rc = 0 and (err = '' or err = rescan recommended warning),
+    False otherwise.
     """
     o, e, rc = run_command([BTRFS, 'qgroup', 'show', '-f', '--raw', mnt_pt])
-    if rc == 0:
+    if rc == 0 and (e[0] == '' or e[0] == 'WARNING: qgroup data inconsistent, '
+                                          'rescan recommended'):
         return True
+    # Note on above e[0] clauses:
+    # Catch rare but observed rc == 0 when quotas disabled, e.g.:
+    # btrfs qgroup show -f --raw /mnt2/rock-pool
+    # WARNING: quota disabled, qgroup data may be out of date
+    # qgroupid         rfer         excl
+    # --------         ----         ----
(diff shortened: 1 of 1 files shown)


schakrava:
Thank you @phillxnet

phillxnet:
@schakrava Cheers, and thanks for merging.

PR #1962, opened 2018-09-12
Add support for mounting devices to Rockons. Fixes #1957

FroggyFlox:
Fixes #1957 
Add support for the docker run `--device` option to Rockons to allow users to add custom device(s) to their Rockon(s), based on the existing support for the docker run 'environment' variables. This PR is now functional and ready for feedback and review.

Briefly, this PR adds support for a new **"devices"** object in Rockons' json definition files, following the same structure as for the **"environment"** variables with "description" and "label" fields in the JSON file (see below).
```
{
  "description": "path to hardware transcoding device (/dev/dri/renderD128)",
  "label": "VAAPI device"
}
```

This, in turn, triggers the Rockon install wizard to create a new **Device Choice** page, with as many free text input fields as devices listed in the JSON file. The text input in these fields is then validated under the following rules:
1. optional: if blank, validate.
2. if filled, the input must begin with '/dev/' and only contain alphanumerical characters (thanks @phillxnet  for your regex definition).

Testing:
All changes in python files pass Flake8.
The style of the `rockon.js` file was fixed by ESLint following style settings described in the `.eslintrc.js` file in the rockstor-core repo.
The JSON file used for testing is the following, with 3 optional devices listed:
https://github.com/FroggyFlox/rockon-registry/blob/emby_dotNET/embyserver3devices.json

Issue #1334, opened 2016-06-02
Allow removal of manually added storage in Rockons


maxhq:
Currently it is only possible to add storage (i.e. docker data volumes) to a Rockon. There seems to be no way to remove storage that was (e.b. accidentially) added before.

rcastberg:
Could i also add request the possibility to edit other properties like environmental parameters?

magicalyak:
Just for clarification, the only way to modify the storage (remove it) is to uninstall the rockon and reinstall it (which is pretty time consuming).  The edit environmental parameters may be best as a separate request.  I would honestly consider using an API or CLI or json import/export technique to do the fancy stuff.  I would image we'd just need to ensure the postgres keeps up with it.
Also this is just me thinking aloud, I wouldn't necessarily make my comment fact just wanted to suggest splitting if the remove button can be done quick and easy so it isn't held up.

PR #1999, opened 2018-12-10
Implement Add Labels feature for already-installed Rock-Ons. Fixes #1998

### LINKED ISSUE #1998: Support Docker container labels
Many Docker containers now make use of containers labels and allowing the user to specify labels would be much helpful in establishing multi-containers and even multi-Rock-ons environment. 

As there isn't--to the best of my knowledge--support for container labels to `docker update`, this would need to be done at `docker run` time, which in Rockstor could be done either at the RockOn definition stage (using the `options` object) or using the `update` function. As `labels` would be specific to each user's environment, the latter is greatly preferred.

I thus suggest to implement an "Add label" feature from the Rock-On Settings page, similar to the existing "Add Storage" feature. This would also need to support adding several labels at once as multiple labels are usually needed to setup a custom multi-container multi-RockOn environment.

I'm currently finishing a related branch and will prepare the PR quite shortly.

FroggyFlox:
Implement new feature to allow users to specify Docker container labels to already-installed Rock-Ons. Fixes #1998 

### Concept and implementation  
This new feature is modeled after the current "Add storage" option to allow the users to facilitate the creation of multi-containers Rock-ons and most importantly multi-Rock-Ons environments, for which container labels can be crucial for identification of containers by other containers in the same environment. The Docker-aware Traefik reverse-proxy, for instance, makes [heavy use of labels](https://docs.traefik.io/user-guide/docker-and-lets-encrypt/#labels) to set its options and recognize which containers to place behind its proxy.

In Rockstor, and as mentioned in the corresponding issue (#1998), these containers labels can be defined either at the definition stage (`json` file) or from the webUI by the user, making use of the already setup container_update feature for the "Add Storage" option. As the latter is the only one allowing the user to define its own custom labels easily, it is greatly preferred.
As a result, this PR implement a new "Add Label" button found under the `RockOn Settings` window, next to the existing "Add Storage" button. Its function is showcased below:  

issue1998_rockonaddlabel

  
As each label needs to be applied at the **container level** rather than the Rock-On level, the first step requires the user to select the container to which the label(s) are to be applied through a drop-down menu showing all containers included in the Rockon:  

image

Then, a dynamic series of `input text` fields allow the user to define one or more labels. The current limit is set at 10, mainly for cosmetic reasons as the window tends to overflow on shorter screen spaces.

Then, the list of container --> label mapping is summarizing for the user to confirm. Upon submission, the `update` procedure is started, in which Rockstor uninstalls the Rock-On before re-installing it with the new label(s) applied using the `--label xxx` flag to the `docker run` command, repeated as many times as there are labels defined. Note that although Docker established some character restrictions in labels, these are not enforced (yet?) so I chose not to enforce them either. **Let me know if you'd like this validation added.** See the relevant [Docker documentation](https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations):  

> Label keys should begin and end with a lower-case letter and should only contain lower-case alphanumeric characters, the period character (.), and the hyphen character (-). Consecutive periods or hyphens are not allowed.
> The period character (.) separates namespace “fields”. Label keys without namespaces are reserved for CLI use, allowing users of the CLI to interactively label Docker objects using shorter typing-friendly strings.
> These guidelines are not currently enforced and additional guidelines may apply to specific use cases.  

This PR was rebased on Rockstor 3.9.2-45 and tested under a clean new build.

### Known limitations
@schakrava , some points could be further improved, but let me know if any is considered blocking for now:  

1. The `input type=text` fields are not perfectly aligned horizontally, as well as their dynamic buttons.  
2. Related to the arbitrary limit of labels set to 10, no scrollbar appears when needed... I couldn't find a clean and robust way (robust between browsers) to implement it as of yet.  
3. Similarly, their horizontal alignment was forced in [CSS](https://github.com/FroggyFlox/rockstor-core/blob/9d03d2bd472ae46891002bdf91fc1986fc5f27d4/src/rockstor/storageadmin/static/storageadmin/css/style.css#L1251-L1259) as repasted below:     

```
.input-btn {
    display: inline !important;
    width: 70% !important;
}


#field1 {
    display: inline !important;
    width: 83% !important;
}
```

### Main changes 

1. New storageadmin model to store newly-defined labels: _DcontainerLabel_ A migration file is included in the PR. It was generated and tested as follows:  
```
[root@rockdev ~]# /opt/build/bin/django makemigrations storageadmin
Migrations for 'storageadmin':
  0007_auto_20181210_0740.py:
    - Create model DContainerLabel
    - Alter unique_together for dcontainerlabel (1 constraint(s))

[root@rockdev ~]# /opt/build/bin/django migrate storageadmin
Operations to perform:
  Apply all migrations: storageadmin
Running migrations:
  Rendering model states... DONE
  Applying storageadmin.0007_auto_20181210_0740... OK
The following content types are stale and need to be deleted:

    storageadmin | networkinterface
    storageadmin | poolstatistic
    storageadmin | sharestatistic

Any objects related to these content types by a foreign key will also
be deleted. Are you sure you want to delete these content types?
If you're unsure, answer 'no'.

    Type 'yes' to continue, or 'no' to cancel: no

[root@rockdev ~]# /opt/build/bin/django showmigrations
admin
 [ ] 0001_initial
auth
 [X] 0001_initial
 [X] 0002_alter_permission_name_max_length
 [X] 0003_alter_user_email_max_length
 [X] 0004_alter_user_username_opts
 [X] 0005_alter_user_last_login_null
 [X] 0006_require_contenttypes_0002
contenttypes
 [X] 0001_initial
 [X] 0002_remove_content_type_name
django_ztask
 (no migrations)
oauth2_provider
 [X] 0001_initial
 [ ] 0002_08_updates
sessions
 [ ] 0001_initial
sites
 [ ] 0001_initial
smart_manager
 [ ] 0001_initial
 [ ] 0002_auto_20170216_1212
storageadmin
 [X] 0001_initial
 [X] 0002_auto_20161125_0051
 [X] 0003_auto_20170114_1332
 [X] 0004_auto_20170523_1140
 [X] 0005_auto_20180913_0923
 [X] 0006_dcontainerargs
 [X] 0007_auto_20181210_0740
```

2. Additions and changes to [storageadmin JS models](https://github.com/FroggyFlox/rockstor-core/blob/9d03d2bd472ae46891002bdf91fc1986fc5f27d4/src/rockstor/storageadmin/static/storageadmin/js/models/models.js#L175-L193): Fix URL for `Container` and `ContainerCollection`, needed to pull up the containers included in the Rock-On of interest. After looking throughout the repo for the use of the `Container` model but failing to see any, I decided to modify it to fit my needs. In particular, the `model` definition for the `ContainerCollection` was switched from `Image` to `Container`, and the corresponding urls for API calls were corrected. 
`url(r'^/docker/containers/(?P<rid>\d+)$', RockOnContainerView.as_view(), ),` and 
`    url(r'^/labels/(?P<rid>\d+)$', RockOnLabelView.as_view(), ),`.  

3. Additions and changes to [storageadmin JS models](https://github.com/FroggyFlox/rockstor-core/blob/9d03d2bd472ae46891002bdf91fc1986fc5f27d4/src/rockstor/storageadmin/static/storageadmin/js/models/models.js#L753-L771): `RockOnLabel` and `RockOnLabelCollection`, with their corresponding additions in urls for API calls.  

4. Changed template for `RockonSettingSummaryWizard` to a new one dedicated one with the addition of the new "Add Label" button next to the existing "Add Storage" one.  

5. Similarly, add new templates for `add-labels` and `add-labels-form`.  

6. Fix previously incorrect fetch of Rock-On _Devices_ to be displayed in the RckOnSettingsSummary table. I improperly implemented that in #1962; I apologize for missing this in this previous PR.  

7. Flake8 and ESLint formatting on newly-added code was verified.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/add_labels_form.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/add_labels_form.jst
@@ -1,0 +1,35 @@
+<div class="row">
+  <div class="col-md-10">
+    <label class="control-label"></label>
+    <div class="form-box">
+      <form id="container-select-form" name="aform" class="form-horizontal">
+	<div class="messages"></div>
+	{{#unless containers}}
+	<h3>There are no containers to label.</h3>
+	{{else}}
+	<div class="form-group">
+	  <label class="col-sm-3 control-label" for="containers">Container: <span class="required">*</span></label>
+
+	  <div class="controls col-sm-5">
+	    <select class="form-control" id="container" name="container" placeholder="Select a container. ">
+	      <option></option>
+	      {{#each containers}}
+	      <option value="{{this.name}}">{{this.name}}</option>
+	      {{/each}}
+	    </select>
+	  </div>
+	  <i class="fa fa-info-circle fa-lg" title="Select the container to which to a label should be added."></i>
+	</div>
+ 	<div id="label-box1" class="form-group">
+	  <label class="col-sm-3 control-label" for="labels">Label:  <span class="required">*</span></label>
+	  <div class="controls col-sm-5">
+          <input class="form-control input-btn" name="labels[]" id="field1" placeholder="Enter a label" type="text" /><button id="b1" class="btn" type="button">+</button>
+	  </div>
+	  <i class="fa fa-info-circle fa-lg" title="Enter the desired label in the following form: mycustomlabel"></i>
+	</div>
+	<div class="label-box-new"></div>
+    {{/unless}}
+	  </form>
+  	</div>
+  </div>
+</div>
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/add_labels_form.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/add_labels_form.jst
@@ -18,7 +18,7 @@
 	      {{/each}}
 	    </select>
 	  </div>
-	  <i class="fa fa-info-circle fa-lg" title="Select the container to which to a label should be added."></i>
+	  <i class="fa fa-info-circle fa-lg" title="Select the container to which a label should be added."></i>
 	</div>
  	<div id="label-box1" class="form-group">
 	  <label class="col-sm-3 control-label" for="labels">Label:  <span class="required">*</span></label>
--- a/src/rockstor/storageadmin/migrations/0007_auto_20181210_0740.py
+++ b/src/rockstor/storageadmin/migrations/0007_auto_20181210_0740.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('storageadmin', '0006_dcontainerargs'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DContainerLabel',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('key', models.CharField(max_length=1024, null=True)),
+                ('val', models.CharField(max_length=1024, null=True)),
+                ('container', models.ForeignKey(to='storageadmin.DContainer')),
+            ],
+        ),
+        migrations.AlterUniqueTogether(
+            name='dcontainerlabel',
+            unique_together=set([('container', 'val')]),
+        ),
+    ]
--- a/src/rockstor/storageadmin/models/__init__.py
+++ b/src/rockstor/storageadmin/models/__init__.py
@@ -45,7 +45,8 @@ from pool_balance import PoolBalance  # noqa E501
 from tls_certificate import TLSCertificate  # noqa E501
 from rockon import (RockOn, DImage, DContainer, DPort, DVolume,  # noqa E501
                     ContainerOption, DCustomConfig, DContainerLink,  # noqa E501
-                    DContainerEnv, DContainerDevice, DContainerArgs)  # noqa E501
+                    DContainerEnv, DContainerDevice, DContainerArgs,
+                    DContainerLabel)  # noqa E501
 from smart import (SMARTAttribute, SMARTCapability, SMARTErrorLog,  # noqa E501
                    SMARTErrorLogSummary, SMARTTestLog, SMARTTestLogDetail,  # noqa E501
                    SMARTIdentity, SMARTInfo)  # noqa E501
--- a/src/rockstor/storageadmin/models/rockon.py
+++ b/src/rockstor/storageadmin/models/rockon.py
@@ -165,3 +165,13 @@ class DContainerDevice(models.Model):
     class Meta:
         unique_together = ('container', 'dev')
         app_label = 'storageadmin'
+
+
+class DContainerLabel(models.Model):
+    container = models.ForeignKey(DContainer)
+    key = models.CharField(max_length=1024, null=True)
+    val = models.CharField(max_length=1024, null=True)
+
+    class Meta:
+        unique_together = ('container', 'val')
+        app_label = 'storageadmin'
--- a/src/rockstor/storageadmin/serializers.py
+++ b/src/rockstor/storageadmin/serializers.py
@@ -24,8 +24,8 @@ from storageadmin.models import (Disk, Pool, Share, Snapshot, NFSExport,
                                  NFSExportGroup, SFTP, AdvancedNFSExport,
                                  OauthApp, NetatalkShare, Group, PoolBalance,
                                  SambaCustomConfig, TLSCertificate,
-                                 RockOn, DVolume, DPort, DCustomConfig,
-                                 DContainerEnv, DContainerDevice,
+                                 RockOn, DContainer, DVolume, DPort, DCustomConfig,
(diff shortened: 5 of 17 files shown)


FroggyFlox:
Prior to last commit bdfa1bd, newly-defined container labels were not deleted from the storageadmin DContainerLabel object for the Rock-on in question. As a result, these labels persisted after Rock-on uninstall and were automatically re-applied upon Rock-on re-install. 
This was of course not the expected behavior as the Rock-on uninstall should remove all user-defined settings. 

The latest commit bdfa1bd corrects this by deleting all labels at Rock-on uninstall:
https://github.com/rockstor/rockstor-core/blob/bdfa1bdc7eb90f627193e4707320391fe86d1ba0/src/rockstor/storageadmin/views/rockon_id.py#L184

I apologize again for this omission in the original PR submission. @schakrava, I believe this is now ready for review.

phillxnet:
@FroggyFlox Just had an initial look at this pr: all as you describe for me also.
Will get back to this on Friday to complete review and take it from there.
All looking good, and a very nice pull request. Thanks for all the testing info also.
Apologies for taking so long at my end.

FroggyFlox:
Thanks for having a first look, @phillxnet!
It actually made me realize I was not preventing the user to click the "Add Labels" button if the rock-on is running the same way as for the "Add Storage" feature. I fixed it with commit ae8f31a by extending the check for "running status" of the rock-on to all wizard buttons on this page.
This will also be also needed for my ongoing work.

Sorry again for the omission, I hope there isn't more like that.

src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/add_labels_form.jst
```diff
@@ -0,0 +1,35 @@
...
+
+	  <div class="controls col-sm-5">
+	    <select class="form-control" id="container" name="container" placeholder="Select a container. ">
+	      <option></option>
+	      {{#each containers}}
+	      <option value="{{this.name}}">{{this.name}}</option>
+	      {{/each}}
+	    </select>
+	  </div>
+	  <i class="fa fa-info-circle fa-lg" title="Select the container to which to a label should be added."></i>
```
phillxnet:
```suggestion
	  <i class="fa fa-info-circle fa-lg" title="Select the container to which a label should be added."></i>
```

phillxnet: Just a trivial typo, but user facing. Marking in passing.

FroggyFlox:
Thanks for your review @phillxnet!
I committed your suggested change (nice and convenient feature there!) for the typo, and will answer your other concerns below.

> No way to remove a label once added; other than uninstall and reinstall. Not major as is in keeping but just noting.

Yes, I have noted that one and was planning on addressing this in a future PR. Indeed, removing rock-on post-install customizations in general would be needed and was actually mentioned in a previous issue (#1334). As noted in this issue, its resolution may rely on another one (#2000), so that's why I originally was thinking of keeping this separate from this PR. Currently, removing a label would be done following the same logic as the removing a storage (uninstall-reinstall). I agree it's not the best way to do so, so let me know if you'd rather have me include it in this one.

> * It's possible to add an empty label when adding more than one at a time. But this seems harmless currently and doesn't translate through to the "docker inspect" output. But this might trip things up going forward.

If my memory is correct, the empty value is simply ignored and not transmitted to the api call, this is why I kept the form validation as it. As you mentioned, docker currently does not enforce any restriction on labels (see my original post in this PR), so I took advantage of that. As for the previous point, let me know if you'd like some form validation there. 

Thanks again!

phillxnet:
@FroggyFlox Merging as this is really very nice and we need to get it in ready for your exciting pending follow ups.

> (nice and convenient feature there!)

Yes, thank for the pointers (back channel) on that beta GitHub feature. Seems to work a treat for the little stuff at least. We can keep an eye on how it goes and only use for trivial stuff: it's likely intended use.

Thanks for tending to the typo and pointers re the related issues for Rock-on editing:

> so that's why I originally was thinking of keeping this separate from this PR.

Agreed.

> the empty value is simply ignored and not transmitted to the api call,

I still see db entries:

empty-container-label-value-in-db

and consequent Web-UI entries for these null labels:

empty-label-via-spanner

after the following multiple label entry:

two-labels-one-empty

So it appears we have some more JS work on our hands re the label addition input:

must-be-stopped-before-it-can-be-updated

Noting so we have a record and can follow up in due course. Plus I expect, given your report, that with your pending changes this 'running container' label addition issue may already be fixed.

No worries on these, we are always on a successive approximation on the code anyway and I want to get this in to ease further development and tend to our backlog.

Thanks for all your efforts, they are helping to lift our Rock-ons subsystem to something a lot more capable which will be much appreciated.

FroggyFlox:
Thanks a lot @phillxnet!

I sincerely apologize for dropping the ball on the empty labels bug. I was convinced I tested that one out but I obviously did not (or not correctly at least).
I'll try to get that fixed as soon as possible, probably fixing the parsing bit of code.

> Plus I expect, given your report, that with your pending changes this 'running container' label addition issue may already be fixed.

Indeed, that one should have been fixed by my last commit.

Thanks again!

FroggyFlox:
> (...) probably fixing the parsing bit of code.

Scratch that, I just tested and realized why you mentioned it needed to be fixed at the JS part. I'll get to that as soon as possible.

Issue #1795, opened 2017-08-12
rock-on UI fails to reflect removed container


phillxnet:
Thanks to forum member absentbri for highlighting this issue. When a container is removed via command line the installed status of that container is not reflected in the UI. It is further proposed that the Web-UI uninstall procedure attempts to remove the image if it is found as well as attempting to remove the container: this leads to a cleaner re-installation of the same rock-on and some images, such as linuxserver.io's plex server (when no plex.tv account is associated), fail to fetch the latest version with the existing mechanism as the prior image is simply re-used rather than being re-downloaded/refreshed.

Steps to reproduce this issue:

Install HTTP to HTTPS redirect rockon
```
docker ps # to list the instance
CONTAINER ID        IMAGE                   COMMAND              CREATED             STATUS              PORTS                NAMES
dbc07c961b0d        geldim/https-redirect   "/bin/sh -c nginx"   7 minutes ago       Up 7 minutes        0.0.0.0:80->80/tcp   redirect-http-to-https
```
and it's associated image:
```
docker images
REPOSITORY              TAG                 IMAGE ID           ...
geldim/https-redirect   latest              ccf561ed3790        ...
```
Turn off rock-on in UI and we have only the image and no running container instance of that image: as expected.
```
docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES

docker ps -a
CONTAINER ID        IMAGE                   COMMAND              CREATED             STATUS                       PORTS               NAMES
dbc07c961b0d        geldim/https-redirect   "/bin/sh -c nginx"   12 minutes ago      Exited (137) 2 minutes ago                       redirect-http-to-https
```

When the now off state Rockon is Uninstalled via the Web-UI the non running container instance is successfully removed:
```
docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```
but the image remains:
```
docker images
REPOSITORY              TAG                 IMAGE ID           ...
geldim/https-redirect   latest              ccf561ed3790        ...
```
As per current behaviour.

But if we reinstall this same rock-on and then turn it off (as the default to being on); but this time we remove the now non running container instance from underneath Rockstor (ie via the command line):

```
docker ps -a
CONTAINER ID        IMAGE                   COMMAND              CREATED             STATUS                        PORTS               NAMES
[...]

Issue #1653, opened 2017-02-15
[Code investigation - fixing] Rockstor Rock-ons and "orphans" subvolumes


MFlyer:
Starting from [this](https://forum.rockstor.com/t/solved-docker-hub-applications/447/3) forum post discovered that when creating Rock-ons we have some "orphan/ghost" subvols.

Checking over Rock-ons install code actually did not find where those start.
After that we need anyway to enforce volume_usage (old share_usage) to check for missing qgroup assigns : every subvol/snapshot under a share must have its qgroup assigned to parent 2015/* qgroup to have share usage right

Mirko

MFlyer:
To @schakrava 
Edit after @phillxnet suggestion: those orphans by docker itself, right - I was focused on Rockstor, without thinking to docker eheh

Anyway, agree with Phil saying we should decide how to handle them (delete? / update docker to latest and use new prune func? / etc)

phillxnet:
The referenced thread in this issue and more specifically my post in it:
https://forum.rockstor.com/t/solved-docker-hub-applications/447/11
states that as of docker version 1.13 (which as of current Rockstor versiont 3.9.1-4 is default) we have included now the facilities within docker itself to remove unused rock-on snapshots/subvols:

"But as of docker version 1.13 there is a facility to 'prune' unused elements:

    docker system prune

The above also has container, volume, and image variants in place of the system option."

A reference for this docker feature is sighted in that forum post.

PR #2016, opened 2019-02-08
honour tag element within rock-on json. Fixes #2014

### LINKED ISSUE #2014: honour tag element within rock-on json
The Rock-on systems is a work in progress and as such has 'planned' elements defined within the README.md (rockon-registry: https://github.com/rockstor/rockon-registry/blob/master/README.md) that are as yet unimplemented. One such element is the per image "tag".

Complete our existing support for the "tag" element as defined in the above repo. Currently we process and store that is found per image but it is not actually honoured.

phillxnet:
Works by simply appending the already processed 'tag' json element within Rock-on definition files to docker image related commands.

### Summary:
- honour existing "tag" element as defined in rockon-registry repo
- add debug logging to the main docker commands.
- minor docstring addition.

Fixes #2014 

Ready for review by contributor with docker and Rock-on code experience.

### Testing:
A test case Rock-on used was the existing duplicati2-canary.json as it has, unlike almost all our Rock-ons, a 'tag' element specified.
This Rock-on was installed pre-pull request and the indicated image used was latest (Docker Hub default):

| CONTAINER ID | IMAGE | COMMAND | CREATED | STATUS | PORTS | NAMES |
| ----------------- | ---------------- | ----------------- | -------------- | ------------ | ---------- | ----------- |
| b0967279e794 | intersoftlab/duplicati | "/entrypoint.sh"  | 6 minutes ago | Up 6 minutes | 0.0.0.0:8200->8200/tcp  | duplicati2-canary |

Post pr we get the stipulated image variant (via it's tag):

| CONTAINER ID | IMAGE | COMMAND | CREATED | STATUS | PORTS | NAMES |
| ----------------- | ---------------- | ----------------- | -------------- | ------------ | ---------- | ----------- |
| f017b9e812eb | intersoftlab/duplicati:canary | "/entrypoint.sh"  | About a minute ago |  Up About a minute | 0.0.0.0:8200->8200/tcp  | duplicati2-canary |

### Caveats:
When removing and re-installing this Rock-on, having applied the changes proposed here. we are left with the old image:

```
docker image list

REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE

intersoftlab/duplicati   latest              276bb62b547d        6 months ago        477MB

intersoftlab/duplicati   canary              bbeb28956224        9 months ago        476MB
```

Currently published affected Rock-ons:
- owncloud (suspected broken currently for unrelated reasons)
- duplicati2-canary

Haproxy-letsencrypt, and Collabora Online both also have the tag element but it is set to 'latest'.

### DIFF
--- a/src/rockstor/storageadmin/views/rockon_helpers.py
+++ b/src/rockstor/storageadmin/views/rockon_helpers.py
@@ -57,8 +57,8 @@ def rockon_status(name):
 
 
 def rm_container(name):
-    o, e, rc = run_command([DOCKER, 'stop', name], throw=False)
-    o, e, rc = run_command([DOCKER, 'rm', name], throw=False)
+    o, e, rc = run_command([DOCKER, 'stop', name], throw=False, log=True)
+    o, e, rc = run_command([DOCKER, 'rm', name], throw=False, log=True)
     return logger.debug(('Attempted to remove a container ({}). Out: {} '
                         'Err: {} rc: {}.').format(name, o, e, rc))
 
@@ -74,7 +74,7 @@ def generic_start(rockon):
     try:
         for c in DContainer.objects.filter(
                 rockon=rockon).order_by('launch_order'):
-            run_command([DOCKER, 'start', c.name])
+            run_command([DOCKER, 'start', c.name], log=True)
     except Exception as e:
         logger.error(('Exception while starting the '
                      'rockon ({}).').format(rockon.name))
@@ -97,7 +97,7 @@ def generic_stop(rockon):
     try:
         for c in DContainer.objects.filter(
                 rockon=rockon).order_by('-launch_order'):
-            run_command([DOCKER, 'stop', c.name])
+            run_command([DOCKER, 'stop', c.name], log=True)
     except Exception as e:
         logger.debug(('Exception while stopping the '
                      'rockon ({}).').format(rockon.name))
@@ -226,7 +226,8 @@ def generic_install(rockon):
     for c in DContainer.objects.filter(rockon=rockon).order_by('launch_order'):
         rm_container(c.name)
         # pull image explicitly so we get updates on re-installs.
-        run_command([DOCKER, 'pull', c.dimage.name])
+        image_name_plus_tag = c.dimage.name + ':' + c.dimage.tag
+        run_command([DOCKER, 'pull', image_name_plus_tag], log=True)
         cmd = list(DCMD2) + ['--name', c.name, ]
         cmd.extend(vol_ops(c))
         # Add '--device' flag
@@ -241,35 +242,47 @@ def generic_install(rockon):
         cmd.extend(port_ops(c))
         cmd.extend(container_ops(c))
         cmd.extend(envars(c))
-        cmd.append(c.dimage.name)
+        cmd.append(image_name_plus_tag)
         cmd.extend(cargs(c))
-        run_command(cmd)
+        run_command(cmd, log=True)
 
 
 def openvpn_install(rockon):
+    """
+    Custom config for the openvpn Rock-on install.
+    :param rockon:
+    :return:
+    """
     # volume container
     vol_co = DContainer.objects.get(rockon=rockon, launch_order=1)
     volc_cmd = list(DCMD) + ['--name', vol_co.name, ]
     volc_cmd.extend(container_ops(vol_co))
-    volc_cmd.append(vol_co.dimage.name)
-    run_command(volc_cmd)
+    image_name_plus_tag = vol_co.dimage.name + ':' + vol_co.dimage.tag
+    volc_cmd.append(image_name_plus_tag)
+    run_command(volc_cmd, log=True)
     # initialize vol container data
     cco = DCustomConfig.objects.get(rockon=rockon)
     oc = DContainer.objects.get(rockon=rockon, launch_order=2)
     dinit_cmd = list(DCMD) + ['--rm', ]
     dinit_cmd.extend(container_ops(oc))
-    dinit_cmd.extend([oc.dimage.name, 'ovpn_genconfig', '-u',
+    image_name_plus_tag = oc.dimage.name + ':' + oc.dimage.tag
+    dinit_cmd.extend([image_name_plus_tag, 'ovpn_genconfig', '-u',
                       'udp://%s' % cco.val, ])
-    run_command(dinit_cmd)
+    run_command(dinit_cmd, log=True)
     # start the server
     server_cmd = list(DCMD2) + ['--name', oc.name, ]
     server_cmd.extend(container_ops(oc))
     server_cmd.extend(port_ops(oc))
     server_cmd.append(oc.dimage.name)
-    run_command(server_cmd)
+    run_command(server_cmd, log=True)
 
 
 def owncloud_install(rockon):
+    """
+    Custom config for the owncloud Rock-on install.
+    :param rockon:
+    :return:
+    """
     for c in DContainer.objects.filter(rockon=rockon).order_by('launch_order'):
         rm_container(c.name)
         cmd = list(DCMD2) + ['--name', c.name, ]
@@ -292,7 +305,8 @@ def owncloud_install(rockon):
                         '-e', 'HTTPS_ENABLED=true'])
             cmd.extend(['-e', 'DB_USER=%s' % db_user, '-e',
                         'DB_PASS=%s' % db_pw, ])
-        cmd.append(c.dimage.name)
+        image_name_plus_tag = c.dimage.name + ':' + c.dimage.tag
+        cmd.append(image_name_plus_tag)
         logger.debug('Docker cmd = ({}).'.format(cmd))
         run_command(cmd)
         if (c.dimage.name == 'postgres'):


phillxnet:
@FroggyFlox I'm popping this one in now as it's a dependency for another pr I'm working on in rockon-registry. I will rebase later if need be as we go as I know you have made a number of changes in this area that are both pending in pr and pre pr.

FroggyFlox:
Thanks for fixing this! I didn't even realize that we had the "tag" supported and parsed... I should have paid more attention.

Regarding the image uninstall, I believe it is "simply" due to the fact that we do not remove the docker image upon Rock-on uninstall. Luckily the `docker rmi` command support tags so if/when we implement that, we would be able to easily deal with such situation. Should we open a new issue to remove docker images? That would need a quick usage check for the image before removing it, but I believe we have already all elements in our database to do that.


> I know you have made a number of changes in this area that are both pending in pr and pre pr

I do, indeed, have extensive additions and edits to `rockon_helpers.py` in "pre-PR" state, but they won't be ready that soon so do not worry about these for now.

I'll test and make a proper Github review of this one if you want, but that seems good at first glance. I'll have a better look at it as soon as possible.

phillxnet:
@FroggyFlox Thanks for the review offer. Much appreciated.
As per orphaned image / container clean-up, there were plans to use a 'new at the time' docker command to do just that, can't find the discussion now unfortunately. May have just been in my notes.
Also see #1795 for a prior discussion on leaving images post Rock-on uninstall. We could re-visit that of course as ultimately it's a non sustainable practice. Have a think and we can spin off another issue if need be. Pretty much a corner case with regard to this issue but I wanted to mention it as it's kipple.

phillxnet:
@FroggyFlox We can always just lookup which share is our Rock-ons-root as we do when blocking that share's deletion. So we can likewise look to a snapshots parent share to see if it is thus. At least I think so (bit late here now).

Plus I found the prior chat re image pruning via docker:
See pr: #1653 and it's comments.
Plus the following forum thread referenced there:
https://forum.rockstor.com/t/solved-docker-hub-applications/447/11
```
docker system prune
```
And it's variations is what we were waiting for and now have. More details in forum thread.

FroggyFlox:
Thanks for all the pointers and references of prior discussions, @phillxnet ; as I didn't know about them, it's always very interesting and useful ;-)

> Also see #1795 for a prior discussion on leaving images post Rock-on uninstall. We could re-visit that of course as ultimately it's a non sustainable practice. Have a think and we can spin off another issue if need be. Pretty much a corner case with regard to this issue but I wanted to mention it as it's kipple.

Agreed to re-visit this... I'll be meaning to open a discussion about that (here or forum), especially after repeatedly discussing in the forum the use of an "advanced scripts" section that could include `delete-rockon` and `docker prune` tools... Off topic here, but definitively worth laying out the best way to implement a solution to these issues.

> We can always just lookup which share is our Rock-ons-root as we do when blocking that share's deletion. So we can likewise look to a snapshots parent share to see if it is thus. At least I think so (bit late here now).

I forgot about that one... I've been so focus on the network side of Rockstor lately that I can't see anything else. I'll keep this for the corresponding issue too.

I'll try to come back with a review to the current PR as soon as possible.

FroggyFlox:
I started testing your PR and it does work as intended.

First, thanks for adding `log=true` to these commands. This is something I almost always do in my dev environment but always ignore when committing as I thought it would not have been desired behavior. It is extremely helpful not only for debugging but also for troubleshooting issues any user may have so I'm definitely happy to see this :-)

Regarding your PR, I'll submit a proper review with comments where I do have some, but those are only that: comments. The code works as indicated.

There is one comment I have that is not related to your commits, but actually concerns how we are parsing the `"tag"` object from the json. Note also that it is a "fringe" case that I do not really see happening in reality.

The code in question is from [rockon.py](https://github.com/rockstor/rockstor-core/blob/5046c02e69f6bd2b70cec43e04886a93d5a4163f/src/rockstor/storageadmin/views/rockon.py#L200-L201): 
```
            io, created = DImage.objects.get_or_create(name=c_d['image'],
                                                       defaults=defaults)
```
As stated above, we are feeding the `DImage` model filtering by image name, for each container in the json, so this does not allow to have the same image with different tags in the same rock-on, such as in the following test json:
```
		"containers": {
			"alpine36": {
				"image": "alpine",
				"tag": "3.6",
				"launch_order": 1,
				"ports": {},
				"volumes": {},
				"environment": {}
			},
  			"alpine_edge": {
				"image": "alpine",
				"tag": "edge",
				"launch_order": 2,
				"ports": {},
				"volumes": {},
				"environment": {}
			},
			"alpine3": {
				"image": "alpine",
				"tag": "3",
				"launch_order": 3,
				"ports": {},
				"volumes": {},
				"environment": {}
			}
```
As mentioned above, however, I'm not sure this would happen in real use, but should we prepare for that? It does not break the system, only prevents the use of different tags.

Notably, it also indicates another issue more likely to arise as the above applies if _two different_ rock-ons use the same image but with a different tag. The first rock-on to be processed with take priority in this case, such that the second one will use the tag of the first one.

Again, these are not directly related to the issue your PR addresses, so I can create a dedicated one if you prefer. 

Cheers,

phillxnet:
@FroggyFlox Thanks for taking a look at this pr and checking functionality. Much appreciated.

Re your comment on our image/tag db table (storageadmin_dimage) that's a nice find, missed that myself, and definitely deserves a focused issue. But I agree with you that it's rather in the corners, but if we have the issue we can more easily keep it in mind. Especially as we get more Rock-ons / images.

Re the log=True I think is the way to go for this as it's still very much in development (by you mainly) and only logs if debug is enabled anyway so shouldn't affect production releases. And as you say makes for easier diagnosis 'in the field' as when debug is on we can see those docker commands and their call order.

Cheers.

src/rockstor/storageadmin/views/rockon_helpers.py
```diff
@@ -241,35 +242,47 @@ def generic_install(rockon):
...
+    :param rockon:
+    :return:
+    """
     # volume container
     vol_co = DContainer.objects.get(rockon=rockon, launch_order=1)
     volc_cmd = list(DCMD) + ['--name', vol_co.name, ]
     volc_cmd.extend(container_ops(vol_co))
-    volc_cmd.append(vol_co.dimage.name)
-    run_command(volc_cmd)
+    image_name_plus_tag = vol_co.dimage.name + ':' + vol_co.dimage.tag
```
FroggyFlox: My only concern is that this function does not seem to be in use anymore. It doesn't hurt to also implement your changes there, of course.

src/rockstor/storageadmin/views/rockon_helpers.py
```diff
@@ -292,7 +305,8 @@ def owncloud_install(rockon):
                         '-e', 'HTTPS_ENABLED=true'])
             cmd.extend(['-e', 'DB_USER=%s' % db_user, '-e',
                         'DB_PASS=%s' % db_pw, ])
-        cmd.append(c.dimage.name)
+        image_name_plus_tag = c.dimage.name + ':' + c.dimage.tag
```
FroggyFlox: My only concern is that this function does not seem to be in use anymore. It doesn't hurt to also implement your changes there, of course.

src/rockstor/storageadmin/views/rockon_helpers.py
phillxnet: @FroggyFlox Yes I thought this myself so during development I changed their names and then build from scratch. Should have indicated this, partly why I put the docstrings in. If you build with these missing/removed/renamed then the post install instructions in the more_info of for example the OpenVPN Rock-on fails. Also note the command executions that only exist in these as I did logging checks as well. So yes it would be nice to indicate where they are run from and pop that in the docstrings as well. Sorry but I just moved on and didn't follow through.

src/rockstor/storageadmin/views/rockon_helpers.py
phillxnet: See my previous comment. Pretty sure they are used but would be great if you confirmed this as well.

phillxnet:
@FroggyFlox Let me know if  you manage to confirm the openvpn / owncloud related code changes as I'm pretty sure I found they were not in dead code, as per my replies to your code comment review.

Thanks for the review by the way and see how you get on with that confirmation.
I can then mark as resolved at this end.

src/rockstor/storageadmin/views/rockon_helpers.py
phillxnet: My check for the owncloud_install was to turn on log for the db specific command and then install, you can then see the db stuff logged. Apologies for not noting it's caller, and PyCharm didn't find it either. But they are both called upon the specific rock-ons being executed. We will find it and must then mark otherwise this will haunt us again. But not necessarily in this pr :).

phillxnet:
@FroggyFlox Thanks for being so diligent on this by the way.

src/rockstor/storageadmin/views/rockon_helpers.py
FroggyFlox: Thanks for the information, these have been bothering me for quite a while and I can't seem to find their trigger anywhere. I'll follow the same route you did and hopefully will find where they're coming from.
Thanks again!

FroggyFlox:
> @FroggyFlox Let me know if you manage to confirm the openvpn / owncloud related code changes as I'm pretty sure I found they were not in dead code, as per my replies to your code comment review.

I finally tracked them down... I should have been more attentive while reading this, but here's where these methods are called:
In [rockon_helpers.py](https://github.com/rockstor/rockstor-core/blob/5046c02e69f6bd2b70cec43e04886a93d5a4163f/src/rockstor/storageadmin/views/rockon_helpers.py#L119-L124):
```
def install(rid):
    new_state = 'installed'
    try:
        rockon = RockOn.objects.get(id=rid)
        globals().get('%s_install' % rockon.name.lower(),
                      generic_install)(rockon)
```

Thanks a lot for finding these out, @phillxnet!

phillxnet:
@FroggyFlox Well done tracking that down, good to have gotten that sorted. So are we happy with this pull request's changes as they stand, review wise; so we can put it to bed? At least then we can 'mostly' honour tags going forward, given your nice fine re limitation of scope bug and the linked issue you opened.

FroggyFlox:
Yes, I'm all ok with your changes, sorry I sidetracked the conversation :-(

PR #2019, opened 2019-02-24
Skip empty label fields

### LINKED ISSUE #2018: [Rock-ons] Users can add empty labels
As detailed in #1999 by @phillxnet, one can add empty labels (if no characters are entered in the form field). This is currently not dealt with and result in a partly empty record in the database. Most importantly, this situation is without any use and needs to be corrected.

This was entirely due to an oversight on my end as part of PR #1999.

I have a simple fix that I am about to submit as a new PR fixing this issue.

FroggyFlox:
Add simple if statement to ignore fields with empty values for labels.

Fixes #2018.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
@@ -1241,7 +1241,9 @@ RockonAddLabel = RockstorWizardPage.extend({
             return $.Deferred().reject();
         }
         var field_data = $('input[name^=labels]').map(function(idx, elem) {
-            return $(elem).val();
+            if ($(elem).val() != "") {
+                return $(elem).val();
+            }
         }).get();
         var new_labels = {};
         field_data.forEach(function(prop) {


phillxnet:
@FroggyFlox Thanks for the quick fix, well done. Tested via custom RPM build rev 48.2019.

By way of a quick functional test we have entry of an empty value:

request-multiple-labels-one-empty

resulting in that empty value being ignored in Web-UI:

no-empty-web-ui-or-db-counterpart-results

with of course the mostly redundant double check on the storageadmin_dcontainerlabel db table:

storageadmin_dcontainerlabel

Best we move towards input validation as we go along, and as it's spec emerges in docker, but this avoids an easy inadvertent null db entry for now which would only come back to haunt us.

FroggyFlox:
Thanks a lot!
Sorry again I (somehow) missed something like that and had to make a patch.

phillxnet:
@FroggyFlox No worries at all. This is just part of the practicality of what we do and we just have to muddle through, oh and try and review each others code of course :). Speaking of which I'm wrapping up the revision 48 release now so we need to review all that's going into 49. Any chance you could glance over my pending pr. Not in the Rock-ons area but it's almost all Python so there's a bonus.

PR #2037, opened 2019-04-23
add known non unique product_uuid on GIADA N70E-DR. Fixes #2036

### LINKED ISSUE #2036: add known non unique product_uuid on GIADA N70E-DR
Thanks to forum member hammerite for reporting GIADA N70E-DR motherboards as having  non unique product_uuids. When identified during install we avoid additional complexities/delays with required stable subscription order interventions there after but at the cost of non persistent 'Appliance ID' over re-install which with non unique product_uuids is not workable anyway.

phillxnet:
Thanks to forum member hammerite for reporting GIADA N70E-DR motherboards as having non unique product_uuids. Identifying during install avoids creating duplicate Appliance IDs which can hamper some
sub-systems that depend upon a unique system id such as replication.

Fixes #2036

Ready for review.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -2056,10 +2056,13 @@ def hostid():
     ASRock J3455 ITX - Thanks to forum member adworacz for reporting this one.
     - Examples of the second listed fake_puuids
     ZOTAC 880G-ITX (880GITX-A-E) - Thanks to forum member mmmdonuts
+    - Examples of the third listed fake_puuids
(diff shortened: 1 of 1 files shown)


phillxnet:
Tested via trial build and initial setup of a proposed 3.9.2-49 version including this patch. As this pr is functionally only a single element addition to a list of known non unique product_uuid's I'm merging without review given it's simplicity and non critical code nature.

Thanks again to forum member hammerite for contributing, via private message, to discover and confirm this additional problematic product_uuid.

PR #2042, opened 2019-05-03
[openSUSE] don't rely on PGDATA env var. Fixes #2041

### LINKED ISSUE #2041: [openSUSE] don't rely on PGDATA env var
Thanks to @FroggyFlox and forum member jcdick1 for highlighting this issue. Previously we implicitly relied upon the PGDATA env variable. Some more recent openSUSE Tumbleweed releases (as of writing) no longer define this variable for the postgres user so it is suggested that, for our openSUSE code path only, we use initdb's -D option to explicitly define our desired Postresql 'data' directory.

Linking to the relevant forum thread for exposition and context:
https://forum.rockstor.com/t/build-instructions/5946/17
through to:
https://forum.rockstor.com/t/build-instructions/5946/21

phillxnet:
Explicitly define Postgresql db 'data' directory, in openSUSE code path, via initdb's -D option. Use existing in code pg_data var which is set to the normal default of '/var/lib/pgsql/data' anyway. Prior use of pg_data in this code path was ineffectual.

Fixes #2041 

Ready for review.

Please see issue text for context, including forum origin references.

### DIFF
--- a/src/rockstor/scripts/initrock.py
+++ b/src/rockstor/scripts/initrock.py
@@ -357,7 +357,7 @@ def main():
             if os.path.isfile('/usr/bin/initdb'):
                 logger.debug('running generic initdb on {}'.format(pg_data))
                 run_command(
-                    ['su', '-', 'postgres', '-c', '/usr/bin/initdb', pg_data])
+                    ['su', '-', 'postgres', '-c', '/usr/bin/initdb -D {}'.format(pg_data)])
         logging.info('Done.')
         run_command([SYSCTL, 'restart', 'postgresql'])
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox Thanks for the review, much appreciated.

As this is only a trivial single line change and it's a pain to have to account for it during development and without it, due to upstream changes, our build on an openSUSE target is non functional I'll break normal protocol and merge now ahead of the next merge round. Also doesn't look to affect any outstanding earlier pull requests so shouldn't break their prior merge-ability.

FroggyFlox:
That makes sense! Thanks for preparing and writing this PR, @phillxnet!

Issue #2060, opened 2019-08-04
support.rockstor.com needs to point to forum.rockstor.com


magicalyak:
support.rockstor.com is no longer working and needs to be replaced with references to forum.rockstor.com for issues and links in rockstor-core

phillxnet:
@magicalyak Thanks for creating this issue.

Linking to our forum discussion for context:
https://forum.rockstor.com/t/support-rockstor-com-down/6264

PR #2061, opened 2019-08-04
fixed issue 2060 with support link

magicalyak:
This fixes the support.rockstor.com link in ./src/rockstor/storageadmin/static/storageadmin/js/templates/common/popuperr.jst
It simply removed the if statement for stable and leaves the forum link for opening an issue.
https://github.com/rockstor/rockstor-core/issues/2060


phillxnet:
@magicalyak Nice. And super fast work here.

We just need a quick double check review and this can go in (given it's relatively simple nature). Adding relevant "needs review" label.

PR #2063, opened 2019-08-13
[openSUSE] improve JeOS compatibility re ROOT label. Fixes #2062

### LINKED ISSUE #2062: [openSUSE] improve JeOS compatibility re ROOT label
OpenSUSE JeOS defaults to mounting by label. This differs from current openSUSE Leap 15.* and Tumbleweed that use mount by UUID. Both Leap and Tumbleweed also currently default to having no root volume label. Rockstor expects and enforces btrfs volume labels, using the btrfs volumes UUID in the case of data pools and 'system' in the case of root pools. However btrfs based openSUSE JeOS (Just enough Operating System) images default to a root volume (pool in Rockstor speak) label of "ROOT".

It is proposed that we adopt this "ROOT" default in favour of our existing "system" one to improve our JeOS compatibility and enable a consistent experience re system pool naming across Leap 15.* Tumbleweed and openSUSE JeOS image based installs.

This change should only affect existing openSUSE source based installs. But is intended to be in-place prior to distributing openSUSE based rpms.

phillxnet:
To enable the use of btrfs based openSUSE JeOS images we required a change from our default root pool label expectation (enforced default where non found) to be in line with the default found, and depended upon (fstab mount by label) in these JeOS images.

Summary of changes:
- Change 'no root pool label' auto label mechanism to use 'ROOT'; in line with btrfs based openSUSE JeOS images.
- Provide a mechanism to migrate existing source installs over to this new default: prior default was 'system'.

Fixes #2062 

Testing.
An existing openSUSE (Tumbleweed) source based install was recompiled in place and found to successfully self migrate from a 'system' root pool label to the new default of 'ROOT'. Note that source based installs are expected to be recompiled and there are, as yet, no rockstor rpms distributed for openSUSE.

### DIFF
--- a/conf/settings.conf.in
+++ b/conf/settings.conf.in
@@ -279,7 +279,7 @@ SFTP_MNT_ROOT = '/mnt3/'
 
 # System volume label when no btrfs volume label is set as per default openSUSE
 # install: ie 'btrfs fi show' gives 'Label: none'
-SYS_VOL_LABEL = 'system'
+SYS_VOL_LABEL = 'ROOT'
 
 TAP_DIR = '${django-settings-conf:taplib}'
 TAP_SERVER = ('127.0.0.1', ${django-settings-conf:tapport})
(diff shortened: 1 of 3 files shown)


phillxnet:
Merging to aid in our openSUSE development effort as it enables us to use otherwise unaltered JeOS images. Also, given this is potentially a breaking change for those developing on openSUSE I'd like to get it in sooner rather than later.

Issue #1954, opened 2018-08-21
[Feature] Add an existing user as Rockstor managed during a config restore


doenietzomoeilijk:
As discussed [here](https://forum.rockstor.com/t/solved-how-hosed-am-i-cleared-out-the-database-turns-out-not-that-hosed/5115/15?u=doenietzomoeilijk): when restoring a config backup that contains user johndoe, the restore of that user will be skipped if that user is already present on the system. It'd be nice if - if the uid matched - this user would instead be considered the user we were talking about, and added as a Rockstor-managed user instead.

phillxnet:
@doenietzomoeilijk Thanks for creating this issue. I've just had a quick look for an initial pointer on this and the following code is be a good place to start:
https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/storageadmin/views/config_backup.py#L52-L75

Noting in case this issue takes anyone's fancy.

phillxnet:
Just a note re:
> - if the uid matched -
I think we should consider matching names here, all depends on how the share access is setup and given we currently specify uid gid in Rock-ons these make sense but who ever takes this one on should consider this as the first user is always 1000 and each subsequent user has their UID incremented which makes for arbitrary uid to username allocation. Anyway just a thought.

Also note openSUSE / CentOS differences re the users default group and it's differing treatment.
I.e:
CentOS: defaults to each user having their own group created by user name.
openSUSE: defaults to all users belonging to the pre-existing users (GID=100) group.

I think in a NAS scenario the latter is more appropriate as it can ease beginner admin of shared access rights.

FroggyFlox:
@phillxnet, I'm currently looking into this one and wanted to verify something. Indeed, when you write the following:

> openSUSE: defaults to all users belonging to the pre-existing users (GID=100) group.

... do you think that, during a restore into openSUSE, we should:
1. _force_ all Rockstor-managed users present in a backup to the GID=100 group?
or
2. should we simply keep the same UID/GID structure as present in the backup?

Option 2 would be a lot easier to implement and--I believe--should guarantee all permissions are correct system-wide, but I do agree with you that option 1 is more appropriate in a NAS scenario.



> * I think we should consider matching names here

Agreed, especially considering the CentOS/openSUSE differences you noted above.

phillxnet:
@FroggyFlox Re:
> 2\. should we simply keep the same UID/GID structure as present in the backup?

[...]

PR #2066, opened 2019-08-24
[WIP] Include rock-ons to config backup/restore feature

### LINKED ISSUE #2065: Config backup/restore does not include rock-ons
Currently, the config backup/restore feature does not include information related to rock-ons. As a result, all settings specified by the user during a rock-on installation or post-installation are not saved and thus need to be re-entered manually by the user if needed. 

It is thus proposed to include rock-ons-related information to the config backup, and implement corresponding restore function(s).

I currently have a working "proof of concept" branch related to this issue and will submit a draft pull-request to present a suggested resolution.

FroggyFlox:
**_Note_**: this pull request is currently a **work-in-progress** and aims at presenting a current strategy to resolve #2065. It is not complete yet and thus not ready for review.

#### Backup
All rock-ons-related models are added to the storageadmin database dump call. In addition, the `sotrageadmin.share` model is also included in the datadump call as it is required to fetch the new share id for the `DVolumes` objects that need to be re-created during restore. 

#### Restore
As all rock-ons-related models are interconnected by their respective primary key, it is necessary to recreate the request to be sent to the rockons API with the correct information by fetching each information one at a time for each rock-on to be re-created. Once the request is re-created, it is used to send a call to the rock-on API with the `install` command. The normal back-end install procedure thus takes care of re-installing the rock-on with the same settings as present in the config backup.

Notably, as the storageadmin models related to rock-ons contain information for all rock-ons, installed or not, it is necessary to first extract the list of rock-ons that need to be re-installed, _i.e._ those with the `state == 'installed'`--as these correspond to rock-ons that were installed (started or stopped) at time of the creation of the config backup.

Moreover, the `restore_rockons()` procedure is called last, to ensure the restoration of the rock-on service has completed before trying to install rock-ons.

### What's working
All settings currently set during the rock-on install wizard are currently implemented, allowing users to restore their current rock-ons environment _in the same state as their fresh installation_. Post-install cutomizations remains to be implemented, however.

### What remains to be implemented
Post-installation customizations, including `DVolume` objects (need to set those with `uservol=true`) and `DContainerLabel` objects.

We currently have a specific check for pending rock-on install that prevents a simple for-looping through rock-ons requests as one install process takes some time before receiving the `installed` state. We may need to find a way to wait for this `installed` state before sending the next rock-ons API call.

### Potential improvements
To prevent any risk for conflicting database information, or re-writing the wrong model object, it is required to make sure the current rock-ons-related database information is up-to-date by clicking on the "Update" button in the webUI. There is an initial check for proper matching between rock-on ID and its name before proceeding with the restore, but this check currently fails silently and does not warn the user about the need to update the database (if any).

Similarly, any rock-on that was installed from local metastore (`rockons-metastore`) will not be restored if the JSON file and its information has not already been detected and processed by clicking on the "Update" button in the rock-ons webUI page.

Due to the fact that the rock-ons-related storageadmin models include information related to **all** available rock-ons (installed or not), the config backup is a lot bigger than before. For instance, the new config backups weigh a little more than 23 kB, while the same system would create a backup of less than 1 kB previously. While these file sizes are still very small, they do represent a sizable jump in size (relatively) that should be noted. If this represents a problem, we can try to filter for installed rock-ons prior to backup creation. As the `dumpdata` call does not support this kind of filtering (to the best of my knowledge), however, this would require manual filtering work on the backup file before saving it to disk.


FroggyFlox:
I now have all information needed for both the initial install _and_ the post-install update parsed and processed correctly. Moreover, as mentioned above, the restore procedure for a single rock-on without update needed works well. I am still encountering problems with either multiple rock-ons restore, or restore of a rock-on needing post-install update.

In these cases, we need to account for an internal check (`_pending_check()`) present in most rock-on procedures (including install and update) verifying that no other rock-on operation is currently in progress before doing anything. In our case, this prevents us from simply starting several install requests back-to-back. 

A first workaround I tried was to use `time.sleep` until `_pending_check()` returns `False` before proceeding to the next step of the restore process. Unfortunately, `time.sleep` used in this context simply pauses everything and as a result does not let the previous pending operation finish.

I thus tried to experiment with using threading to start the wait loop in a different thread so that `time.sleep()` wouldn't pause the rock-on install process, but unfortunately it restarts the entire `restore_config()` from scratch each time, which is not a good solution either. Maybe this is related to the fact that `restore_config()` is itself ran as a `django-ztask`?

I'll try to explore a little more, but any feedback or suggestion is welcome.

FroggyFlox:
> I am still encountering problems with either multiple rock-ons restore, or restore of a rock-on needing post-install update.

Some progress on this. Starting these install procedures in ztasks themselves seems to help a lot with the procedure: a single rock-on with post-install customizations, which need install + update procedures, now work. However, this "success" seems limited to two rock-ons at a time. Indeed, if trying to restore 4 rock-ons, only two are processed correctly and the other two continue to fail due to the same error ("another rock-on in transition").

Re-triggering the config backup restore from the webUI will work, however, as it is skipping the first two (now already restored) and can thus restore the remaining two rock-ons successfully.

I'm not sure triggering these in async (as per django-ztask) is a good idea, but it seems to be helping this situation so I'll keep trying to dig deeper in this direction. @phillxnet, let me know if it's a no-go for you and I'll try to find alternative(s).

FroggyFlox:
As of the latest commit, this branch seems to reliably function as tested with the restore of 4 rock-ons at once, including one multi-containers rock-on, and with or without post-install customization.

I'm thus going to focus on writing new unit tests for the new parsing logic in `views/config_backup.py`.

FroggyFlox:
The restore logic now seems to be reliably working for multiple rock-ons with or without post-install customizations (shares and labels) and related testing has been expanded and improved.
In particular, I have corrected the POST tests to use a mock ConfigBackup object to prevent restoring an actual backup object that was very likely to be present in the database (see bbac9ae).

As this branch will need to be re-based if/when #2059 is merged, I'll go ahead and do some further testing before submitting a proper PR.

FroggyFlox:
After further testing, it appears that if a rock-on is present in the backup as "to be restored" but not on the current machine (as it can be the case for locally defined rock-ons via rockons-metastore), the whole restore-rockons process fails and even those present on the current machine fail to be restored.

We thus need to improve this by skipping rock-ons that are not detected on the current machine.

FroggyFlox:
> We thus need to improve this by skipping rock-ons that are not detected on the current machine.

This has now been taken care of by 335effb.

FroggyFlox:
As we have a lot of rock-ons using the _PUID_ / _PGID_ environment variables, we may prefer to have #1954 resolved before proceeding further. As @phillxnet notes in the issue linked above, we need to also account for differences in GID attribution between CentOS / openSUSE, so we may have to modify these ENV from backup config to match current system/Rockstor users before restoring them.

FroggyFlox:
> As we have a lot of rock-ons using the _PUID_ / _PGID_ environment variables, we may prefer to have #1954 resolved before proceeding further.

After further thinking, #1954 may not represent a blocking issue for this PR as we only need to worry about correctly restoring what is present in the backup (as per the focus of this PR).

We should, however, improve our restore of _PUID_ / _PGID_ environment variables by trying to first match the corresponding _user_ / _group_ from the previous system to the new system by name and use the latter's UID / GID. Thus, in the even the corresponding user was already created (or restored) using different UID / GID, these _PUID_ / _PGID_ environment variables will still point to valid references.

FroggyFlox:
A proper PR has now been submitted as #2100 . 

I will thus close this draft one.

PR #2081, opened 2019-11-08
[openSUSE] use zypper for package installs. Fixes #2071

### LINKED ISSUE #2071: [openSUSE] use zypper for rockstor package install
Experimentation has indicated that although yum is mostly functional for our purpose in our openSUSE variant their are complications re reported conflicts of rockstor rpm dependencies that are not similarly reported by zypper. So we should, at least for install only initially, use zypper. Currently it is though that our use of YUM for reporting / changelog etc is working as intended so we need only change the actual install side of our Web-UI initiated 'rockstor' package management for now.

It is similarly very likely better to move to zypper (on openSUSE) for our regular non rockstor package updates also. The same code manages both so this issue may be able to address both also.

phillxnet:
Add openSUSE variant distro specific zypper equivalents of our prior yum only functions re updates and testing/stable repo management.

Fixes #2071 

Includes:
- Contextual zypper calls replace equivalent yum ones when on openSUSE.
- Refactoring: including function/procedure renaming.
- Include Black code syntactic reformatting.
- Fix prior hard wired x86-64 file name parsing.
- Add rpm import of ROCKSTOR-GPG-KEY on openSUSE systems.
- Add zypper addrepo/removerepo function along side existing yum counterparts for rockstor stable and testing channels.
- Use zypper up/dup respectively for general Leap/Tumbleweed update/dist-upgrade.
- Use --no-recommends to zypper up/dup commands to keep footprint small.
- Add user text for no version/changelog available on openSUSE.
- Move prior AT initiated delay to within external script and reduce from 1 minute to 10 seconds.
- Move string formatting within pkg_mgmt.py from 'c' style to string.format().
- Avoid removing RPM package managed *.pyc files from ./eggs dir prior to rockstor rpm package install/upgrade.
- Make user facing update dialog/tooltip text non yum specific.
- Begin pkg management unit test coverage.

Ready for review.

Apologies to reviewer in advance, this pr includes Black re-formatting as the associated python file was relatively small but the code changes were larger than I had expected. Should be doable however.

Testing.
Functional testing was performed on all 3 distro bases: CentOS, openSUSE Leap15.1 / Tumbleweed. All three were found to be able to sense and present, and consequently update all non rockstor packages. And testing/stable repo instantiation/switching, along with rpm --import of the rockstor public PGP key (on openSUSE variants only) looked to be working as expected.

Unit tests.
Minimal unit tests were introduced with the following output post pr merge:

```
./bin/test --settings=test-settings -v 3 -p test_pkg_mgmt*
test_pkg_changelog (system.tests.test_pkg_mgmt.SystemPackageTests) ... ok
test_pkg_update_check (system.tests.test_pkg_mgmt.SystemPackageTests) ... ok
test_zypper_repos_list (system.tests.test_pkg_mgmt.SystemPackageTests) ... ok
```

Caveats:
- This pull request does not cover unattended updates for the openSUSE variants, but a TODO was added to the relevant code to indicate this lack of feature parity. It is proposed that this can be addressed in a focused issue of it's own.
- There was not found to be an equivalent, yet, of pending update per package change logs. However a potential fix for this may be in upstream. For now we inform the user of this unavailability and a code TODO has been added linking to the upstream feature addition to zypper.

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -17,6 +17,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
 import os
+import platform
 import re
 import stat
 from tempfile import mkstemp
@@ -24,7 +25,7 @@ from osi import run_command
 from services import systemctl
 import shutil
 import time
-from datetime import (datetime, timedelta)
+from datetime import datetime, timedelta
 import requests
 from django.conf import settings
 from system.exceptions import CommandException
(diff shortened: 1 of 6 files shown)


src/rockstor/system/pkg_mgmt.py
```diff
@@ -33,78 +34,81 @@
...
+                    tfo.write("apply_updates = no\n")
                 updated = True
             else:
                 tfo.write(line)
-    if (not updated):
-        raise Exception('apply_updates directive missing in %s, assuming its '
-                        'is corrupt. No change made.' % YCFILE)
+    if not updated:
+        raise Exception(
+            "apply_updates directive missing in {}, assuming its "
```
FroggyFlox: Minor typo:
```suggestion
            "apply_updates directive missing in {}, assuming it "
```

phillxnet: @FroggyFlox Thanks for this.
Not one of mine surprisingly but good to have it in anyway. Committing suggestion. Tricky with us doing simultaneous Black formatting. We will have to keep that to small files only when rolled into other pull requests.

phillxnet:
@FroggyFlox Thanks for the review. Much appreciated. We obviously have to tread carefully on this one as if we break Web-UI updates it's a broad pain.

Were you able to also test on our legacy CentOS base? All looked OK in my testing but would be good to have a double check of course.

FroggyFlox:
> Were you able to also test on our legacy CentOS base? All looked OK in my testing but would be good to have a double check of course.

I actually just did, yes... I realized after submitting the review that it was a big missing piece so I tried in CentOS. I built your branch from a 3.9.1.0 ISO install. There was thus plenty of updates and they all went through successfully as well from the web-UI. I was able to activate Testing channel too, but I haven' tested starting those updates, however.

PR #2089, opened 2019-11-18
[Config Backup/Restore] Restore Service status

### LINKED ISSUE #2087: [Config Restore] Service status is not restored
Although the configurations of services are included in the Config backups and restored accordingly, their **status** is not.

As we keep their status (ON/OFF) in `smart_manager`, it is in theory possible to include this information in the config backup and use it during restore so that all services in the backup that were configured AND turned on would also be turned back on after a config restore.

@phillxnet, do you think it's an idea worth pursuing or is the service status not restored on purpose?

FroggyFlox:
Fixes #2087.

As described in #2087, although we already restore the configurations of our services, their **status** is NOT restored.

This PR thus proposes to save the status of services in the config backup, and use this information to turn ON the services during the restore procedure. As a result, not only would the configuration of each service be restored, but their ON/OFF status would be the same as at the time of the backup creation.

## Aims & Logic
As the ON/OFF status of each service is already stored in smart_manager under the `servicestatus` model, one can simply dump all entries from this model during the config backup creation. At the end of the current procedure to restore the config of services, we can thus now also use the information from this newly-included model to fetch the ON/OFF status of each service. Then, we can simply send a `POST` call to the sm/services API to start the given service. pending:
- the service was ON in the config backup
- the service is currently OFF on the system

## Demonstration
1. On a CentOS Rockstor install, the following services are configured and turned ON:
- Rock-on
- Samba
image

2. A config backup is created and downloaded in a separate machine.
3. Both the _Rock-on_ and _Samba_ services are turned OFF:
image

4. The config backup created in step 2 above is then restored, and we can confirm that both the _Rock-on_ and _Samba_ services are back ON:
image

5. On a freshly dev-built Leap15.1, the same config backup can be uploaded and restored, resulting in both services being turned back ON:
image

The latter step was also successfully tested in Tumbleweed.

@phillxnet, I believe this is ready for review, but let me know if you'd like additional testing. As I could only test the successful trigger of a few services, I'm not sure if you'd like to have further and more thorough testing done.

## Summary of commits
- Validate service status in config backup from name
- Check the current status of the service and turn ON if:
-- the service was ON in the backup
-- the service is currently OFF on the new install

### DIFF
--- a/src/rockstor/storageadmin/views/config_backup.py
+++ b/src/rockstor/storageadmin/views/config_backup.py
@@ -29,6 +29,7 @@ from rest_framework.response import Response
 
 import rest_framework_custom as rfc
 from cli.rest_util import api_call
+from smart_manager.models.service import Service, ServiceStatus
 from storageadmin.models import ConfigBackup
 from storageadmin.serializers import ConfigBackupSerializer
 from storageadmin.util import handle_exception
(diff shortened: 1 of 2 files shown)


phillxnet:
@FroggyFlox Thanks for an excellent pull request submission.

As discussed in other channels we still have some 'reluctant' services that just don't want to have their ON status be restored properly, ie so far UPS-NUT (works occasionally) and Replication (yet to work). But I think these 'sticky' services indicate a bug elsewhere. I'd also like to see a little more non debug info logging during the config save and restore. Just to help with knowing what's happening on regular / non development builds. All doable in time.

Let get this in and we can deal with these problem services ON state restore in their own focused issues.

Thanks again for your effort on this one. Another likely much appreciated config save/restore enhancement.

PR #2103, opened 2020-01-01
pin django-braces to restore django version compatibility. Fixes #2102

### LINKED ISSUE #2102: latest django-braces breaks build on Django version requirement
On 30th Dec 2019 django-braces version 1.14.0 was release. This has a new minimum Django requirement of 1.11.0. Our current Django version pinning of 1.8.16 is thus incompatible: breaking our current build.

As a makeshift arrangement to return our build to a functioning state it is proposed that we now pin our django-braces version to the previously released version of 1.13.0 released on Apr 22, 2018. This will permit continued development where we can also update our Django version and related dependencies.

## References
django-braces release dates:
30th Dec 2019 1.14.0 https://pypi.org/project/django-braces/1.14.0/
22nd Apr 2018 1.13.0 https://pypi.org/project/django-braces/1.13.0/

phillxnet:
On 30th Dec 2019 django-braces version 1.14.0 was release. This has a new minimum Django requirement of 1.11.0. To restore project django version compatibility we pin django-braces to the previously released version.

Fixes #2102 

Please see referenced issue for details of the observed build failure.

## Testing
Prior builds post 30th Dec 2019 fail as details in referenced issue.
Post pr patch the same build arrangements succeed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,6 +84,7 @@ setup(
         'requests == 1.1.0',
         'six == 1.10.0',
         'distro',
+        'django-braces == 1.13.0',  # 1.14.0 (30 Dec 2019) needs Django 1.11.0+
     ]
 
 )


phillxnet:
@FroggyFlox As this patch is required to return our build to a working state, and is simply a version pinning of a library that hasn't changed since 22nd Apr 2018 I'm going ahead and merging as is.

PR #2109, opened 2020-01-20
Fix samba enabled without config

### LINKED ISSUE #2108: Enabling samba doesn't always trigger configuration
When enabling samba from the UI under services, it opens an edit modal where you can supply a workgroup and custom global config. Filling out this config then overwrites the `/etc/samba/smb.conf` file with the defaults needed for Rockstor.

The relevant code for this can be found here: https://github.com/rockstor/rockstor-core/blob/fd3d07236cd26d968b00ec55d97c10ca4786c9b0/src/rockstor/storageadmin/static/storageadmin/js/views/services.js#L137-L147

This code is generic to be reused for all services listed on that page, but has a check to navigate to the edit URI if needed.

If instead you navigate to Storage->File Sharing->Samba and use the toggle there, this edit modal is never presented. This means if it has not previously been configured, this will enable samba with the default `smb.conf` supplied at install time.

The relevant code for this is here: https://github.com/rockstor/rockstor-core/blob/ba0ee33b7d2cace4e2438362be1cd84b9c84096e/src/rockstor/storageadmin/static/storageadmin/js/views/samba.js#L106-L112

This lacks any check at all for the configuration required.

---

Another example that came up during discussion is the Rock-Ons service, which properly presents the edit dialog in both locations. This is the expected behavior.

It can be seen here that it still checks for the config before toggling the service: https://github.com/rockstor/rockstor-core/blob/ba0ee33b7d2cace4e2438362be1cd84b9c84096e/src/rockstor/storageadmin/static/storageadmin/js/views/docker_service.js#L69-L81

defmonk0:
Resolves #2108

Since this was such a simple fix after I created the issue, I figured I'd put this together real quick.

Simply redirects when enabling samba from the samba page if needed to open the config modal.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/samba.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/samba.js
@@ -104,10 +104,16 @@ SambaView = RockstorLayoutView.extend({
     },
 
     switchStatus: function(event, state) {
-        if (state) {
-            this.startService();
+        if (!this.service.get('config')) {
+            app_router.navigate('services/smb/edit', {
+                trigger: true,
+            });
         } else {
-            this.stopService();
+            if (state) {
+                this.startService();
+            } else {
+                this.stopService();
+            }
         }
     },
 
--- a/src/rockstor/smart_manager/views/base_service.py
+++ b/src/rockstor/smart_manager/views/base_service.py
@@ -33,7 +33,10 @@ logger = logging.getLogger(__name__)
 class ServiceMixin(object):
 
     def _save_config(self, service, config):
-        service.config = json.dumps(config)
+        if config is not None:
+            service.config = json.dumps(config)
+        else:
+            service.config = None
         return service.save()
 
     def _get_config(self, service):
--- a/src/rockstor/system/samba.py
+++ b/src/rockstor/system/samba.py
@@ -74,7 +74,7 @@ def rockstor_smb_config(fo, exports):
             fo.write('    veto files = /.%s*/\n' % e.snapshot_prefix)
         for cco in SambaCustomConfig.objects.filter(smb_share=e):
             if (cco.custom_config.strip()):
-                    fo.write('    %s\n' % cco.custom_config)
+                fo.write('    %s\n' % cco.custom_config)
     fo.write('%s\n' % RS_SHARES_FOOTER)
 
 
@@ -180,7 +180,9 @@ def update_global_config(smb_config=None, ad_config=None):
 
 
 def get_global_config():
-    config = {}
+    # start with config as None so it will return null elsewhere
+    # if no fields are added to it.
+    config = None
     with open(SMB_CONFIG) as sfo:
         global_section = False
         global_custom_section = False
@@ -201,9 +203,9 @@ def get_global_config():
             # we ignore lines outside [global], empty lines, or
             # commends(starting with # or ;)
             if ((not global_section or
-                not global_custom_section or
-                len(l.strip()) == 0 or
-                re.match('#', l) is not None or
+                 not global_custom_section or
+                 len(l.strip()) == 0 or
+                 re.match('#', l) is not None or
                  re.match(';', l) is not None)):
                 continue
             if (global_section and re.match('\[', l) is not None):
@@ -212,7 +214,13 @@ def get_global_config():
             fields = l.strip().split(' = ')
             if len(fields) < 2:
                 continue
+            # reset config to a usable dictionary since we're about to add
+            # something to it
+            if(config is None):
+                config = {}
             config[fields[0].strip()] = fields[1].strip()
+    # return our config variable, which is either None, or a dictionary with
+    # fields inside of it
     return config
 
 


phillxnet:
@defmonk0 Well that was fast. Well done.

I've labelled "needs review" and given you've been working with @FroggyFlox on the forum on this one hopefully he can take a look. If not I can take a look soon and if all works out we can get this into the next release.

Thanks for diving in and having a go at this. Quite a nice find as it goes.

src/rockstor/storageadmin/static/storageadmin/js/views/samba.js
```diff
@@ -104,10 +104,16 @@ SambaView = RockstorLayoutView.extend({
     },
 
     switchStatus: function(event, state) {
-        if (state) {
-            this.startService();
+        if (!this.service.get('config')) {
```
FroggyFlox: Unfortunately, the samba service is an exception to the other services as its `'config'` attribute is not `null` but an empty json string when not configured. This is due to:
https://github.com/rockstor/rockstor-core/blob/ba0ee33b7d2cace4e2438362be1cd84b9c84096e/src/rockstor/system/samba.py#L182-L216

We would thus need to either change the above, or simply change how we check for an unconfigured smb service here:
```suggestion
        if (_.isEmpty(JSON.parse(this.service.get('config')))) {
```

defmonk0: It may be worth noting that the services page may also be an issue here then.

https://github.com/rockstor/rockstor-core/blob/fd3d07236cd26d968b00ec55d97c10ca4786c9b0/src/rockstor/storageadmin/static/storageadmin/js/views/services.js#L137-L147

The if statement there has the same exact check. If I'm understanding that correctly, it wouldn't work either way then, and only ever opens when configuring samba manually by clicking the wrench icon.

If that's the case, it may be best to change the return value to match the other services. This would fix it and keep it consistent with the others.

Since it's python, start as `config = None`, and then swap it to a dictionary right before adding something to it:

```python
if config is None:
    config = {}
config[fields[0].strip()] = fields[1].strip() 
```

I'm not sure if there would be any other ramifications to that though. Does this check happen anywhere else for samba?

FroggyFlox: Very good points and I completely agree with you.

As to why that happens, I believe it "simply" is a matter of order. Note, however, that all I'm going to write is from my current understanding so still needs to be verified before I can say for sure. Indeed, in a brand new Rockstor install, the `config` field is empty (null), and remains as such when the user visits the services page. But, upon visiting the Storage > Samba page, `get_global_config()` is triggered and save `{}` as config in the model.

I believe your approach is far superior to what I originally proposed and would allow us to stay consistent across pages and services.

> I'm not sure if there would be any other ramifications to that though. Does this check happen anywhere else for samba?

That's exactly why I originally was afraid of touching `get_global_config()` in the first place but because this seems to be the only place it's called, it should be ok to apply your suggestions to it. We'll need to be more careful with testing it, though of course.

Thanks for catching this.

defmonk0:
I added another commit to make the changes we talked about for `get_global_config`. Once we get the chance to confirm this is in a good place, I can squash the commits if wanted.

FroggyFlox:
Hi @defmonk0,

Thanks for making those changes. It does work as intended re the return of `None` when the service hasn't been configured, but this in turns is saved as `null` in the model's config field (as opposed to being set to null). As a result, the JS check believes it is configured and starts the service nonetheless.

A quick fix would be to save the config only if not None in `SambaServiceView`:
```
    def get(self, request, *args, **kwargs):
        with self._handle_exception(self.request, msg=None):
            so = Service.objects.get(name=self.service_name)
            config = get_global_config()
            logger.debug('smbconfig is {}'.format(config))
            if config is not None:
                self._save_config(so, config)
            sd = ServiceStatusSerializer(self._get_or_create_sso(so))
            return Response(sd.data)
```

Of course, it may simply mean we're not correctly parsing the save into database part, but I would be worried of messing a lot of things up as that part is shared with the all the other services (I believe):
https://github.com/rockstor/rockstor-core/blob/ba0ee33b7d2cace4e2438362be1cd84b9c84096e/src/rockstor/smart_manager/views/base_service.py#L35-L37

phillxnet:
@defmonk0 @FroggyFlox Just had a quick look at this again.

Is anyone working on this, ie re @FroggyFlox last comment.

defmonk0:
I had meant to come back to this, but haven't had a time to dig in to fully understand whats going on. I didn't want to continue making changes without making informed choices on what I was changing.

If it's as simple as adding the one line if-statement mentioned by @FroggyFlox, I can do it, but I don't have a full understanding of why the value being null in the model allows it to enable the service.

Unless he is referring to another location I'm not aware of, we check the config via `if (!this.service.get('config'))`, which should return true when the config is set to `null`, opening the edit modal (instead of enabling or disabling the service).

phillxnet:
I'm planning, if all goes well, to have a look at this issue tomorrow (UK time), it looks like it's almost there now and have now setup my Pycharm db watcher to have a go at it. Agreed that we should try and normalise on null value in db for all services and I've seen the literal string "null" happen before so if I have a look it may come back to me what's going on. It should always be that a Python None equates to null value when entered in Django db. Anyway I'll have to get in and have a look.
@defmonk0 I'll try and get a docs together on your question re django to js stuff.

FroggyFlox:
@defmonk0 ,

> Unless he is referring to another location I'm not aware of, we check the config via `if (!this.service.get('config'))`, which should return true when the config is set to `null`

My apologies for a poor explanation... @phillxnet did a much better job in the following:

>  the literal string "null"

That's exactly what is happening here. For instance, before any configuration the Samba service, here's how the corresponding `smart_manager` table is filled:
```
linux-ahh6:/opt/build/src/rockstor # psql -U rocky -d smartdb -c "SELECT * from smart_manager_service"
Password for user rocky: 
 id |        name        |   display_name   |                              config                              
----+--------------------+------------------+------------------------------------------------------------------
  9 | rockstor           | Rockstor         | 
 10 | smartd             | S.M.A.R.T        | 
  2 | smb                | Samba            | 
 11 | active-directory   | Active Directory | 
  5 | ntpd               | NTP              | 
 12 | nut                | NUT-UPS          | 
 13 | snmpd              | SNMP             | 
  8 | sftp               | SFTP             | 
 14 | docker             | Rock-on          | 
  1 | replication        | Replication      | 
 15 | shellinaboxd       | Shell In A Box   | {"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}
  3 | nfs                | NFS              | 
  7 | ldap               | LDAP             | 
  6 | nis                | NIS              | 
 16 | netatalk           | AFP              | 
 17 | ztask-daemon       | ZTaskd           | 
 18 | rockstor-bootstrap | Bootstrap        | 
(17 rows)

```

As you can see, the `config` field is indeed empty (true `null`). Upon visiting the Samba page in Rockstor UI, however, the following is called:
https://github.com/rockstor/rockstor-core/blob/ba0ee33b7d2cace4e2438362be1cd84b9c84096e/src/rockstor/smart_manager/views/base_service.py#L35-L37

... which returns the **string** _null_ and saves it into the database:
```
linux-ahh6:/opt/build/src/rockstor # psql -U rocky -d smartdb -c "SELECT * from smart_manager_service"
Password for user rocky: 
 id |        name        |   display_name   |                              config                              
----+--------------------+------------------+------------------------------------------------------------------
  9 | rockstor           | Rockstor         | 
 10 | smartd             | S.M.A.R.T        | 
 11 | active-directory   | Active Directory | 
  5 | ntpd               | NTP              | 
 12 | nut                | NUT-UPS          | 
 13 | snmpd              | SNMP             | 
  8 | sftp               | SFTP             | 
 14 | docker             | Rock-on          | 
  1 | replication        | Replication      | 
 15 | shellinaboxd       | Shell In A Box   | {"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}
  3 | nfs                | NFS              | 
  7 | ldap               | LDAP             | 
  6 | nis                | NIS              | 
 16 | netatalk           | AFP              | 
 17 | ztask-daemon       | ZTaskd           | 
 18 | rockstor-bootstrap | Bootstrap        | 
  2 | smb                | Samba            | null
(17 rows)

```
This is why the JS part (linked below) see it as not null:
https://github.com/rockstor/rockstor-core/blob/ad66a6596e32eaf249d4821e7623eb06088e873c/src/rockstor/storageadmin/static/storageadmin/js/views/samba.js#L107

To the best of my knowledge, this is due to the following:
https://github.com/rockstor/rockstor-core/blob/ba0ee33b7d2cace4e2438362be1cd84b9c84096e/src/rockstor/smart_manager/views/base_service.py#L35-L37

... and more specifically how `json.dumps()` seems to return a `"null"` string for `None`. I'm not familiar with this part of Python so I'm unsure of the best way to deal with such parsing. You may very well know exactly how to deal with that, actually.

Hope this helps, and sorry for not being clearer before.

defmonk0:
@FroggyFlox That makes so much more sense. Thanks for the clarification.

Rather than stopping it from saving altogether if the config is `null`, would it be viable and more consistent to throw an extra check into `_save_config`? IE, only do a `json.dumps` if config is not `null`. That way, it still attempts to save an actual `null` if for some reason we somehow clear the config out somewhere.

It seems safer, but that might just be my remaining unfamiliarity talking. Lol.

phillxnet:
@FroggyFlox Thanks for the confirmation here, and break down thus far. That's pretty much what I though you were getting at my self, as it goes. I'm currently looking at the original and about to pickup @defmonk0 branch thus far with what you have both been up-to to have a look at that 'as is'.

> I'm not familiar with this part of Python so I'm unsure of the best way to deal with such parsing.

I think we are looking at:
```
>>> import json
>>> print (json.dumps(None))
null
```
And as json is a string based notation we have our Python _None_ object returned in json representation as a json string "null", it's most direct counterpart.

Our db model for service (src/rockstor/smart_manager/models/service.py) is a CharField defined thus:
```
class Service(models.Model):

    name = models.CharField(max_length=24, unique=True)
    display_name = models.CharField(max_length=24, unique=True)
    config = models.CharField(max_length=8192, null=True)

    class Meta:
        app_label = 'smart_manager'
```
Note the null=True.

Still fathoming the related mechanisms, haven't looked at this part of the code for a while now.

I'm currently thinking that you @FroggyFlox have already arrived at the last part of the answer here though on your last code suggestion. A little bit clunky some how but if it works! And if it's mostly obvious to our future selves.

phillxnet:
@defmonk0 Our last post were near enough simultaneous so I didn't get a good look at your suggestion.
@FroggyFlox, what do you think of @defmonk0 suggestion re your own, not saving. It would enable us to 'reset' our config's in a way we currently can't. I'm still weighing this lot up myself as it goes. But I like the idea / utility of a 'return to null' option. That may end up being quite useful as currently we don't have this, from my initial perusal thus far the idea is growing on me. Especially since our front end triggers from true null entries in this field, having the ability to pop them back in at will might turn out to be quite useful. Plus it might get this issue put to bed ;)

FroggyFlox:
> @FroggyFlox, what do you think of @defmonk0 suggestion re your own, not saving.

I do prefer it too. I just tested it quickly and saving a config of `None` into the model does set it to a true `null` so that seems to work.

> It would enable us to 'reset' our config's in a way we currently can't. I'm still weighing this lot up myself as it goes. But I like the idea / utility of a 'return to null' option. That may end up being quite useful as currently we don't have this, from my initial perusal thus far the idea is growing on me.

I do like this idea very much, but maybe not for this PR as it would seem to me a tad off-topic. I'm still wondering how we could best do that, for instance, as this would also benefit from being applicable to all configurable services. Should we have a dedicated "Reset" button on the configuration modal, or simply allow empty fields in the configuration modals to set this config to `null`?  Either way, I agree with you that @defmonk0 's idea of changing the `_save_config()` method to allow for saving `None` (into `null`) is not only more elegant than my original band-aid solution, but would also allow for the reset as you stated. 

Thanks a lot, @defmonk0 ! @phillxnet might already been having a look at it, but your suggestion above may be ready to be included in your PR.

phillxnet:
> I do like this idea very much, but maybe not for this PR as it would seem to me a tad off-topic.

Definitely agree this is off-topic for pr but just trying to evaluate the future utility of the various solutions here.

> Should we have a dedicated "Reset" button on the configuration modal, or simply allow empty fields in the configuration modals to set this config to `null`?

I think the former is wiser as our entry forms often have minimum entry which would then be defeated by the latter. Plus a reset is 'serious' so should be accompanied by a "This action will remove all prior Rockstor config for this service. Are you sure" type thing. As for a reset option it would also have to remove all shares and resource say a disto default config file, in the case of the likes of samba, plus our minimum required additions of course. So essentially formalising our clean start scenario maybe. And given we need to pass _None_ when our forms can't, due to input validation, a seperate dedicated path via a 'reset' button looks like the better option. A discussion for the forum perhaps. But such a facility would also help with development actually, as we can then return to default for each specific service to test it's reconfiguration from scratch as it were, while also exercising this reset feature in the process. We will just have to resource the relevant packages distro config as per your recent docker config changes in our "Built on openSUSE" variants.

defmonk0:
> Thanks a lot, @defmonk0 ! @phillxnet might already been having a look at it, but your suggestion above may be ready to be included in your PR.

Added a commit for this change as well and rebased to keep the branch up to date!

FroggyFlox:
Thanks a lot for the quick commit, @defmonk0 .

I quickly tried it in a clean new Leap 15.1 install and it works as intended :) .

I did notice an odd redirection to the Services page from the Samba page if one click cancel after the configuration modal pops-up, but I think that's a more generalized quirk (read common to all services and was not introduced by this PR) and thus might belong to a separate PR. 

I'll test that further and make sure we're still OK with all the other services and in all 3 Rockstor flavors. but it all seems good so far!

phillxnet:
@FroggyFlox 

> I did notice an odd redirection to the Services page from the Samba page if one click cancel after the configuration modal pops-up

I've just noticed / confirmed this also, but not directly after the cancel; only after refreshing the storage - samba page to get a switch update as the switch shows as on for me after this cancel. 
And again on all three builds I get redirected to the services page directly after "OK" on pop-up dialog samba config. I think this is 'normal' though for the time being.

I suspect we agree that these can be dealt with in another pr.

My testing was as follows
- Ensure smb / nmb are disabled. Or switch has been turned off.
- Remove all prior Rockstor 'tags' from the smb.conf
- Visit & refresh the Storage - samba page (this ensures we write our null in our config)
- Try turning on the switch on that page. As we have our no-config-found, null, we are presented with the dialog.

@FroggyFlox If you're happy with this pr I'm happy to merge it in it's current state as it does fix the original issue of a difference between the two switches re smb.conf template use. Which hopefully clears the way for your enhancement in that area. We have some page switching artefacts but that's way better than our silent config strangeness.

Let me know either way here and we can take it form there.

FroggyFlox:
As the functions affected by this PR relate to the configuration prompt for the Samba service as well as the proper population of its related Django model, my testing focused on these two points.

- In a freshly built Rockstor (using this branch), the Samba service is OFF, and the related database info set to `null`:
```shell script
rockdev:/opt/build/src/rockstor # psql -U rocky -d smartdb -c "SELECT * from smart_manager_service"
Password for user rocky: 
 id |        name        |   display_name   |                              config                              
----+--------------------+------------------+------------------------------------------------------------------
  9 | rockstor           | Rockstor         | 
 10 | smartd             | S.M.A.R.T        | 
 11 | active-directory   | Active Directory | 
  5 | ntpd               | NTP              | 
 12 | nut                | NUT-UPS          | 
 13 | snmpd              | SNMP             | 
  8 | sftp               | SFTP             | 
 14 | docker             | Rock-on          | 
  1 | replication        | Replication      | 
 15 | shellinaboxd       | Shell In A Box   | {"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}
  3 | nfs                | NFS              | 
  7 | ldap               | LDAP             | 
  6 | nis                | NIS              | 
 16 | netatalk           | AFP              | 
 17 | ztask-daemon       | ZTaskd           | 
 18 | rockstor-bootstrap | Bootstrap        | 
  2 | smb                | Samba            | 
(17 rows)
```

- Visiting the _Services_ does not affect the database and the user is correctly prompted with the configuration modal if it were to toggle the Samba service ON (as in current upstream/master). Clicking "Cancel" on this modal, returns the user to the _Services_ page and the database is updated with the `null` information. This wasn't the case in current upstream/master where an empty dict `{}` would be populated for the Samba config record. That's a first improvement.

> Note that upon clicking "Cancel" and closing the modal, the Samba service toggle shows as ON, but a refresh of the page returns it to OFF (its true state). This was a quirk that was already present in upstream/master.  




```shell script
rockdev:/opt/build/src/rockstor # psql -U rocky -d smartdb -c "SELECT * from smart_manager_service"
Password for user rocky: 
 id |        name        |   display_name   |                              config                              
----+--------------------+------------------+------------------------------------------------------------------
  9 | rockstor           | Rockstor         | 
 10 | smartd             | S.M.A.R.T        | 
 11 | active-directory   | Active Directory | 
  5 | ntpd               | NTP              | 
 12 | nut                | NUT-UPS          | 
 13 | snmpd              | SNMP             | 
  8 | sftp               | SFTP             | 
 14 | docker             | Rock-on          | 
  1 | replication        | Replication      | 
 15 | shellinaboxd       | Shell In A Box   | {"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}
  3 | nfs                | NFS              | 
  7 | ldap               | LDAP             | 
  6 | nis                | NIS              | 
 16 | netatalk           | AFP              | 
 17 | ztask-daemon       | ZTaskd           | 
 18 | rockstor-bootstrap | Bootstrap        | 
  2 | smb                | Samba            | 
(17 rows)
```

- Navigating to the *Storage* > *Samba* page does refresh the database (as before), but this time a proper `null` is populating the Samba config field (was `{}` prior to this PR).

```shell script
rockdev:/opt/build/src/rockstor # psql -U rocky -d smartdb -c "SELECT * from smart_manager_service"
Password for user rocky: 
 id |        name        |   display_name   |                              config                              
----+--------------------+------------------+------------------------------------------------------------------
  9 | rockstor           | Rockstor         | 
 10 | smartd             | S.M.A.R.T        | 
 11 | active-directory   | Active Directory | 
  5 | ntpd               | NTP              | 
 12 | nut                | NUT-UPS          | 
 13 | snmpd              | SNMP             | 
  8 | sftp               | SFTP             | 
 14 | docker             | Rock-on          | 
  1 | replication        | Replication      | 
 15 | shellinaboxd       | Shell In A Box   | {"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}
  3 | nfs                | NFS              | 
  7 | ldap               | LDAP             | 
  6 | nis                | NIS              | 
 16 | netatalk           | AFP              | 
 17 | ztask-daemon       | ZTaskd           | 
 18 | rockstor-bootstrap | Bootstrap        | 
  2 | smb                | Samba            | 
(17 rows)
```

- Clicking on the ON/OFF toggle on this page properly checks whether the service is already configured and prompts the user with the configuration modal (as it is unconfigured for us at this point).

> Note that, as above, clicking "Cancel" on the modal to close it shows the Samba service as toggled ON, and refresh brings the user to the *Services* page but I believe this is related to the quirk first mentioned above.


Notably, the database is still showing the service as unconfigured:
```shell script
rockdev:/opt/build/src/rockstor # psql -U rocky -d smartdb -c "SELECT * from smart_manager_service"
Password for user rocky: 
 id |        name        |   display_name   |                              config                              
----+--------------------+------------------+------------------------------------------------------------------
  9 | rockstor           | Rockstor         | 
 10 | smartd             | S.M.A.R.T        | 
 11 | active-directory   | Active Directory | 
  5 | ntpd               | NTP              | 
 12 | nut                | NUT-UPS          | 
 13 | snmpd              | SNMP             | 
  8 | sftp               | SFTP             | 
 14 | docker             | Rock-on          | 
  1 | replication        | Replication      | 
 15 | shellinaboxd       | Shell In A Box   | {"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}
  3 | nfs                | NFS              | 
  7 | ldap               | LDAP             | 
  6 | nis                | NIS              | 
 16 | netatalk           | AFP              | 
 17 | ztask-daemon       | ZTaskd           | 
 18 | rockstor-bootstrap | Bootstrap        | 
  2 | smb                | Samba            | 
(17 rows)
```


- From there, click on the wrench icon and configure the Samba service as follows:
image


The database is properly updated:
```shell script
rockdev:/opt/build/src/rockstor # psql -U rocky -d smartdb -c "SELECT * from smart_manager_service"
Password for user rocky: 
 id |        name        |   display_name   |                              config                              
----+--------------------+------------------+------------------------------------------------------------------
  9 | rockstor           | Rockstor         | 
 10 | smartd             | S.M.A.R.T        | 
 11 | active-directory   | Active Directory | 
  5 | ntpd               | NTP              | 
 12 | nut                | NUT-UPS          | 
 13 | snmpd              | SNMP             | 
  8 | sftp               | SFTP             | 
 14 | docker             | Rock-on          | 
  1 | replication        | Replication      | 
 15 | shellinaboxd       | Shell In A Box   | {"detach": false, "css": "white-on-black", "shelltype": "LOGIN"}
  3 | nfs                | NFS              | 
  7 | ldap               | LDAP             | 
  6 | nis                | NIS              | 
 16 | netatalk           | AFP              | 
 17 | ztask-daemon       | ZTaskd           | 
 18 | rockstor-bootstrap | Bootstrap        | 
  2 | smb                | Samba            | {"socket options": "SO_SNDBUF=131072", "workgroup": "MYGROUP"}
(17 rows)
```
The service is correctly started:
```shell script
rockdev:/opt/build/src/rockstor # systemctl status -l smb
â—� smb.service - Samba SMB Daemon
   Loaded: loaded (/usr/lib/systemd/system/smb.service; disabled; vendor preset: disabled)
   Active: active (running) since Mon 2020-02-10 13:11:32 EST; 19s ago
  Process: 6216 ExecStartPre=/usr/share/samba/update-apparmor-samba-profile (code=exited, status=0/SUCCESS)
 Main PID: 6217 (smbd)
   Status: "smbd: ready to serve connections..."
    Tasks: 4 (limit: 4915)
   CGroup: /system.slice/smb.service
           â”œâ”€6217 /usr/sbin/smbd --foreground --no-process-group
           â”œâ”€6225 /usr/sbin/smbd --foreground --no-process-group
           â”œâ”€6226 /usr/sbin/smbd --foreground --no-process-group
           â””â”€6230 /usr/sbin/smbd --foreground --no-process-group

Feb 10 13:11:30 rockdev systemd[1]: Starting Samba SMB Daemon...
Feb 10 13:11:32 rockdev systemd[1]: Started Samba SMB Daemon.
Feb 10 13:11:32 rockdev smbd[6217]: [2020/02/10 13:11:32.652094,  0] ../lib/util/become_daemon.c:138(daemon_ready)
Feb 10 13:11:32 rockdev smbd[6217]:   daemon_ready: STATUS=daemon 'smbd' finished starting up and ready to serve connections
``` 
As we can see, however, it is using the package's default `smb.service` and not Rockstor's one. The `smb.conf` file is written correctly, though:
```shell script
rockdev:/opt/build/src/rockstor # cat /etc/samba/smb.conf 
[global]
    log level = 3
    map to guest = Bad User
    cups options = raw
    log file = /var/log/samba/log.%m
    printcap name = /dev/null
    load printers = no

####BEGIN: Rockstor SAMBA GLOBAL CUSTOM####
    socket options = SO_SNDBUF=131072
    workgroup = MYGROUP
####END: Rockstor SAMBA GLOBAL CUSTOM####
```

Notably, the `smb` service is not enabled, which further supports that the proper config logic in `samba_service.py` was not triggered. This probably needs to be dealt with in the recently opened issue #2127.

FroggyFlox:
@phillxnet , sorry I didn't see your message until now...

I agree with you that all the other problems we just seem to have spotted with Samba should be dealt with in separate PR. I haven't proceeded with unit testing, though.

FroggyFlox:
@defmonk0 , @phillxnet ,

>As we can see, however, it is using the package's default smb.service and not Rockstor's one.

I believe I have identified the culprit. If I'm correct, it relates to the particular structure of `SambaServiceView` and does indeed represents a perfect candidate to be fixed while fixing #2127. While the one line responsible for this behavior most likely is simple to "fix", I believe the overall logic of the `post()` method needs to be made more robust for this kind of changes. The (likely) culprit may be:
https://github.com/rockstor/rockstor-core/blob/52b25dd921d6a45051776c9d2ac8db12f874b7b4/src/rockstor/smart_manager/views/samba_service.py#L93

phillxnet:
@FroggyFlox That's an excellent review there.

> I haven't proceeded with unit testing, though.

@FroggyFlox These mainly pertain to the API though currently (in the samba area) from memory anyway, and they are probably best dealt with in #2127 given you will then have greater freedom to resolve issues there than here.

@FroggyFlox and @defmonk0 
So is the consensus that we call this one done and consider opening fresh issues for the remaining 'artefacts' especially given the pending enhancements in #2127 as it will make the end to end functional testing easier once we are using appropriate smb.conf templates appropriate for the particular distro / samba implementation, so we don't get tripped up with service file errors while testing.

@defmonk0 We normally like to have all commits squashed but this pull request has had quite the history already and I'm hoping to see you in these parts again, once the next bug bites your :) so we could delay this one no longer and try for a squash pre-pr submission or on re-submission, or on moving from draft to ready for review, in the next rounds. Or whatever works at the time I guess. @FroggyFlox's pull requests are an excellent reference in these regards, and it helps my limited brain out with keeping track of changes. But I don't think a few commits per pr, especially to tidy up last minute small changes is too bad.

It's easy for issues / pull requests to have mission creep but that does end up making them harder to review.

All this attention on particularly samba config can only be good. It is after all our primary share mechanism. Thanks to both of you for your efforts on this one. I think you have definitely nudged the one more in the the 'proper' area.

FroggyFlox:
> So is the consensus that we call this one done and consider opening fresh issues for the remaining 'artefacts' 


That's a yes for me.

phillxnet:
As per the previous discussion here I'm going to merge this as-is given it does resolve the linked issue re: opening a config modal when no prior Rockstor stamped config exists in the smb.conf file and the samba switch is going from off to on, for both the specific samba page (the previous failure point in this regard) and the services page samba switches. So it now shares this intended function with for example our Rock-on service switches.

Thanks again to @defmonk0 and @FroggyFlox for getting this rather long standing bug resolved. Always more to do but at least this one is now sorted thus far.

PR #2119, opened 2020-02-04
Remove 'latest' from the Rockstor version message

### LINKED ISSUE #2115: Re-word "System is running the latest" update page text
Thanks to forum member ednt in the following forum thread for highlighting this issue. It can be confusing for new users, who have as yet not read our docs "Update

Given recent updates / improvements to our  "System -> Software Update" page, which now links back to the docs "Update Channels" section to appreciate that the following message:

> System is running the latest Rockstor version:

pertains to the selected channel. As from 3.9.2-52 this page has been much improved with more concise text, and contextual links back to our docs and the forum; I propose that we edit this now legacy wording by removing the words "the latest" as this is overly contextual and potentially misleading. In any given channel it may be the latest, and this message is shown when no updates are available but does not account for lapsed subscriptions, or a failure to check the repos.

So in short I propose we change the above message to:

> System is running Rockstor version:

for both our currency newer Stable channel and our soon to be newer  (post 'Built on openSUSE' transition) testing channel "Software Update" page displays.

jonpwilson:
Hopefully reduces some confusion.

Fixes #2115

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/update/version_info.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/update/version_info.jst
@@ -137,7 +137,7 @@ $(document).ready(function(){
 {{else}}
 <div class="col-md-offset-3 col-md-6 col-md-offset-3">
   <div id="updateInfo">
-    <h3>System is running the latest Rockstor version: {{currentVersion}}</h3>
+    <h3>System is running Rockstor version: {{currentVersion}}</h3>
     <p>If you've updated recently, reload the browser<b>(Ctrl+Shift+R)</b> for latest UI changes.</p><br>
 
     {{#if autoUpdateEnabled}}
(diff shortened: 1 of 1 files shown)


phillxnet:
@jonpwilson This is super fast work, but I've just noticed another area where we make pretty much the same mistake. During a Rockstor update initiated via the Web-UI the following text is displayed in a popup dialog with a spiny graphic:

> "Rockstor is being updated to the latest version. Please wait..."

sorry I forgot about this similarly misleading text all together when I created the issue. Do you fancy tracking this down and changing it to something more appropriate. Maybe something like:

> "Rockstor is being updated to the latest version in the selected update channel. Please wait..."

Or whatever you think might be more appropriate. Sorry to add this to an already complete pull request but if you make the change in your given branch and push it to GitHub it should appear in this pull request. Don't bother with the usual squashing as it's only going to be 2 commits anyway.

If that doesn't work out, or is a hassle, or you'd rather start fresh then no worries, let me know here,and we can just pop this one in as is and circle back around to the update dialog text change.

jonpwilson:
Good point, I should have spotted that - updated now

jonpwilson:
@phillxnet: Ready for review.

phillxnet:
@jonpwilson Thanks for getting this sorted and for the last minute 'mission creep' :).

Happy to merge. A much appreciated addition to help 'reduce confusion' at least that's the hope.

PR #2039, opened 2019-04-29
Scheduled tasks: add feature to scan for network devices before shutting down

p-betula-pendula:
[Forum entry](https://forum.rockstor.com/t/suspend-shutdown-when-certain-devices-are-offline-using-scheduled-tasks-2038) and issue #2038.

This needs to be looked through because of my lack of experience.

especially:
- check grammar and phrasing in the UI (and code)
- linting
- best practices I'm not aware of
- possible conflicts with other components

It's working flawlessly when tested in my virtual environment.

OpenSUSE is yet to be tested.

PR #2135, opened 2020-02-19
 [openSUSE] fix postfix config re ipv4, tlsmgr, & CA file settings. Fixes #2132

### LINKED ISSUE #2035: [openSUSE] disable config.postfix as we edit main.cf directly
In openSUSE there are sysconfig/SuSEconfig components enabled by default that conflict with our existing direct main.cf editing to enable e-mail notifications, namely entailing the following files:
/etc/sysconfig/mail
/etc/sysconfig/postfix
/usr/sbin/config.postfix
We have our own parser / editor for main.cf and the most direct path to maintain compatibility with our existing users and allow for openSUSE compatibility going forward is to disable this sysconfig mail configuration mechanism via the following entry in /etc/sysconfig/mail:
```
# If you don't want to let SuSEconfig generate your
# configuration file, set this to no
#
MAIL_CREATE_CONFIG="yes"
```
which, if found, should be changed to "no".

### LINKED ISSUE #2132: [openSUSE] basic config settings for Postfix on Rockstor install stage
Config **/etc/postfix/main.cf** need modification from default value `inet_protocols = all` to `inet_protocols = ipv4` in case disabling IPV6 in NetworkManager and error in saving alert mail setting in Rockstor GUI. 
Config **/etc/postfix/master.cf** need modification from commented default value `#tlsmgr unix - - n 1000? 1 tlsmg`  to `tlsmgr unix - - n 1000? 1 tlsmg` in case using SMTP TLS encryption for such services as gmail.com or outlook.com

### LINKED ISSUE #2133: [openSUSE] fix cert file path in alert mail config
Need change **/etc/postfix/main.cf** value form `smtp_tls_CAfile = /etc/ssl/certs/ca-bundle.crt` to `smtp_tls_CAfile = /etc/ssl/ca-bundle.pem` as OpenSUSE uses different cert type and path.

phillxnet:
Thanks to @tyukh for initial reporting and initial diagnosis of the root causes. Due to differing defaults from our prior CentOS base, email notification configuration failed. Adjust configuration mechanism as below.

Fixes #2132
Fixes #2133
Fixes #2035 

# Includes:
- Un-remark tlsmgr line in master.cf when on openSUSE.
- Disable on-by-default mail sysconfig in openSUSE (see issue #2035)
- Update /etc files postfix/master.cf & sysconfig/mail only if required.
- Adapt Certificate Authority file bundle path to openSUSE contextually.
- Enforce "inet_protocols = ipv4" in postfix if a prior 'all' setting is found. This avoids postfix service failure on ipv4 only systems.
- Move to string.format in email_client.py.
- Black format email_client.py.
- Black format system/osi.py (large file) - in separate commit.
- Fix buggy use of os.chmod.
- Abstract "/root/.forward" for clarity.
- Abstract "/etc/postfix/generic" for clarity.
- Improve code comments.
- Add comments to existing system/osi inplace_replace() function.
- Add general purpose replace_line_if_found() facility to system/osi.py.
- Establish simple system/osi replace_line_if_found() mechanism.
- Avoid postfix log warnings by removing to-be duplicate config options.
- Ensure postfix service is enabled after each config save. The vendor default is 'disabled'.

Additional unrelated trivial fix without associated issue:
- A trivial debug log var refactor omission relating to pr #2039 

# Testing
No new unit tests were written. However functional testing was performed on all three target dirtos of "rockstor" (CentOS), and our two openSUSE targets: Leap15.1, and Tumbleweed.

From a clean source build, post postfix removal and install (fresh config files & to establish systemd vendor default status) the Rockstor Web-UI was able to successfully test email notification config entry, prior to submitting it, and successfully submit that config there after and subsequently perform the test email send.

All associated config files were inspected pre and post automated editing and the resulting file access rights were confirmed to be as intended.

All 3 platforms, post email notification configuration, were also confirmed to successfully forward self-sent root-directed email using the following procedure:
```
# cli mail client already installed on our CentOS variants
zypper in mailx
echo "Body text of email alert" | mail -s "Alert Email" root
```

@FroggyFlox Ready for review.
I was unfortunately unable to separate the black formatting in the case of the smaller file edited in this pr, but the much larger system/osi.py file does have it's black formatting in a separate commit to ease review.

### DIFF
--- a/src/rockstor/storageadmin/views/email_client.py
+++ b/src/rockstor/storageadmin/views/email_client.py
@@ -19,46 +19,130 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 import os
 import re
 import json
+import stat
+import distro
 from rest_framework.response import Response
 from django.db import transaction
-from storageadmin.models import (EmailClient, Appliance)
+from storageadmin.models import EmailClient, Appliance
 from storageadmin.serializers import EmailClientSerializer
 from storageadmin.util import handle_exception
 import rest_framework_custom as rfc
-from system.osi import run_command, gethostname
+from system.osi import run_command, gethostname, replace_line_if_found
 from shutil import move
 from tempfile import mkstemp
 from system.services import systemctl
 from system.email_util import send_test_email, test_smtp_auth
 import logging
+
 logger = logging.getLogger(__name__)
 
 
-POSTMAP = '/usr/sbin/postmap'
-HEADER = '####BEGIN: Rockstor section####'
-FOOTER = '####END: Rockstor section####'
-MAIN_CF = '/etc/postfix/main.cf'
+POSTMAP = "/usr/sbin/postmap"
+HEADER = "####BEGIN: Rockstor section####"
+FOOTER = "####END: Rockstor section####"
+MAIN_CF = "/etc/postfix/main.cf"
+MASTER_CF = "/etc/postfix/master.cf"
+POSTFIX = "/usr/sbin/postfix"
+SYSCONFIG_MAIL = "/etc/sysconfig/mail"
+ROOT_FORWARD = "/root/.forward"
+GENERIC = "/etc/postfix/generic"
+
+# List of our used config options within main.cf; used to remove prior entries
+# to avoid log warnings of "... overriding earlier entry: X"
+MAIN_CF_OPTIONS = [
+    "relayhost",
+    "smtp_use_tls",
+    "smtp_sasl_auth_enable",
+    "smtp_sasl_password_maps",
+    "smtp_tls_CAfile",
+    "smtp_sasl_security_options",
+    "smtp_sasl_tls_security_options",
+    "smtp_generic_maps",
+]
 
 
 def rockstor_postfix_config(fo, smtp_server, port, revert):
-    if (revert is True):
+    if revert is True:
         return
-    fo.write('%s\n' % HEADER)
-    fo.write('relayhost = [%s]:%d\n' % (smtp_server, port))
-    fo.write('smtp_use_tls = yes\n')
-    fo.write('smtp_sasl_auth_enable = yes\n')
-    fo.write('smtp_sasl_password_maps = hash:/etc/postfix/sasl_passwd\n')
-    fo.write('smtp_tls_CAfile = /etc/ssl/certs/ca-bundle.crt\n')
-    fo.write('smtp_sasl_security_options = noanonymous\n')
-    fo.write('smtp_sasl_tls_security_options = noanonymous\n')
-    fo.write('smtp_generic_maps = hash:/etc/postfix/generic\n')
-    fo.write('%s\n' % FOOTER)
+    distro_id = distro.id()
+    if distro_id == "rockstor":
+        CAfile_path = "/etc/ssl/certs/ca-bundle.crt"
+    else:  # i.e. openSUSE Leap / Tumbleweed
+        CAfile_path = "/etc/ssl/ca-bundle.pem"
+    # If we move in the future to using yast here note /etc/sysconfig/postfix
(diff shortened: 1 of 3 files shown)


src/rockstor/storageadmin/views/email_client.py
```diff
@@ -19,46 +19,130 @@
...
+    "smtp_sasl_tls_security_options",
+    "smtp_generic_maps",
+]
 
 
 def rockstor_postfix_config(fo, smtp_server, port, revert):
-    if (revert is True):
+    if revert is True:
+        return
+    distro_id = distro.id()
```
FroggyFlox: @phillxnet,
A quick comment before I can test this one out: I noticed you are calling `distro.id()` every time here. In line with our discussion on a different PR related to using variables defined at build time, I actually used `distro_id = settings.OS_DISTRO_ID` instead in my recently submitted PR believing it would be slightly faster and/or more efficient.

Would that actually be the case? The difference is probably rather minor but I wanted to know what you think of it.

phillxnet: @FroggyFlox Re:

> In line with our discussion on a different PR related to using variables defined at build time, I actually used `distro_id = settings.OS_DISTRO_ID` instead in my recently submitted PR believing it would be slightly faster and/or more efficient.

Yes, good point, I'd actually forgotten about that; however as you say it's a fairly trivial time saver in this case as it's only going to be called once/twice during each email setup, so maybe no more than a handful of times during an installs life time. My original intent with the settings.OS_DISTRO_ID was to save the "Built on ..." / "Uses ..." info and to avoid the repeated calls that are inevitable with it's use in the Web-UI header. I didn't want the Web-UI refresh to be calling the distro library at every refresh interval, though that's only something like every 10 to 30 minutes it's far more pressing a concern than what we have here.

So in short I'd suggest we absolutely use the settings var for frequent calls where we can, i.e. where Leap to Tumbleweed and back don't matter. But my most recent significant use of this library was in the packaging area where we have to account for folks moving their install from Leap to Tumbleweed and back, as they have different repositories. And that was more fresh than the Web-UI header work. Plus the header then denotes what the rpm was built on, or that was the intention, giving added debug value to a screen shot.

I say we leave this as is for now as it's 'mostly harmless' and have more of a talk on the best way to go in the future. In most cases the Django settings is probably best, so in turn I say we also leave that 'as-is' in your current pull request. But we should very definitely avoid frequent calls to distro if we can, and the setting var gives us that option. Plus there may be a related memory trade-off in pulling in the settings but I'm just not up enough of Django workings to know. So swings and roundabouts and I say we go with what we have both used thus far but definitely avoid, if possible, repeated calls. You samba config calls are also likely to only be used once per config, which is not that often. And if there turns out to be a difference, say between Leap / Tumbleweed and folks move there system under us then we will miss it, at least until a new rpm is pulled down. But that is unlikely currently. So yes lets chat on this but get these two pull requests sorted if we can in the mean time.

Nicely spotted though, and brings to the fore the question of how we use this. It's all a bit messy really but I couldn't thing of another way around it while we are in this transition. We might consider ripping out all uses of this now oft used library once we are done with the CentOS target all together.

Incidentally in the recent package work I also tried to call a minimum number of times by calling early and passing the result to the multiple child processes. I had though of doing that here but for a single call per email setup it seemed like unnecessary complexity. I should probably have noted this in the pr.

Thanks for bringing this up.

FroggyFlox: Thanks a lot for the very detailed answer!

phillxnet:
@FroggyFlox Thanks for the review, much appreciated.

Glad you tested with gmail as I didn't, but I did use another TLS  email service (with the 587 port) and that worked a treat as well. I'll go ahead and merge this one soon then. Thanks for submitting the preexisting table typo fix, I'd completely missed it.

PR #2142, opened 2020-03-02
[openSUSE] adapt to shellinabox package differences. Fixes #2006

### LINKED ISSUE #2006: [openSUSE] shellinabox
`shellinabox` will need to be packaged. However:

1. It's most current release is a bit over two years old,
2. It uses openssl v1.0 (Tumbleweed defaults to 1.1)
3. Upstream doesn't include a systemd unit

So I'm wondering if there are better alternatives - but these must be written in systems languages, not for eg, python or node-js, these are just far too heavy-weight.

Fedora package source is [here](https://src.fedoraproject.org/cgit/rpms/shellinabox.git/tree/)

At this point I also suggest creating an _official_ rockstor account on [openSUSE OBS](https://build.opensuse.org/) to ease package building and supply, you should be able to get packages merged to official channels if required also.

phillxnet:
A recently available OBS shells.repo hosted shellinabox package had breaking differences from our prior epel CentOS pacakge. The following code changes were made to support both packages concurrently ~~(see testing section for pending changes to the OBS package)~~:

Fixes #2006 

- Move to re-configure service on every service enable and black format the associated file: shellinaboxd_service.py, with minor refactoring.
- Remove build time 'initial' config file instantiation, along with the prior listed change this establishes a single point, within code, where our desired configuration is written. This also ensures we overwrite package default config prior to initial service start.
- Establish distro specific configuration options: the proposed openSUSE shells repo package has a different certificates location.
- Go with both package defaults re user/group for running the shellinabox deamon.
- Account for differing default configuration file names.
- Account for differing systemd service file names by normalising on our existing 'shellinaboxd' and establishing this if not found during the initrock.py script run by rockstor-pre.service.
- Improve comments in prep_db.py pertaining to shellinaboxd + black format.
- Remove unused status() and associated import from shell.py.
- Pass systemd service name to shell.py to reduce hardcoded instances.
- Move shell.py to string.format + black formatting.
- Fix existing bug where shellinabox config file options had no closing quote and carriage return.

@FroggyFlox Ready for review.
Apologies as I didn't manage to split out all of the black formatting, but the largest file was blacked in it's own commit. Not all files were blacked but bit by bit.

## Testing:
In our openSUSE targets I used a slightly modified version of the indicated OBS shell.repo shellinabox package ~~that carries two proposed, and submitted for inclusion upstream, systemd service file fixes by this pr's author:
https://build.opensuse.org/package/show/home:rockstor:branches:shells/shellinabox~~
EDIT: The proposed / submitted upstream changes have now been accepted upstream. As part of this edit I have update the indicated repos to reference the upstream one that now incorporates the referenced systemd servcie file fixes.

This package was chosen via research by @FroggyFlox and myself as it had recently been adopted and moved into the official OBS shells.repo repository. It also contains fixes for shortfalls of a few other existing shellinabox packages available in other individual OBS repos, ie no system, firewalld, and openssl v1.1 compatibility. All of which we require.

### Leap15.1:
```
zypper addrepo --refresh -p105 http://download.opensuse.org/repositories/shells/openSUSE_Leap_15.1/ shells
```
### Tumbleweed:
```
zypper addrepo --refresh -p105 http://download.opensuse.org/repositories/shells/openSUSE_Tumbleweed/ shells
```
with subsequent:
```
zypper refresh
(trust temporarily or always)
```
A lower priority than default (99) is suggested to allow for default priority override (edit) and to avoid other shell entries within this repo from superseding default packages.  ~~if this package enters the standard repositories; although a vendor change would likely also be required.~~

Functionally tested to work as expected on a 'rockstor' CentOS base and the openSUSE Leap15.1, & Tubleweed distributions, all via a from scratch source build of rockstor-core.

## Caveates:
1. Under certain timings the initial invocation of the shell when started from the Web-UI header link and it's own page System - System Shell results in "502 Bad Gateway"
To reproduce:
When the service is off and is enabled via:
- System - System Shell
- or Web-UI header “System Shell”
Work around = Web Browser Page refresh (does not work with popup window option).

2. Popup window mode flaky.
When “Open shell in a popup windows” config option is selected (non default):
- Works but is a little flaky as needs a master window refresh to re-enable popup window invocation once one is closed.

These 2 caveats were observed in Leap15.1, and have been seen in our to-be legacy CentOS base but have yet to be seen in the Tumbleweed instance; but are believed to be generic to the base packages' function or our general implementation and were considered out side the realm of this pull request.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -24,7 +24,6 @@ parts =
 #      postgres-conf
       gunicorn
       nginx-conf
-      shellinabox-conf
       stop-shellinabox
       django-settings-conf
       test-settings-conf
@@ -59,11 +58,6 @@ command =
 recipe = collective.recipe.template
 input = ${buildout:directory}/conf/nginx.conf.in
 output = ${buildout:directory}/etc/nginx/nginx.conf
-
-[shellinabox-conf]
-recipe = collective.recipe.template
-input = ${buildout:directory}/conf/shellinaboxd.in
-output = /etc/sysconfig/shellinaboxd
 
 [stop-shellinabox]
 recipe = plone.recipe.command
--- a/conf/settings.conf.in
+++ b/conf/settings.conf.in
@@ -385,12 +385,20 @@ NUT_LISTEN_ON_IP = '0.0.0.0'
 NUT_SYSTEM_SHUTDOWNCMD = '/sbin/shutdown -h +0'
 
 # Shell In A Box base settings
-SHELLINABOX = {
-	'user': 'root',
-	'group': 'root',
-	'port': '4200',
-	'certs': '/var/lib/shellinabox'
-}
+if distro.id() == 'rockstor':
+    SHELLINABOX = {
+        'user': 'shellinabox',
+        'group': 'shellinabox',
+        'port': '4200',
+        'certs': '/var/lib/shellinabox'
+    }
+else:
+    SHELLINABOX = {
+        'user': 'shellinabox',
+        'group': 'shellinabox',
+        'port': '4200',
+        'certs': '/etc/shellinabox/certs'
+    }
 
 UPDATE_CHANNELS = {
 		'stable': {
--- a/conf/shellinaboxd.in
+++ b/conf/shellinaboxd.in
@@ -1,7 +1,0 @@
-# Shell In A Box configured by Rockstor
-
-USER=root
-GROUP=root
-CERTDIR=/var/lib/shellinabox
-PORT=4200
-OPTS="--no-beep --localhost-only --disable-ssl -s /:LOGIN --css white-on-black.css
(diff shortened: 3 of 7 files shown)


FroggyFlox:
Hi @phillxnet,

I just tested on a clean new Leap 15.1 VM and it worked like a charm.

I first added the new official repo, refreshed it, installed shellinabox (`zypper in shellinabox`), and then proceeded with building Rockstor as usual.

The UI behaved like it currently does in legacy Rockstor, and all seems to be working on my end!

Thanks a lot for fixing this!

phillxnet:
@FroggyFlox Thanks for the review, much appreciated. I'll merge this shortly ready for the next release. I just need to make a little tweak to our buildbot to accommodate the additional new repo and test this change prior to merging.

PR #2147, opened 2020-03-15
update known non unique product uuids. Fixes #2143

### LINKED ISSUE #2143: add known non unique product_uuid on HP MicroServer Gen8
Thanks to David via support email for helping to diagnose another now known non unique product_uuid on an HP / HPE ProLiant MicroServer Gen8 appliance.

```
31393138-3538-5A43-3135-353130323750
```

It is proposed that we add this to our list of known fake_puuids within:
https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/system/osi.py
so that future ISO installers can side-step this motherboard's non unique firmware derived uuid and create a properly unique Appliance ID in it's stead.

This issue should be addressed prior to the next ISO release to avoid a repeat of the associated inconvenience for others using this popular device.

phillxnet:
Includes 3 additional non unique product_uuids reported by Appman but only one with user follow-up and so motherboard / model info.

Fixes #2143

Summary:
- Adds 3 additional fake product uuids.
- Move associated Python file to string.format.
- Re-apply black formatting.

@FroggyFlox Ready for review
I've split off the string.format and minor re-black formatting into their own shared commit to ease review.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -248,7 +248,7 @@ def scan_disks(min_size, test_mode=False):
                 cur_val = cur_val + sl[i]
                 i = i + 1
             else:
-                raise Exception("Failed to parse lsblk output: %s" % sl)
+                raise Exception("Failed to parse lsblk output: {}".format(sl))
         # md devices, such as mdadmin software raid and some hardware raid
         # block devices show up in lsblk's output multiple times with identical
         # info.  Given we only need one copy of this info we remove duplicate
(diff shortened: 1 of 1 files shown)


FroggyFlox:
@phillxnet , thanks for this PR!

I just tested and could build it on a JeOS Leap15.1 install without problem and couldn't notice any problem in the webUI afterwards. Given the very minor change, I believe it should be OK.

Thanks for taking care of the string.format bit... I tested a few os operations from the webUI (scrub, balance) and everything went as expected and provided the expected output in the logs so I don't see any syntax error introduced by all that!

Good for me.

phillxnet:
@FroggyFlox Thanks for the review and build test. I'll get this one merged soon then.

PR #2152, opened 2020-04-19
improve password handling during config restore. Fixes #2151

### LINKED ISSUE #2151: improve password handling during config restore
Within the configuration restore mechanism a fresh password is generated for each user saved in the configuration backup that does not already existing on the target system. The associated logging of this restore process includes these generated passwords. It is proposed that we remove this element of the restore logging process and improve our password generation process to be more in-keeping with accepted security practices.

phillxnet:
## See the following issue text for context.

Fixes #2151

Includes:
- Remove password elements from all config restore payloads prior to logging success. Failed component restores still log the full payload that lead to each failure.
- Add check for dict payload prior to dict.pop password & logs if not.
- Improve default password generation for users within a configuration backup that do not exist on the target system.
- Add comment on potential improvements in random string generation post python3 migration.
- Move to string.format within the file altered during this commit/pr.
- Apply black formatting on the file altered during this commit/pr.

### Testing.

A sample config backup containing a regular and a Web-UI enabled user and was restored to a system that did not have these users already. The users were successfully restored as expected. Subsequent Web-UI initiated password resets resulted in both users having the designated passwords as expected. Log entries of successful restores of these users did not indicate any password elements. Subsequent restore of the same config, where the users already existed, did show, as intended, the full payload of the failed user creation (as those users had already been restored).

### DIFF
--- a/src/rockstor/storageadmin/views/config_backup.py
+++ b/src/rockstor/storageadmin/views/config_backup.py
@@ -20,6 +20,8 @@ import gzip
 import json
 import logging
 import os
+import string
+import random  # Post Python3 move, replace random with: openid.cryptutil randomString
 from time import sleep
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox Merging this as-is as the functional changes are pretty simple and a restore test of both types of users looks to have completed OK. It will also ease my working on the testing issue as it's now down to just the test_users.py file.

Issue #2156, opened 2020-04-27
[NG] ROOT pool user share not displayed on page refresh


phillxnet:
When running on a system with boot to snapshots enabled, default when sufficiently large system disk is used (>19 GB or so), user created shares on this ROOT labelled pool initially appear as expected, but post page refresh are then no longer known to the Rockstor Web-UI. This does not affect non boot to snapshot configurations and is thus suspected as relating to a bug in our treatment of user created subvol paths on the system disk when that disks ROOT pool is mounted in boot to snapshot mode.

Our "Built on openSUSE testing channel live (early-adopters/developers only)" forum post:
https://forum.rockstor.com/t/built-on-opensuse-testing-channel-live-early-adopters-developers-only/6592
has 'featured' this issue as an example of a lack of parity between our current CentOS based Stable channel offering and our current testing channel 'Built on openSUSE' endeavour. This issue is therefore intended as a focus for work to address/resolve this disparity.

phillxnet:
## Boot to snapshot config enabled (Leap15.2beta)
ROOT pool user share (+ default 'home') as initially created and displayed before page refresh.
As displayed in Shares overview table:
```
rw,relatime,space_cache,subvolid=260,subvol=/@/home 
rw,relatime,space_cache,subvolid=292,subvol=/@/.snapshots/1/snapshot/test_share_01
```
Note the Pool (Active mount options / Status) Quotas column for both contains:
```
ROOT (rw,relatime,space_cache,subvolid=259,subvol=/@/.snapshots/1/snapshot) Enabled 
```
## Non boot to snapshot config (Leap15.2beta)
Here we used a small 16GB system disk and so boot to snapshots was auto disabled during install. Same single additional user share created as in above boot to snapshot config.
As displayed in Shares overview table:
```
rw,relatime,space_cache,subvolid=264,subvol=/@/home
rw,relatime,space_cache,subvolid=270,subvol=/@/test_share_01
```
Note the Pool (Active mount options / Status) Quotas column for both contains:
```
ROOT (rw,relatime,space_cache,subvolid=256,subvol=/@) Enabled 
```

phillxnet:
And with debug mode on (default with source build) via:
```
/opt/rockstor/bin/debug-mode on
```
During a shares page refresh we have:
## Boot to snapshot config enabled (Leap15.2beta)
```
[...]

PR #2159, opened 2020-05-04
inadvertent use of internal -1/-1 pool quote group flag. Fixes #2158

### LINKED ISSUE #2158: inadvertent use of internal -1/-1 pool quote group flag
Internally a pqgroup, pool quota group, of -1/-1 is used as a place holder for pool quotas disabled state or where a new quota group cannot otherwise be created, i.e. on a read only file system. In some 'corner cases' this internal value 'flag' is being used, inadvertently, in actual qgroup assignments. E.g.:
```
Houston, we've had a problem.
Error running a command. cmd = /usr/sbin/btrfs qgroup assign 0/343 -1/-1 /mnt2/pre-rock-pool. rc = 255. stdout = ['']. stderr = [''] 
```
This issue's author has only, as yet, observed this behaviour in our 'Built on openSUSE' testing channel incarnations but it is though to similarly affect our current Stable Channel offerings. A suggestion of this is evident in a recent report by forum member Stefan in the following post on a thought to be unrelated replication failure:

https://forum.rockstor.com/t/replication-failing/7047

Reproducer:
- Create a new share (subvol) on either the system pool or a data pool; both are similarly affected.
- Clone the above share.
- Turn quotas for the associated pool off.
- Turn quotas for the associated pool on and rapidly perform the next step.
- Take a snapshot of the above clone share (subvol) while quotas are in 'indeterminate state'.

This reproducer is highly timing, and existing snapshot number, dependant but has been used effectively, with intermittently success, on both Leap15.1 and Leap15.2beta to reproduce the issue.

The request snapshot is created successfully but without the initially expected pool quota group assignment. This assignment may later be corrected by additional mechanism but is outside this issues remit.

phillxnet:
Correct programming error re misuse of identity test where equality test should have been used.

When comparing our internal flag of pool quota group value of "-1/-1" the wrong test was applied. This resulted in inadvertent execution of "btrfs qgroup assign" with target qgroups of the flag value; as detailed in the issue text.

Fixes #2158 

Please see linked issue text for more context.

### DIFF
--- a/src/rockstor/storageadmin/views/clone_helpers.py
+++ b/src/rockstor/storageadmin/views/clone_helpers.py
@@ -133,7 +133,7 @@ def create_clone(share, new_name, request, logger, snapshot=None):
         new_share = Share(pool=share.pool, qgroup=qgroup_id, pqgroup=pqid,
                           name=new_name, size=share.size, subvol_name=new_name)
         new_share.save()
-        if pqid is not PQGROUP_DEFAULT:
+        if pqid != PQGROUP_DEFAULT:
             update_quota(new_share.pool, pqid, new_share.size * 1024)
             share_pqgroup_assign(pqid, new_share)
         # Mount our new clone share.
--- a/src/rockstor/storageadmin/views/share.py
+++ b/src/rockstor/storageadmin/views/share.py
@@ -186,7 +186,7 @@ class ShareListView(ShareMixin, rfc.GenericView):
             # The following pool.save() was informed by test_share.py
             pool.save()
             s.save()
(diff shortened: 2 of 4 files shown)


phillxnet:
@FroggyFlox As this pr is a simple swap out of "is not" identity test for "!-" equality test for the same comparison pqgroup test project wide I'm hoping you are OK with me merging directly as I need this fix in place to re-basing my ongoing #2156 related code changes onto for final testing prior to pull request submission.

FroggyFlox:
@phillxnet,
Of course!

PR #2167, opened 2020-05-10
Update libraries needed for Rsync over SFTP

### LINKED ISSUE #2155: libpopt0 not in /lib64 on openSUSE
As reported in https://forum.rockstor.com/t/can-not-create-an-sftp-share-on-opensuse/7026

It appears libpopt0 is not in /lib64 and instead in /usr/lib64 on openSUSE. This leads to a failure when creating SFTP shares.

The offending code is at https://github.com/rockstor/rockstor-core/blob/dc98a1cc59587cd9b5225da7cad0556cb84781e1/src/rockstor/system/ssh.py#L123-L127

When I checked on my system _only_ libpopt0 is in /usr/lib64, all the other libs are still in /lib64, as such a special case for libpopt should likely be introduced.

FroggyFlox:
Fixes #2155 
@phillxnet, ready for review.

As described by @freaktechnik in #2155, we currently cannot export a share via SFTP in Rockstor built on openSUSE due to out-of-date shared libraries. This pull request thus proposes to update the shared libraries required by `/bin/bash` and `/usr/bin/rsync` in both Leap15.2 beta and Tumbleweed.

The libraries were determined as described in https://github.com/rockstor/rockstor-core/issues/2155#issuecomment-626228322.

## Logic
We retain the same logic as before:
1. the shared libraries required in each system are hardcoded in `rsync_for_sftp()`
2. we copy these over to the chroot directory defined in `django.settings`

The only difference is that we now store these required libraries in a dictionary keyed by distribution id. We can thus simply refer to the `OS_DISTRO_ID` defined at build time to loop into the correct list of libs and copy each one over to the chroot.

## Caveats
Especially true in Tumbleweed, this approach is limited by the fact that we have the libraries hard-coded (this is actually the reason why Issue #2155 occurred). We could dynamically fetch those by writing some sort of wrapper around `ldd` for instance, but that would also add another system call each time we need to refresh our chroot dir setup. @phillxnet , do you think that would be worth the cost? Parsing all these libs and their respective paths might end up cost a little in terms of resources, and thus I'm not sure it would be worth it. Thanks for letting me know if you think otherwise and I'll try to work on that.

### Rockstor built on Leap 15.2
In the webUI, export the share `sftp_test` via SFTP and verify the correct directory structure was created:
*logs*
```
[10/May/2020 10:42:12] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/bin
[10/May/2020 10:42:12] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/usr/bin
[10/May/2020 10:42:12] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/lib64
[10/May/2020 10:42:12] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/usr/lib64
[10/May/2020 10:42:12] DEBUG [system.osi:157] Running command: /usr/sbin/usermod -s /bin/bash test
[10/May/2020 10:42:12] DEBUG [system.osi:157] Running command: /usr/bin/systemctl reload sshd
```
and structure:
```
rockdev:~ # tree /mnt3/
/mnt3/
└── test
    ├── bin
    │   └── bash
    ├── lib64
    │   ├── ld-linux-x86-64.so.2
    │   ├── libacl.so.1
    │   ├── libattr.so.1
    │   ├── libc.so.6
    │   ├── libdl.so.2
    │   ├── libpthread.so.0
    │   ├── libreadline.so.7
    │   ├── libtinfo.so.6
    │   └── libz.so.1
    ├── sftp_test
    └── usr
        ├── bin
        │   └── rsync
        └── lib64
            ├── libcrypto.so.1.1
            ├── libpopt.so.0
            └── libslp.so.1

7 directories, 14 files
```

### Rockstor built on Tumbleweed (20200501)
In the webUI, export the share `sftp_test` via SFTP and verify the correct directory structure was created:
*logs*
```
[10/May/2020 10:58:19] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/bin
[10/May/2020 10:58:19] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/usr/bin
[10/May/2020 10:58:19] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/lib64
[10/May/2020 10:58:19] DEBUG [system.osi:157] Running command: /usr/bin/mkdir -p /mnt3/test/usr/lib64
[10/May/2020 10:58:19] DEBUG [system.osi:157] Running command: /usr/sbin/usermod -s /bin/bash test
[10/May/2020 10:58:19] DEBUG [system.osi:157] Running command: /usr/bin/systemctl reload sshd
```
and structure:
```
rockdev:~ # tree /mnt3/
/mnt3/
└── test
    ├── bin
    │   └── bash
    ├── lib64
    │   ├── ld-linux-x86-64.so.2
    │   ├── libc.so.6
    │   ├── libdl.so.2
    │   ├── libpthread.so.0
    │   ├── libreadline.so.8
    │   ├── libtinfo.so.6
    │   └── libz.so.1
    ├── sftp_test
    └── usr
        ├── bin
        │   └── rsync
        └── lib64
            ├── libacl.so.1
            ├── libcrypto.so.1.1
            ├── libpopt.so.0
            └── libslp.so.1

7 directories, 13 files
```

## Unit tests
Unit tests outputs are identical to current master (not shown).

### DIFF
--- a/src/rockstor/system/ssh.py
+++ b/src/rockstor/system/ssh.py
@@ -17,7 +17,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
 import re
-from shutil import move
+from shutil import move, copy
 from tempfile import mkstemp
 from services import systemctl
 from system.osi import run_command
@@ -113,19 +113,56 @@ def sftp_mount(share, mnt_prefix, sftp_mnt_prefix, mnt_map, editable='rw'):
 
 def rsync_for_sftp(chroot_loc):
     user = chroot_loc.split('/')[-1]
-    run_command([MKDIR, '-p', ('%s/bin' % chroot_loc)])
-    run_command([MKDIR, '-p', ('%s/usr/bin' % chroot_loc)])
-    run_command([MKDIR, '-p', ('%s/lib64' % chroot_loc)])
+    run_command([MKDIR, '-p', '{}/bin'.format(chroot_loc)], log=True)
+    run_command([MKDIR, '-p', '{}/usr/bin'.format(chroot_loc)], log=True)
+    run_command([MKDIR, '-p', '{}/lib64'.format(chroot_loc)], log=True)
+    run_command([MKDIR, '-p', '{}/usr/lib64'.format(chroot_loc)], log=True)
 
-    import shutil
-    shutil.copy('/bin/bash', ('%s/bin' % chroot_loc))
-    shutil.copy('/usr/bin/rsync', ('%s/usr/bin' % chroot_loc))
-    libs = ('ld-linux-x86-64.so.2', 'libacl.so.1', 'libattr.so.1',
-            'libc.so.6', 'libdl.so.2', 'libpopt.so.0', 'libtinfo.so.5',)
-    for l in libs:
-        shutil.copy('/lib64/%s' % l,
-                    ('%s/lib64' % chroot_loc))
-    run_command([USERMOD, '-s', '/bin/bash', user])
+    copy('/bin/bash', '{}/bin'.format(chroot_loc))
+    copy('/usr/bin/rsync', '{}/usr/bin'.format(chroot_loc))
+
+    libs_d = {
+        "rockstor": [
+            '/lib64/ld-linux-x86-64.so.2',
+            '/lib64/libacl.so.1',
+            '/lib64/libattr.so.1',
+            '/lib64/libc.so.6',
+            '/lib64/libdl.so.2',
+            '/lib64/libpopt.so.0',
+            '/lib64/libtinfo.so.5',
+        ],
+        "opensuse-leap": [
+            "/lib64/libacl.so.1",
+            "/lib64/libz.so.1",
+            "/usr/lib64/libpopt.so.0",
+            "/usr/lib64/libslp.so.1",
+            "/lib64/libc.so.6",
+            "/lib64/libattr.so.1",
+            "/usr/lib64/libcrypto.so.1.1",
+            "/lib64/libpthread.so.0",
+            "/lib64/ld-linux-x86-64.so.2",
+            "/lib64/libdl.so.2",
+            "/lib64/libreadline.so.7",
+            "/lib64/libtinfo.so.6",
+        ],
+        "opensuse-tumbleweed": [
+            "/lib64/libc.so.6",
+            "/usr/lib64/libacl.so.1",
+            "/lib64/libz.so.1",
+            "/usr/lib64/libpopt.so.0",
+            "/usr/lib64/libslp.so.1",
+            "/lib64/ld-linux-x86-64.so.2",
+            "/usr/lib64/libcrypto.so.1.1",
+            "/lib64/libpthread.so.0",
+            "/lib64/libdl.so.2",
+            "/lib64/libreadline.so.8",
+            "/lib64/libtinfo.so.6",
+        ]
+    }
+
+    for l in libs_d[settings.OS_DISTRO_ID]:
+        copy(l, "{}{}".format(chroot_loc, l))
+    run_command([USERMOD, '-s', '/bin/bash', user], log=True)
 
 
 def is_pub_key(key):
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox Thanks for seeing to this so quickly.

Just had a quick look and this seems like a nice solution. I'll have a proper look soon.

Re:
> We could dynamically fetch those by writing some sort of wrapper around `ldd` for instance, but that would also add another system call each time we need to refresh our chroot dir setup.

Not necessarily. If we assume this is not likely to change that often we could run something in initrock that created a json file that had the info required from ldd. Either pre or post processed. And then with a single file read, with json to dict conversion, know our library location. We could even go crazy and have a db table, that is similarly populated on boot, which in turn could be augmented by a Web-UI to allow for custom library additions. But more seriously; I think what you've done here is perfect as a like for like fix on what we have now.

> @phillxnet , do you think that would be worth the cost?

No, I think I'm in agreement with yourself here. Lets keep this simple for now, it can always be revisited later if need be.

Thanks again for stepping up to this on. I'll take a proper look shortly. And thanks for helping with the review by separating out the string.format() and black associated changes.

phillxnet:
@FroggyFlox Apologies for the delay. I've just had another look at this and it works for me as you described on Legacy CentOS (rockstor), Leap15.1, Leap15.2 beta, and Tumbleweed. After the ssh service cycle currently indicated as required in:

"SFTP server runs unconfigured by default #2168"

### SFTP
I was able to gain access via prior configuration as described in: http://rockstor.com/docs/sftp/sftp.html
via:
```
sftp sftp@leap15-2.lan
```
where sft was the use setup to own the sftp exported share. And further to this the resulting exported share and it's contents were displayed as expected:
```
sftp> ls
bin        lib64      sftp-test  usr   
```
where sftp-test is the share name exported from within the Rockstor Web-UI.
```
sftp> cd sftp-test/
sftp> ls
testfile
sftp> exit
```
Where "testfile" was the result of:
```
touch /mnt2/sftp-test/testfile
```
via a ssh session as root user.

### Rsync over ssh with configured user and share:
user: sftp, share sftp-test
```
rsync -avhe ssh 9999.patch sftp@leap15-2.lan:/sftp-test
Password:                                                                                                                                                                   
sending incremental file list                                                                                                                                               
9999.patch                                                                                                                                                                  
                                                                                                                                                                            
sent 119 bytes  received 35 bytes  44.00 bytes/sec                                                                                                                          
total size is 9  speedup is 0.06
```
With the resulting local 999.patch file then found to have been successfully located on the remote Rockstor system:
```
leap15-2:~ # find / -name 9999.patch
...
/mnt2/rock-pool/sftp-test/9999.patch
/mnt2/sftp-test/9999.patch
/mnt3/sftp/sftp-test/9999.patch
```
In order these are the pool mount, share mount, and sftp-user/sftp-test-share-name mount points.

The resulting rsynced file was then confirmed as owned by the sftp user:
```
leap15-2:/home/sftp # ls -la /mnt2/sftp-test/
total 4
drwxr-xr-x 1 sftp root   36 May 18 20:00 .
drwxr-xr-x 1 root root  346 May 18 18:50 ..
-rw-r--r-- 1 sftp users   9 Oct 21  2019 9999.patch
-rw-r--r-- 1 root root    0 May 18 19:18 testfile
```
The same tests were performed successfully using Rockstor source builds on Legacy CentOS (rockstor), Leap15.1, Leap15.2 beta, and Tumbleweed.

@FroggyFlox Thanks again for stepping up to sort out this issue. Nicely done. Your diligence re the pr text is most appreciated.

We have both left an string.format() un-done on this pass but I'm eager to get this in without further delay. We have plenty of these left project wide so we can always see to this on our next pass re string.format() changes elsewhere.

And thanks also to @freaktechnik for the detailed report referenced.

Again apologies for taking so long to get this merged.

PR #2179, opened 2020-06-03
Delete 90-default.preset. Fixes #2178

### LINKED ISSUE #2178: [openSUSE] upstream defaults for systemd-presets are ignored
We currently use a dedicated systemd-presets file (https://github.com/rockstor/rockstor-core/blob/master/conf/90-default.preset) in the high priority location `/etc/systemd/system-preset/` to define defaults enabled/disabled state for all systemd services.

Notably, this file ends with `disable *`, which sets a default disabled state for all services not defined in this file, and thus bypasses all defaults defined by upstream. 

As the openSUSE wiki reads:  
[https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Enabling_systemd_unit_files](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Enabling_systemd_unit_files)

> systemd-presets-common-SUSE is intended for services that are mandatory for a proper function of the system or the package. Placing preset into this file means: All systems should use this value as a default. This package is a common base for both open and enterprise products and it is part of all base installations. 

> systemd-presets-branding-openSUSE is indented for vendor customization. Placing a preset here means: It is a wise default for openSUSE, but not enabling this service in other product makes perfect sense. The package structure makes possible to create a custom brand with a custom systemd-presets-branding-brand as a part of custom branding. Generic rules for branding packages are explained in openSUSE:Packaging_Branding.

Based on this, it would seem the recommended way would be to use our own systemd-presets-branding-rockstor package instead.

FroggyFlox:
Fixes #2178.
@phillxnet , ready for review.

As detailed in #2178, it is proposed to remove our custom default systemd-presets and rely instead on our systemd-presets-branding-rockstor package.

Notably, this allows to account for and rely on upstream's default settings as well.


phillxnet:
@FroggyFlox As per our discussion I'm merging this removal of the presets file as it not longer plays a part in our system config. Thanks for also seeing to the other elements of this system preset config management, ie within our in-development systemd-presets-branding-rockstor OBS package.

> Notably, this allows to account for and rely on upstream's default settings as well.

Thanks for the clarification on this one. Much appreciated. And important as we try to become that bit less 'custom distro' and reply on the greater expertise of upstream where ever we can.

PR #2184, opened 2020-06-17
failure to remove detached disks via pool resize. Fixes #2099

### LINKED ISSUE #2077: remove/resize pool for detached single dev pool fails
Thanks to forum member Zonk for highlighting this issue. When a single disk pool suffers disk failure the suggested, via bubble, "Pool remove/resize" procedure to remove the associated pool / disk it fails with:

```
File “/opt/rockstor/eggs/gunicorn-19.7.1-py2.7.egg/gunicorn/workers/sync.py”, line 68, in run_for_one
self.accept(listener)
File “/opt/rockstor/eggs/gunicorn-19.7.1-py2.7.egg/gunicorn/workers/sync.py”, line 27, in accept
client, addr = listener.accept()
File “/usr/lib64/python2.7/socket.py”, line 202, in accept
```

Please update the following forum thread with this issue resolution:
https://forum.rockstor.com/t/cannot-delete-remove-a-failed-disk/6453

### LINKED ISSUE #2099: failure to remove detached disks via pool resize
Thanks to form member Azzazel for highlighting this issue. Post our recent disk management code revamp in 3.9.2-49 we fail to deal appropriately with detached prior pool members. The sighted forum thread details a reproducer but it is believed to be a generic problem with detached disk removal.

This has been demonstrated on 3.9.2-50.

The observed behaviour is a pool resize detached disk removal operation silently failing; leaving the requested detached disks in place.

https://forum.rockstor.com/t/cannot-remove-detached-disks/6572/8

### LINKED ISSUE #2182: usage displayed instead of available free in disk removal check
We have a cosmetic bug where usage is displayed where available free space is indicated/intended. This can lead to confusing reports such as:

> Removing disks ([u'detached-64e525c4fe1b4437ad2c925d291643cb']) may shrink the pool by 548864 KB, which is greater than available free space 4702668 KB. This is not supported. 

The disk removal blocker calculation is actually as intended but we simply display the wrong value.
I.e.:
```
if size_cut >= (pool.size - usage):
                    e_msg = ('Removing disks ({}) may shrink the pool by '
                             '{} KB, which is greater than available free '
                             'space {} KB. This is '
                             'not supported.').format(dnames, size_cut, usage)
```
Where available free is (pool.size - usage), not usage as is currently displayed.

phillxnet:
Fixes #2099 
"failure to remove detached disks via pool resize"
Fixes #2077
"remove/resize pool for detached single dev pool fails"

Additional unrelated trivial fix included:
Fixes #2182
"Usage displayed instead of available free space Issue" #2182

## Detached disk management changes proposed:
- When a pool has no missing members remove db held pool associations from all prior ‘detached’ members. These disks then show up as regular detached devices and can be removed by their ‘bin’ icons. This allows for continued surfacing of prior attached status for these devices while returning source of truth to the pool itself re detached members. Issue #2099
- Special-case the removal of a detached device from a pool that has only detached members. Issue #2077
- Restrict detached device removal selection: one-at-a-time and with no detached/attached mixing. This reduces our problem space and so eases sanity checking. Attempted multi detached device selection informs that all detached will be removed on pools with only detached members anyway.
- Improve error message and log entry on missing device removal failure: reiterating degraded,rw advise and reboot suggestion.

## Note on interplay between the first two elements of above.
For pools with only detached members, we preserve the function of maintaining the members pool association (in light of the first element above), even though we normally assess missing device status via a pool model property (has_missing_dev) which can’t in the case of no attached members be known. This functionality is due to a prior check where, in the case of zero attached members, we proceed no further as no mount is then possible, and so do not inadvertently remove that pools single disk association. This maintains the possibility to detach all of a pools members and then re-attach them ‘hole sale’ (in powered off state) and for all associated configs (shares, exports, etc) to be re-established automatically. As if we were to remove all disk pool associations automatically in the case of all detached members, all it’s associated shares/exports would also be removed. (TEST THIS).

## Testing:

Changing all serials for all drives, during power down in KVM, results in none of the ‘detached’ former members having a pool association. This means these prior detached pool associated members are then removable via their bin icon, and as such we still surface their prior attached status. #2099

Removing a single detached member from pools that have only detached members results, by design, in all detached members having their pool association removed, and on next pool state refresh, the pool itself being removed. See #2077 detailed caveat below.

## Caveats:
N.B. as is common on a few of our Web-UI ‘operations’ two pool details page refreshes are often required for final pool state to be displayed. I.e. a removed detached disk entry will linger, post the completion of the operation, until this double refresh has been performed. And in the case of detached disk removal from pools that have only detached member, we have the ‘Ragged Edge’ issue detailed in my comments on issue #2077 which can be addressed in a future pull request.

Ready for review, requesting @FroggyFlox .
I have separated out the various main elements, as detailed above, into their own commits to aid review and have also ensured our project goals of string.format() and black formatting to all modified files.

### DIFF
--- a/src/rockstor/storageadmin/views/command.py
+++ b/src/rockstor/storageadmin/views/command.py
@@ -90,6 +90,16 @@ class CommandView(DiskMixin, NFSExportMixin, APIView):
                     "are no attached devices. Moving on.".format(p.name)
                 )
                 continue
+            # If pool has no missing remove all detached disk pool associations.
+            # Accounts for 'end of run' clean-up in removing a detached disk and for cli
+            # maintenance re pool returned to no missing dev status. Also re-establishes
+            # pool info as source of truth re missing.
+            if not p.has_missing_dev:
+                for disk in p.disk_set.filter(name__startswith="detached-"):
+                    logger.info("Removing detached disk from Pool {}: no missing "
+                                "devices found.".format(p.name))
+                    disk.pool = None
+                    disk.save()
             try:
                 # Get and save what info we can prior to mount.
                 first_dev = p.disk_set.attached().first()
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -1751,7 +1751,30 @@ def start_resize_pool(cmd):
     :param cmd: btrfs dev add/delete command in run_command() format (ie list).
     """
     logger.debug("Resize pool command ({}).".format(cmd))
-    run_command(cmd)
+    # N.B. in some instances, such as live disk removal, an attempt to remove missing:
+    # btrfs device delete missing /mnt2/pool-label
+    # results in rc=1, out=[""], and err=
+    # ["ERROR: error removing device 'missing': no missing devices found to remove", '']
+    # return advice, in this case, to improve usability. Customised exception message is
+    # presented in Pool details page - Balances tab, Errors or Notes column.
+    try:
+        run_command(cmd, log=True)
+    except CommandException as e:
+        emsg = (
+            "ERROR: error removing device 'missing': no missing devices found to remove"
+        )
+        if e.err[0] == emsg:
+            msg = (
+                "Missing Device removal failed, ensure degraded,rw mount options are used "
+                "(a reboot may be required)."
+            )
+            logger.error(msg)
+            raise Exception(
+                "{} Command was ({}). Command Exception was ({}).".format(
+                    msg, " ".join(cmd), emsg
+                )
+            )
+        raise e
 
 
 @task()
--- a/src/rockstor/storageadmin/views/pool.py
+++ b/src/rockstor/storageadmin/views/pool.py
@@ -460,15 +460,31 @@ class PoolDetailView(PoolMixin, rfc.GenericView):
                 # change on unmounted Volumes (pools).
                 return self._quotas(request, pool)
 
-            if not pool.is_mounted:
-                e_msg = ('Pool member / raid edits require an active mount. '
-                         'Please see the "Maintenance required" section.')
-                handle_exception(Exception(e_msg), request)
-
+            # Establish missing and detached disk removal request flag defaults:
+            remove_missing_disk_request = False
+            all_members_detached = False
             if command == 'remove' and \
                     request.data.get('disks', []) == ['missing']:
+                remove_missing_disk_request = True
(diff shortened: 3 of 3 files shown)


src/rockstor/storageadmin/views/command.py
```diff
@@ -90,6 +90,16 @@ def _refresh_pool_state():
                     "are no attached devices. Moving on.".format(p.name)
                 )
                 continue
+            # If pool has no missing remove all detached disk pool associations.
+            # Accounts for 'end of run' clean-up in removing a detached disk and for cli
+            # maintenance re pool returned to no missing dev status. Also re-establishes
+            # pool info as source of truth re missing.
+            if not p.has_missing_dev:
+                for disk in p.disk_set.filter(name__startswith="detached-"):
+                    logger.info("Removing detached disk from Pool {}: no missing "
```
FroggyFlox, line 99: Quick thought: Could we further detail the disk name (or any meaningful info) on the disk in question here? While testing, two detached disks needed to be removed and that results in the following:
```
[19/Jun/2020 07:59:36] INFO [storageadmin.views.command:100] Removing detached disk from Pool test_pool_03: no missing devices found.
[19/Jun/2020 07:59:36] INFO [storageadmin.views.command:100] Removing detached disk from Pool test_pool_03: no missing devices found.
```

It thus results in duplicates (or tri-, etc...) when that happens. Tiny detail, but if there's some some of identification readily available for the detached disk in question being removed, it could be helpful.

phillxnet: Good point except that the name of a detached disk is essentially random. The bit after detached is just a uuid (long historical reason to that) so the name wouldn't help. That random element is also changed on every disk refresh!! And other than that I guess the best thing would be the serial. Needn't hold up the pr merge though as almost all removals will be singular devices anyway given only meaningful removals (of viable pools) in raid 6 will be more than one given it's the only raid level that can handle more than one disk failure currently anyway (bar upstream btrfs raid1c3 and raid1c4).

I say we earmark this as a future logging improvement as we will be back here before long anyway. Nice idea though.

phillxnet:
@FroggyFlox Thanks for the review, double check. I think we are fairly safe with this one as it stands.

Much appreciated. I'll let is sit for a bit and see if anything more dawns on us and if not then I'll pop it in ready for the next testing release rpm.

PR #1992, opened 2018-11-21
add non legacy distro aware repo configuration. Fixes #1991

phillxnet:
Adds distro awareness to the existing repository configuration to encompass our next generation openSUSE Leap and Tumbleweed distribution bases. The repository structure assumed is as proposed/detailed in the associated issue text.

Summary:

- Add disto aware repo configuration for openSUSE Leap and Tumbleweed (next gen Rockstor).
- Fix Rockstor NG source to rpm install transition.
- Fix compliance with unconfigured repo. Allows for updates in only one channel i.e. preserving the user option to change the update subscription channel when defaulting to, or having selected, a deprecated or yet to be configured update channel.
- Fix unintended repository file permissions: auto resolves (post patch) by visiting “Software Update” page.
- Fix Rockstor NG version display for source installs.
- Fix missing yum-changelog package dependency: package name tested with yum on legacy and NG distros.
- For dev/source installs (no rockstor rpm installed) list entire changelog of available rpm. Prior use of ‘None’ for date parameter was silently interpreted as ‘all’ in older (legacy) yum-changelog but caused command exceptions in newer openSUSE versions of the same. Manual concurs with new use.
- Trivial code comment addition and typo.
- Changed rockstor repos from 1m to 1h for metadata_expire (to reduce server load).

Fixes #1991 

@schakrava Ready for review. See issue text for repo structure assumed within pr.
Git history caveat: I neglected to mention the metadata_expire change in the git commit message. Let me know if you would like for me to re-present this pr due to this omission.

Testing:

A source build, categorised as ‘unknown version’, offers an ‘update’ when any ‘rockstor’ rpm version is available: tested using existing CentOS based repos. Caveat for subsequent ‘upgrade’ is that the source build is assumed NOT to be in /opt/rockstor.

Legacy Rockstor (CentOS based) and Rockstor NG (openSUSE Leap & Tumbleweed based):
Successfully offered and configured both update channels from clean source builds as per repo dir structure proposed/detailed in issue text (unchanged for existing legacy repos).

PR #2189, opened 2020-06-24
add architecture aware repo configuration. Fixes #2188

### LINKED ISSUE #1991: add non legacy distro aware repo configuration
It is proposed that we add distro awareness to our yum repository configuration. This should assist with a smoother transition from existing CentOS installs to the non legacy openSUSE based installs. I.e. we can then initiate the alpha/beta rpm releases of openSUSE based installs while maintaining our existing CentOS based install updates; at least until such time as we have feature parity between the non-legacy and legacy releases. We can also prove the update system within the new openSUSE base, over a number of  Rockstor rpm updates, prior to releasing our first stable channel rpm update for an openSUSE based install (which could coincide with the first openSUSE based iso release; and in turn signalling our move 'proper' to an openSUSE base.)

It is further proposed that we use the currently dormant Testing Channel (in it's openSUSE variants) during this transitional phase. With the hope that we can rejuvenate this channel in our openSUSE based installs (assuming this distro move frees up some of our resources). Although I think the 'value add' of more regular updates for stable channel subscribers should be maintained, at least for the time being; but with an accompanying periodic testing channel update taken from latest stable. Which in turn would warrant a doc update on these channels. I.e testing for testing Rockstor's viability, stable for more rapid updates in exchange for assisting financially with it's development / sustainability.

Our existing repo structure is as follows:

| repo id | repo name | url
| --------- | -------------  | ---- |
| rockstor | Rockstor 3 - x86_64 | rockstor.com/rockrepo/ |
| Rockstor-Testing | Subscription channel for testing updates | updates.rockstor.com:8999/rockstor-testing |
| Rockstor-Stable | Subscription channel for stable updates | credentials@updates.rockstor.com:8999/rockstor-stable |

So the suggestion is that we enable a distro aware repo structure by maintaining the above for our existing CentOS based installs (distro.id() = 'rockstor') but augment this arrangement via additional sub-directories. This would ensure that existing subscribers, and associated back end support servers, should be unaffected (bar the re-install requirement / additional sub-directories, respectively).

Post 'fix' repository structure:

| distro base | iso (default) | Testing Channel | Stable Channel |
| ------------------- | -------------------------- | -------------- | ---------|
| CentOS installs | rockstor.com/rockrepo | updates.rockstor.com:8999/rockstor-testing | credentials@updates.rockstor.com:8999/rockstor-stable |
| openSUSE Leap 15 | rockstor.com/rockrepo/leap/15.0 | updates.rockstor.com:8999/rockstor-testing/leap/15.0 | credentials@updates.rockstor.com:8999/rockstor-stable/leap/15.0 |
| Tumbleweed | rockstor.com/rockrepo/tumbleweed | updates.rockstor.com:8999/rockstor-testing/tumbleweed | credentials@updates.rockstor.com:8999/rockstor-stable/tumbleweed |

I am currently looking into the rockstor-core code component of this proposal.

### LINKED ISSUE #2188: add architecture aware repo configuration
As we move toward providing official AArm64 rpms it has been found necessary to have the associated rpm repository discrete from our existing x86_64 ones. It is proposed that we do this by simply adding the machine arch, as returned by our exiting use of "platform.machine()" to our final repo url. But to preserve our existing installs validity we only do this if the arch is found to be NOT x86_64. This same scheme will then extend as-and-when/if we build for additional architectures, and will also help to keep each repo as small as possible as each will only contain rpms for it's specific/specified architecture, as apposed to all variants of Arm for example.

| Distro base | Arch | rpm repo url |
| ---| --- | --- |
| Leap 15.1 | x86_64 | http://updates.build.lan:8999/rockstor-testing/leap/15.1 |
| Leap 15.1 | aarch64 | http://updates.build.lan:8999/rockstor-testing/leap/15.1_aarch64 |
| Leap 15.2 | x86_64 | http://updates.build.lan:8999/rockstor-testing/leap/15.2 |
| Leap 15.2 | aarch64 | http://updates.build.lan:8999/rockstor-testing/leap/15.2_aarch64 |

An additional sub directory has not been proposed as we then keep all architectures at the same directory level which eases existing back-end rpm publishing mechanisms and avoiding 'embedded' repos that would then require createrepo "--excludes" for each and every architecture we add on our existing x86_64 repo. While at the same time keeping the required changes in rocsktor-core simple.

This proposal builds on the distro specific extensions added in:

"add non legacy distro aware repo configuration. Fixes #1991" #1992

phillxnet:
Provide additional independent repositories configuration for non x86_64 architectures by appending the output of platform.machine() to the existing x86_64 repo url. An underscore is used in this concatenation to aid in url readability.

Fixes #2188

See issue text for context.

- Includes unrelated removal of redundant call to platform.machine().

Ready for review.

## Testing:

On a 15.2 aarch64 install the resulting yum and zypper repo urls were as follows:
```
cat /etc/yum.repos.d/Rockstor-Testing.repo | grep baseurl
baseurl=http://updates.rockstor.com:8999/rockstor-testing/leap/15.2_aarch64

zypper repos -u | grep Rockstor
2 | Rockstor-Testing     | Rockstor-Testing          | Yes     | ( p) Yes  | Yes     | http://updates.rockstor.com:8999/rockstor-testing/leap/15.2_aarch64
```
While on an existing x86_64 install the urls are unaffected, as intended:
```
cat /etc/yum.repos.d/Rockstor-Testing.repo | grep baseurl
baseurl=http://updates.rockstor.com:8999/rockstor-testing/leap/15.2

zypper repos -u | grep Rockstor
 2 | Rockstor-Testing          | Rockstor-Testing                   | Yes     | (r ) Yes  | Yes     | http://updates.rockstor.com:8999/rockstor-testing/leap/15.2
```
And to account for any potentially stray/non-visible characters or the like the resulting repos were tested for activity, respectively, via:
```
yum --showduplicates list rockstor

zypper search -s rockstor
```
Both of which functioned as expected in that existing populated repositories (x86_64 in this case) were showing as functioning as their hosted rpms were indicated.

### Caveat
Ancillary note on testing via Web-UI: when subscribed to the testing channel and the stable channel is selected and activated with an empty activation code (produces an explanatory error), we are removing the existing testing channel config for both zypper and yum. Visiting the dashboard and returning to the Software Updates page re-asserts both the yum and zypper repos (to testing).

### This pull request represents an extension to:
"add non legacy distro aware repo configuration. Fixes #1991" #1992

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -167,6 +167,7 @@ def switch_repo(subscription, on=True):
     # Historically our base subscription url denotes our CentOS rpm repo.
     subscription_distro_url = subscription.url
     distro_id = distro.id()
+    machine_arch = platform.machine()
     use_zypper = True
     repo_alias = "Rockstor-{}".format(subscription.name)
     logger.debug("########### SWITCH REPO repo-alias = {}".format(repo_alias))
@@ -176,6 +177,8 @@ def switch_repo(subscription, on=True):
         subscription_distro_url += "/tumbleweed"
     else:
         use_zypper = False
+    if machine_arch != "x86_64":
+        subscription_distro_url += "_{}".format(machine_arch)
     # Check if dir /etc/yum.repos.d exists and if not create.
     if not os.path.isdir(repos_dir):
         # Can use os.makedirs(path) if intermediate levels also don't exist.
@@ -301,7 +304,6 @@ def rockstor_pkg_update_check(subscription=None):
         if not available:
             continue
         if new_version is None:
-            machine_arch = platform.machine()
             if re.match("rockstor-", l) is not None:  # legacy yum
                 # eg: "rockstor-3.9.2-51.2089.x86_64"
                 new_version = (


phillxnet:
@FroggyFlox This is another trivial architecture awareness patch earmarked for our next release as I need it in place prior to populating the pending AArch64 repos with rpms containing this code. But it does add a tiny additional mechanism. Looks to be fairly safe though.

Incidentally we use the yum (dnf-yum) configuration just for Rockstor available update change-log functionality.

PR #2193, opened 2020-07-14
add recognition of system disk & pool when on SD or MicroSD card. Fixes #2192

### LINKED ISSUE #2192: System disk not recognised when it is an SD or MicroSD card
When the system disk/device is a MicroSD card this is not recognised as the "Rockstor System drive". As a result the user is presented with various inappropriate options, such as import wipe etc via the cog icon, rather than the "r" icon used to denote, and protect, the system drive from regular data drive options. Further to this the system pool is also not recognised.

The cause for this issue has been identified as an omission in the root_disk() function which identifies the base name for the device hosting "/". More specifically there is no entry associated with the common SD/microSD card reader device name of:

```
Leap15-2Pi4:~ # lsblk 
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
mmcblk1     179:0    0 119.1G  0 disk 
├─mmcblk1p1 179:1    0    64M  0 part /boot/efi
├─mmcblk1p2 179:2    0     2G  0 part [SWAP]
└─mmcblk1p3 179:3    0   117G  0 part /
```

phillxnet:
For installs using an SD or MicroSD card as the system disk there is a failure to recognise this device as hosting the system with a consequent failure to surface the system pool. Further to this the user is presented with inappropriate choices; as if this disk was a regular data device.

Fix by adding the necessary device name match, and appropriate base name extraction filter, too root_disk(). In this case our existing mdraid device name treatment shares the same name structure assumptions with regard to base device name ("mmcblk1") in relation to partition name ("mmcblk1p3"). And so this patch proposes that an 'or' match is added to this existing section.

Fixes #2192
Ready for Review

## Testing
A unit test was developed for the existing and new functionality but as there is currently difficulty in mocking 'open' in Python 2 it has been included to avoid duplication of effort but remarked out as it was non functional. Functional tests were performed on a bare metal install using a MicroSD card as the system disk. Post proposed patch the system disk appeared, along with it's pool, as expected. Images of before and after will be attached in comments.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -1031,7 +1031,7 @@ def root_disk():
     Returns the base drive device name where / mount point is found.
     Works by parsing /proc/mounts. Eg if the root entry was as follows:
     /dev/sdc3 / btrfs rw,noatime,ssd,space_cache,subvolid=258,subvol=/root 0 0
-    the returned value is sdc
+    the returned value is /dev/sdc
     The assumption with non md devices is that the partition number will be a
     single character.
     :return: /dev/sdX type device name (with path) where root is mounted.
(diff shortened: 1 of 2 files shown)


phillxnet:
Prior to this patch the indicated issue was expressed thus:

sdcard-system-drive-not-recognized

Post patch we have our device correctly identified as the system device and it's pool (labled 'ROOT') is also surfaced.

sdcard-system-drive-recognized

Issue #2139, opened 2020-02-29
[OpenSUSE] IPV6 isn’t disabled by Yast and/or sysctl config


tyukh:
**Problem:** 
```
rockstor:~ # ip a 
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00 
    inet 127.0.0.1/8 scope host lo 
       valid_lft forever preferred_lft forever 
2: eno1: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond0 state UP group default qlen 1000 
    link/ether b0:5a:da:87:90:c3 brd ff:ff:ff:ff:ff:ff 
3: eno2: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond0 state UP group default qlen 1000 
    link/ether b0:5a:da:87:90:c3 brd ff:ff:ff:ff:ff:ff 
4: bond0: <BROADCAST,MULTICAST,MASTER,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000 
    link/ether b0:5a:da:87:90:c3 brd ff:ff:ff:ff:ff:ff 
    inet 192.168.1.10/24 brd 192.168.1.255 scope global dynamic noprefixroute bond0 
       valid_lft 86229sec preferred_lft 86229sec 
rockstor:~ # lsof -i 
COMMAND    PID   USER   FD   TYPE DEVICE SIZE/OFF NODE NAME 
NetworkMa 1180   root   22u  IPv4  34099      0t0  UDP rockstor.local:bootpc->192.168.1.1:bootps 
chronyd   1214 chrony    5u  IPv4  30050      0t0  UDP localhost:323 
chronyd   1214 chrony    6u  IPv6  30052      0t0  UDP localhost:323 
sshd      1217   root    3u  IPv4  30163      0t0  TCP *:ssh (LISTEN) 
sshd      1217   root    4u  IPv6  30174      0t0  TCP *:ssh (LISTEN) 
sshd      1611   root    4u  IPv4  33710      0t0  TCP rockstor.local:ssh->DESKTOP-IPI71Q1.local:53257 (ESTABLISHED) 
rockstor:~ # ss -tulwn 
[...]

PR #2205, opened 2020-08-20
Vagrant/Ansible to auto-build a test server and/or provision a real host.

mikemonkers:
Ansible to deploy/provision Rockstor to a host (refactored from Stephen Brown II work reported here: https://forum.rockstor.com/t/opensuse-ansible-playbook-to-get-rockstor-installed-on-fresh-system/7185 - accredited in the role README.md)

Vagrant to create a test host and use the ansible provisioner part of vagrant to join them together.

One command to build a VM with Rockstor installed ready to go.
```
Vagrant up
```
Or you can use the ansible directly on you own VM/physical host.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ node_modules/
 
 # VS Code
 .vscode
+
+# Vagrant working files.
+.vagrant/
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,0 +1,89 @@
+Introduction
+------------
+You can use this ansible environment to provision a VM or physical hose with a confirmed working 
+specification with rockstor.
+
+Pre-requesits
+-------------
+It is assumped that you have the follow software packages install on you host machine:
+
+- ansible
+
+There are a number of ways to install this depending on your Host OS.
+ 
+####Mac OSX
+
+1. Install [brew](https://brew.sh)
+    ```
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+    ```
+2. [option 1] Install ansible from brew:
+    ```
+    brew install ansible
+    ```
+    or
+   
+   [option 2] If you feel like running ansible from 'pip' under python 3:
+    ```
+    brew install pyenv
+    pyenv install 3.8.5
+    pyenv local 3.8.5
+    pip install ansible
+    ```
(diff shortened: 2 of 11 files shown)


mikemonkers:
I suspect you're comments to Stephen regarding to the IPv6 disablement will apply as I haven't attended to any of the comments in the thread. I am however, happy to do so.

phillxnet:
@mikemonkers Hello again. Re:

> I suspect you're comments to Stephen regarding to the IPv6 disablement will apply as I haven't attended to any of the comments in the thread. I am however, happy to do so.

Yes I'm afraid so. Could just be an additional sed and consequent "grub2-mkconfig -o /boot/grub2/grub.cfg". With ipv6 enabled some of our stuff breaks. See:

issue: [OpenSUSE] IPV6 isn’t disabled by Yast and/or sysctl config #2139
which in turn lead to additional instruction and pre-public changes to our rockstor-installer repo.

So if that and all the other requirements specified in:
up to where it goes to the source build, as per your docs (nice) arn't done they we don't have a compatible system. Also be careful with the use of "openSUSE" as it is a trademark, although we can use for example "Built on openSUSE" or "Uses openSUSE", there are a few other options specified in:

https://en.opensuse.org/openSUSE:Trademark_guidelines

But those are the two we have chosen to use. And we can apparently, from that doc, use these in combination with our own trademark; hence the use you've already seen on the forum etc of "Rockstor built on openSUSE".

Haven't looked pull request really but just wanted to drop these notes. I'll re-read what I said in the forum and take a proper look at this pr soon hopefully.

Thanks for formalising this and well done on the attribution side re Stephen Brown II for the forum. Oh and for doing docs as well. Nice. I'm not so sure of the various additional software packages listed to be installed. These could be remarked out however so folks could add them if they fancied and it still shows them how such packages might be added in the first place.

All in I think it would be good to the like of this folded into the project, but I'm just no keen on it straying from, for example, our rockstor-installer config. Ideally we want to add nothing extra but what is required, almost, hence the rockstor-installer example where we are almost a JeOS config before we add our hard dependencies. Anyway I've not taken a proper look just yet but those are my initial impressions.

Thanks again for preparing this pull request. We can now at least work through what we have here and work out between us what might work best. I'd also like for @FroggyFlox to take a look at this at some point.

PR #2208, opened 2020-08-24
Issue#2206 vagrant env

mikemonkers:
A Vagrant/ansible environment to build Rockstor from source, or install from RPMs on either Leap and Tumbleweed.

Deploying from RPMs as an alternative helps speed up our ability to construct a test VM based on RPMs for comparison against code changes and or reproducing reported issues.

Installing from ISO which is:

- quite mandraulic and therefore slow
- has potential for human error

For these reasons would result in a tendency to keep 'test' VMs around and live. Where as with Vagrant they can be more ephemeral, but more consistent. The vagrant and ansible also forms the 'recipe' documentation to a degree.

More, importantly I'd already got that working before raising the ticket so it was no trouble! ;-)

I'm about to submit the PR for review. A couple of things to point out:

1. The user vagrant need to maintain ssh access. You'll see this added in the core code - open for discussion?
2. In the ansible I've install the python2 packages from Python PIP instead or from the OSS repo RPMs. This is allows you to have both pip2 and pip3 in parallel and therefore python2 and python3 variants of the packages. It also gives you more up to date versions of these packages. It is possible to pin package versions with PIP if you prefer - which I imagine you might in a release.
3. Number 2 also fixes the Tumbleweed build dependencies - I think? Well it seem to compile but fails in Rockstor-pre due to the DM migration (shock news!)

Otherwise it defaults to building from source since that seems to be your preference and the most useful.

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,0 +1,89 @@
+Introduction
+------------
+You can use this ansible environment to provision a VM or physical hose with a confirmed working 
+specification with rockstor.
+
+Pre-requesits
+-------------
+It is assumped that you have the follow software packages install on you host machine:
+
+- ansible
+
+There are a number of ways to install this depending on your Host OS.
+ 
+####Mac OSX
+
+1. Install [brew](https://brew.sh)
+    ```
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+    ```
+2. [option 1] Install ansible from brew:
+    ```
+    brew install ansible
+    ```
+    or
+   
+   [option 2] If you feel like running ansible from 'pip' under python 3:
+    ```
+    brew install pyenv
+    pyenv install 3.8.5
+    pyenv local 3.8.5
+    pip install ansible
+    ```
+    Note: you may find you need to install the OSX Xcode commandline build tools for pyenv to build your python version.
+    
+####Linux
+
+Working under linux will be slightly distribution dependent but similar to Mac OSX options.
+
+OpenSUSE:
+1. Install from your package manager:
+    ```
+    zypper install ansible
+    ```
+2. Install through 'pip':
+    ```
+    pip install ansible
+    ```
+
+####Windows
+
+Follow this [guide](https://www.jeffgeerling.com/blog/2017/using-ansible-through-windows-10s-subsystem-linux) by Jeff Geerling.
+
+
+It is also assumed that you have a local copy of this repository [rockstor-core](https://github.com/rockstor/rockstor-core) 
+on your machine, see the top level README.md, and are currently within the 'vagrant_env' directory: where the 
+Vagrantfile is located.
+
+The Inventory
+-------------
+The file inventory/hosts.yml contains details of the host you are provisioning with ansible. You will need to add your
+hosts IP to this line:
+```
+rockstor ansible_ssh_host=<your_host_ip>
+```
+If your host has a different user from 'root' for provisioning you can also set that in the hosts.yml on this line:
+```
+[all:vars]
+ansible_ssh_user=root
+```
+Running the Ansible
+-------------------
+To run the provisioner:
+```
+ansible-playbook Rockstor.yml --ask-pass
+```
+This will prompt you for the password of the user defined in the 'ansible_ssh_user' in the inventory/hosts.yml.
+
+If you would like a little debug for the ansible execution (and this is my normal level of debug) try:
+```
+ansible-playbook Rockstor.yml --ask-pass -vv
+```
+
+If you happen to already have an authorised ssh key added to the 'ansible_ssh_user' on the rockstor host
+then you can just type the following:
+```
+ansible-playbook Rockstor.yml
+```
+
+
--- a/ansible/Rockstor.yml
+++ b/ansible/Rockstor.yml
@@ -1,0 +1,17 @@
+---
+- hosts: all
+  become: yes
+  become_user: root
+
+  roles:
+    - role: rockstor_installer
+      vars:
+        rs_inst_install_from_repo: yes
+#        rs_inst_set_root_pass: yes
+#        rs_inst_experimental: yes
+#        rs_inst_useful_server_packages:
+#            - ffmpeg
+#            - mediainfo
+#            - mosh
+#            - tmux
+#            - tree
--- a/ansible/Rockstor.yml
+++ b/ansible/Rockstor.yml
@@ -5,8 +5,8 @@
 
   roles:
     - role: rockstor_installer
-      vars:
-        rs_inst_install_from_repo: yes
+#      vars:
+#        rs_inst_install_from_repo: yes
 #        rs_inst_set_root_pass: yes
 #        rs_inst_experimental: yes
 #        rs_inst_useful_server_packages:
--- a/ansible/Rockstor.yml
+++ b/ansible/Rockstor.yml
@@ -5,7 +5,8 @@
 
   roles:
     - role: rockstor_installer
-#      vars:
+      vars:
+        rs_inst_update_os: yes
 #        rs_inst_install_from_repo: yes
 #        rs_inst_set_root_pass: yes
 #        rs_inst_experimental: yes
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -1,0 +1,10 @@
+[defaults]
+inventory = inventory
+roles_path=./installed_roles:./roles
+nocows = 1
+host_key_checking = False
+transport = paramiko
+#interpreter_python = python # Forces the ansible python interpret to follow pyenv
+
+[ssh_connection]
+ssh_args=-o ForwardAgent=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null
--- a/ansible/inventory/hosts
+++ b/ansible/inventory/hosts
@@ -1,0 +1,5 @@
+rockstor ansible_ssh_host=<your_host_ip>
+
+[all:vars]
+ansible_ssh_user=root
+
--- a/ansible/roles/rockstor_installer/README.md
+++ b/ansible/roles/rockstor_installer/README.md
@@ -1,0 +1,76 @@
+This role has is a refactoring of the good work by Stephen Brown in his repo 
+here: https://gitlab.com/StephenBrown2/rockstor-opensuse
+
+Kudos to him for this excellent work.
+
+## Main variables
+
+There are some variables in default/main.yml which can (or need to) be overridden:
+
+* `rs_inst_install_from_repo`: no[default] or yes. Install direct from testing repo.
+* `rs_inst_experimental`: no[default] or yes. Controls whether the latest BTRFS is included.
+* `rs_inst_set_root_pass`: no[default] or yes. Sometimes needed on first time installs of OpenSUSE.
+* `rs_inst_root_password`: The root password to be set. Must be encrypted eg. 
+    ```yaml
+    rs_inst_root_password: "{{ 'root' | password_hash('sha512', 'mysecretsalt') }}"
+    ```
+* `rs_inst_useful_server_packages`: A list of useful addition packages to install on the server (eg. to support a Rock-On).
+
+## Vars in role configuration
+Including an example of how to use your role (for instance, with variables passed in as parameters):
+```yaml
+---
+- hosts: all
+  become: yes
+  become_user: root
+
+  roles:
+    - role: rockstor_installer
+      vars:
+        rs_inst_install_from_repo: yes
+        rs_inst_set_root_pass: yes
+        rs_inst_useful_server_packages:
+            - ffmpeg
+            - mediainfo
+            - mosh
+            - tmux
+            - tree
+```           
+# rockstor-opensuse
+
+1. A typical rockstor 'Built on OpenSUSE' host should be configure following the instructions 
+here: https://forum.rockstor.com/t/built-on-opensuse-dev-notes-and-status/5662. 
+
+    *This ansible role is intended to be aligned with those install instructions.*
+
+2. After the install and initial reboot, create an SSH Key  and load it via `ssh-copy-id`. 
+Make sure you can log in using it (ie. without needing a password.)
+
+    *Note: if using vagrant this is typically already available for the vagrant user.*
+
+3. Create an inventory file for ansible, something like the following in `hosts.yml`:
+    ```yaml
+    ---
+    all:
+        hosts:
+            rockstor:
+                # Use your server's IP
+                ansible_ssh_host: 192.168.88.100
+                # Use the path to your local SSH Private Key
+                ansible_ssh_private_key_file: /home/youruser/.ssh/id_rsa_ansible
+                ansible_user: root
+    ```
+4. Create a playbook called 'main.yml'
+    ```yaml
+    ---
+    - hosts: rockstor
+      become: yes
+      become_user: root
+    
+      roles:
+        - role: rockstor_installer
+    ```
+5. Run the ansible playbook:
+    ```sh
+    ansible-playbook -i hosts.yml main.yml
+    ```
--- a/ansible/roles/rockstor_installer/README.md
+++ b/ansible/roles/rockstor_installer/README.md
@@ -14,6 +14,7 @@ There are some variables in default/main.yml which can (or need to) be overridde
     ```yaml
     rs_inst_root_password: "{{ 'root' | password_hash('sha512', 'mysecretsalt') }}"
     ```
+* `rs_inst_update_os`: no[default] or yes. Choose to update the OS packages or not.
 * `rs_inst_useful_server_packages`: A list of useful addition packages to install on the server (eg. to support a Rock-On).
 
 ## Vars in role configuration
--- a/ansible/roles/rockstor_installer/defaults/main.yml
+++ b/ansible/roles/rockstor_installer/defaults/main.yml
@@ -12,6 +12,9 @@ rs_inst_root_password: "{{ 'password' | password_hash('sha512', 'mysecretsalt') 
 # Rockstor currently requires IPv6 to be disabled.
 rs_inst_ipv6_disable: yes
 
+# Choose to update the OS or not.
+rs_inst_update_os: no
+
 # The user can use this to install any additional packages they may require.
 #rs_inst_useful_server_packages:
 #    - ffmpeg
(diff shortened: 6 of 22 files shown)


mikemonkers:
My test environments were:

Mac OSX High Sierra:
- Vagrant 2.2.9
- ansible 2.9.11
- VirtualBox 6.1.12

Mac OSX Catalina
- Vagrant 2.2.9
- ansible 2.9.11
- VirtualBox 6.1.10

phillxnet:
@mikemonkers  This is nice. Well done. Again I can't have a proper look for a while but I did spot:

> ... Aarch64 TW Rockstor package repo yet?

No, we only added AArch64 rpms recently, post Tumbleweed dropping the rpms, but we do, for the time being, have x86_64 Tumbleweed rpms built using a Tumbleweed instance that had the required python 2 libraries grandfathered in. Repo details in the following forum thread:
https://forum.rockstor.com/t/built-on-opensuse-testing-channel-live-early-adopters-developers-only/6592
Not sure how long I'll be able to maintain that Tumbleweed instance though. A plus for not using a fresh VM instance every time :) . We can't really spend any time on those though until we get our Python 3 move underway. It will then serve as a good test case.

And as for he 'controversial' adding of the vagrant user to ssh. Agreed this is suspect. But given, if I understand correctly, this is just to allow for the continued access of vagrant post the build (where the initrock etc change ssh stuff) why not just sed edit the code in the required places to insert the additional lines, as I really don't think we can add those lines for general distribution and building. This way the dev environment that this system is presumably setup to instantiate works, and we don't contaminate the source with this otherwise suspicious addition. Folks will often expect root to have this access but not so some random user. @FroggyFlox input on this would be good here as he is more recently up on ssh config within Rockstor.

> * quite mandraulic and therefore slow

Mandraulic! That a new one :). Although I expect openQA (one of our mid term test goals) could do this :) . I believe they already test default openSUSE installers this way. It's only a few button presses and easy targets for the included openCV. Plus we get a video of proceedings. All in good time I suspect.

All in I think this is a good idea for folks wanting to simply stand up a development instance. We just need to review carefully and be careful not to bend the rockstor-core code to this method. Hence the sed suggestion.

mikemonkers:
I'll look into amended the ssh controversy!

Not looked at OpenQA or OpenCV ... some late. Ugh the reading perhaps!

mikemonkers:
Ok... finally got back to this and fixed the controversial ssh tweaking for the vagrant user. I suspect we may stumble upon a few 'gaps' when reinstalling Rockstor Nukes the vagrant user access. If so we may be able to look at switching the vagrant user over to be the root user - although not entirely sure how at the moment.

phillxnet:
@mikemonkers Also of note is I've recently renegade on my preference for GRUB_CMDLINE_LINUX_DEFAULT
and have now in:
changed to suggesting the easier to match GRUB_CMDLINE_LINUX

Not that important but it's way easier to match via sed and keep things simpler. Our installer still does it 'properly' but that's an installer so should put it's 'stuff' in GRUB_CMDLINE_LINUX_DEFAULT where as here we are just modifying whatever we find on an existing install. May help with having a simpler more robust sed in your method here. That forum post has the sed I tried and found to work just fine.

I've been a bit short on time for reviewing I'm afraid but things should improve time wise soon hopefully.

Thanks for tending to this ssh issue, far better that we mess about with dev code, even via a sed, than upstream to core stuff we just don't need there. Agreed it's likely to catch us out later but this is development stuff anyway not for standing up installs which is what the installer is for here:

https://github.com/rockstor/rockstor-installer

And we have to keep that as the canonical reference really given development will always entail far more package surface area than is ever needed on a production/product install.

I'll have to up my ansible game to review this also I suspect. Given this and your prior commit is essentially my introduction to it beyond a concept. Still not enamoured with not basing off of JeOS but hopefully, such images will be available shortly and we can pop them in. Although I may have the nature of the image used here wrong of course.

Thanks for your continued attention on this pull request by the way.

If anyone fancies reviewing @mikemonkers contribution here then please feel free, especially if you familiar with vagrant/ansible.

Cheers.

mikemonkers:
> @mikemonkers Also of note is I've recently renegade on my preference for GRUB_CMDLINE_LINUX_DEFAULT
> and have now in:
> https://forum.rockstor.com/t/built-on-opensuse-dev-notes-and-status/5662
> changed to suggesting the easier to match GRUB_CMDLINE_LINUX
> 
> Not that important but it's way easier to match via sed and keep things simpler. Our installer still does it 'properly' but that's an installer so should put it's 'stuff' in GRUB_CMDLINE_LINUX_DEFAULT where as here we are just modifying whatever we find on an existing install. May help with having a simpler more robust sed in your method here. That forum post has the sed I tried and found to work just fine.
> 
> I've been a bit short on time for reviewing I'm afraid but things should improve time wise soon hopefully.
> 
> Thanks for tending to this ssh issue, far better that we mess about with dev code, even via a sed, than upstream to core stuff we just don't need there. Agreed it's likely to catch us out later but this is development stuff anyway not for standing up installs which is what the installer is for here:
> 
> https://github.com/rockstor/rockstor-installer
> 
> And we have to keep that as the canonical reference really given development will always entail far more package surface area than is ever needed on a production/product install.
> 
> I'll have to up my ansible game to review this also I suspect. Given this and your prior commit is essentially my introduction to it beyond a concept. Still not enamoured with not basing off of JeOS but hopefully, such images will be available shortly and we can pop them in. Although I may have the nature of the image used here wrong of course.
> 
> Thanks for your continued attention on this pull request by the way.
> 
> If anyone fancies reviewing @mikemonkers contribution here then please feel free, especially if you familiar with vagrant/ansible.
> 
> Cheers.

@phillxnet ansible has it's own 'replace' function that uses a regex:

```
- name: Set GRUB_CMDLINE_LINUX_DEFAULT
  replace:
    dest: /etc/default/grub
    regexp: ^GRUB_CMDLINE_LINUX=["'](.*)["']$
    replace: 'GRUB_CMDLINE_LINUX="ipv6.disable=1 \1"'
    backup: yes
```

Although I suspect it uses sed under the hood. 

I can switch the ENV VAR used to be 'GRUB_CMDLINE_LINUX'  to correlate with the instructions.

mikemonkers:
@phillxnet Regarding Vagrant Boxes... I just checked here: https://app.vagrantup.com/opensuse

It appears there are now official boxes for leap 15.2 x86_64 and aarch64 so I'm just testing and looking to switch to those official images in place of the bento box.

I don't see any of the JeOS flavour though.

mikemonkers:
@phillxnet I was actually wondering where this Dev environment would be better placed in it's own repository? Might help highlight the boundary and prevent crazy ideas like adding spurious users to the sshd_config!

mikemonkers:
Yes, the official OpenSUSE Leap 15.2 vagrant box works as you would hope. So lets use that.

phillxnet:
> It appears there are now official boxes for leap 15.2 x86_64 and aarch64

Nice. And yes, agreed on:

> So lets use that.

None of the others are anywhere close really.

> @phillxnet I was actually wondering whether this dev environment would be better placed in it's own repository?

Funny you should say that, it had 'half' occurred to me actually. I.e partial overload on rockstor-core. And it is a little odd, as you say, having this embedded in what is our 'core' app code. So then the question is what do we call the new repo?

I'm happy to make one, we just need to agree on a name, i.e. rockstor-dev-env, long but in keeping with our other redundantly named "rockstor-" but this can help with identifying when it's an isolated git pull; plus it's mostly what we have now. And 'dev-env' seems to cover it. I.e. this is a way to instantiate a development environment. @FroggyFlox, we will need your input on such a decision really. Also I'd like GPLv3 myself, as it's the same as we have in this repo?

A separate repo may also make for easier/quicker iteration as the hope is we get additional input on this method of standing up a development environment. It also means we can have this method pull rockstor-core at a tag version for instance; while still allowing progress on the config/methods introduced here.

Thanks for putting so much thought into all this. Lets see what the consensus is re separate repo and what it's name might be. And once that's sorted I can create as an empty repo with a GPLv3 licence, if that's the way to go.

phillxnet:
@mikemonkers Also meant to say: in your last pull request in the installer repo you mentioned that we can have both VirtualBox and KVM providers specified and vagrant will 'adopt' the first it arrives at, or finds installed presumably. I'm personally not a fan of the kernel module requirement of VirtualBox so can't see me using this much. But if it's 'free' to simply add KVM capability then it's looking a lot more attractive to the linux users who simply don't need Virtual Box installed except for instances such as this, given we generally have KVM build into the distro.

So in short can we simply add the a KVM provider and have this more portable to those who will not be using, or would rather not have what is, in their land, the additional dependency of Virtual Box?

phillxnet:
@mikemonkers given your recent official leap vagrant images find we also have the following in your last pull request:

`NOTE: Currently, only the x86_64 profile is usable due to the availability of suitable vagrant boxes for AArch64 in VirtualBox.
`

Which means that if we switch to the official repo there we should gain the aarch64 installer build capability also. Is this correct. If so one of us could do a further pr on the rockstor-install repo to switch that up to the official and hopefully also ensure it can do an aarch64 build. If in fact that is what is opened up here. For my own preperations for aarch64 installer building I've used Pi4 actually but, with a great deal more patience, it may be that folks without access to aarch64 hardware for this purpose could then build their own aarch64 installers. I'm pretty keen on catering for Arm64 where ever we can really.

Anyway just thought I'd note this here as it was relevant to your recent find re the official image and aarch64 variant availability. We are getting more interest shown in the rockstor-installer repo and your recent additions have definitely helped on that front. Still trying to entice a Win10 tester for your method there though!

mikemonkers:
@phillxnet So I've added the extra providers for 'libvirt' which ultimately uses QEMU and KVM. I'm hoping the links provided in the README.md are enough to explain and/or that you already have these even though you refer to KVM above? This was my first foray into the world of libvirt/QEMU/KVM to may have taken a detour to two!

I had quite a bit of trouble getting this to work in my Mac world (in fact I didn't succeed) and had to build an OpenSUSE Linux VM to Host libvirt/QEMU/KVM and test in there. However, the whole nested VMs malarky didn't give me much confidence that I had truly replicated your Linux world. None the less I was able to build a Leap VM in libvirt, build Rockstor, launch it and access it.

Access is now through port forwarding to allow you to run leap, Tumbleweed and Aarch64/Pi side by side.

Talking of Aarch64/Pi I couldn't get a VM to start here. I think it may be a conflict with the whole nested VMs business. I have put my best guess into the Pi VM Vagrant config (and left it commented out) and would be interested in someone else giving it a try. Please not that Virtual Box does NOT provide support for aarch64 and therefore the only available through libvirt.

Regarding moving to a new repo. I'm good with the name you suggest. Let me know when and where to move things to.

Finally, I can take a quick look at switching images in the rockstor-installer repo and raise a new PR there. I was intending to anyway once we'd pinned this down.

phillxnet:
@mikemonkers  This is looking very nice.

> @phillxnet So I've added the extra providers for 'libvirt' which ultimately uses QEMU and KVM. I'm hoping the links provided in the README.md are enough to explain and/or that you already have these even though you refer to KVM above?
Yes, I tend to use KVM Kernel Virtual Machine as shorthand for it's most common use in cohorts with qemu and libvirt.

> I had quite a bit of trouble getting this to work in my Mac world (in fact I didn't succeed)

Maybe this is just that the KVM bit is, I believe, linux kernel only. And I think this kernel base is almost ubiquitous with qemu as in other 'worlds' virtual box seems far more popular. It in turn has it's own kernel component which is basically a hassle for at least my linux use as it's 'non standard' and is an add on to the distro kernel. Where as KVM is and has been baked into all distro kernels for years now and so is a smooth user experience in that regard. But when combined with vagrant, which seems to favours the Virtual box world which in turn seems to favours the win/osx lands we hit the seemingly add-on nature of the vagrant-libvirt add on. As in it's not included by default. All swings and roundabouts I guess as all this stuff is relatively new. I had looked at vagrant earlier for my own internal build requirements and upon reaching the 'add-on' non default vagrant-libvirt plugin you references dropped it there as it was just a step too far. So well done on getting expanding our horizons a little with this more recent addition. I'd really like to see KVM, used by the likes of Digital Ocean and the like, get more central treatment in projects like vagrant. Oh well, maybe in time.



> However, the whole nested VMs malarky didn't give me much confidence that I had truly replicated your Linux world.

I know right. It does get a little wacky with VM's in VM's. But as long as the original (bare metal) host can support this then it should be just fine. In a recent 'event' in my own experiments in this area I was denied this nested capability on an install that closely mimicked a kin machine that could this. This anomaly lead me to the following:

https://docs.fedoraproject.org/en-US/quick-docs/using-nested-virtualization-in-kvm/

It seems that I had used a cutting edge KVM instance that had, by default, rewritten this capability config some how.

> None the less I was able to build a Leap VM in libvirt, build Rockstor, launch it and access it.

Nice. 

> Access is now through port forwarding to allow you to run leap, Tumbleweed and Aarch64/Pi side by side.

That a nice touch. In my own manually assembled test setup I pass each machine through to the local lan so they all appear as if they were individually plugged into the network each with their own IP. This is a more powerful config as then all services and port are available on all machines. But for basic building this port forward approach should work to enabling a build setup/basic test arrangement.

> Talking of Aarch64/Pi I couldn't get a VM to start here.

This is a tricky one yes. I too have failed on this front. I have to animalisticly write an image to a real sdcard and boot it!!!
I've read that a Pi4 profile is in the works for KVM libvirt so hopefully we can incorporate that when it arrives. Pi3 is a little different on the boot front as they have incorporated, in the Pi4, more boot code into the actual firmware. This in turn allows for USB boot which is nice. And our current image works this way also if the appropriate files are changed.

> Please not that Virtual Box does NOT provide support for aarch64 and therefore the only available through libvirt.

Good to know. I really think our developer audience is far more likely to be linux based, but I don't actually know. However as a point of interest when I ran a haswell KVM x86_64 hosting an emulated aarch64 it took 2h 50 minutes per build. When I moved all aarch64 builds to a Pi4 (running JeOS Leap 15.2 with a couple of KVM instances) the native cpu passthrough from VM to host enabled a build time of around 15 minutes or so. A little more reasonable I thought. So I now have a dedicated Pi4 to do all aarch64 testing. But as stated I have then to re-use this to then do real boots to test our installer which is a pain. So in short I think it's going to be unworkable to expect folks to emulate aarch64 in x86_64 as it's simply too slow. But given the growing abundance of Pi4 and it's massive speed-up from Pi3 A52 rather than A72 I think, we may just have to rely on our more major contributors having native Arm64 hardware. Incidentally the "ARM64EFI" profile contributed recently by @mcbridematt works a treat within libvirt. So we can far more easily test that profiles installer virtually as I currently do, given it's one of our supported profiles. So you could switch to building for that profile as it may well be far more straight forward than the 'special' Pi4 installer build.

> Regarding moving to a new repo. I'm good with the name you suggest.

@FroggyFlox Can we have your input on this one. @mikemonkers project here is showing great promise as a stand-up dev setup for those wanting to test their changes, and having it in it's own repo would make send I think, as discussed earlier in this pr. So I think we just need a name and then we can get this committed and ready for others to tweak. Current name suggestion is "rockstor-dev-env" but I'm up for other suggestion. And I'm currently viewing @mikemonkers excellent work here as a way to enable prospective code contributors to easily stand up virtual instances upon which to deploy and test their changes.

> Finally, I can take a quick look at switching images in the rockstor-installer repo and raise a new PR there.

That would be great.

Thanks again for all of this. I think, once we have the details pinned down, this could server as a nice standardised way for folks to pre-test their changes which can only help. And with this element of on-boarding simplified we are likely to more easily attract developers that would otherwise be put off with all my long hand explanations of such things.

I'll make the new repo once we have @FroggyFlox input on naming/positioning of the proposed repo. We do have an internal build/final-test/deployment system based on buildbot but it's main focus is currently on deployment. We will have to see how the sands of time play out on all of this. It would also be great to have a dedicated issue/pr listing for what you have build here.

FroggyFlox:
> > Regarding moving to a new repo. I'm good with the name you suggest.
> 
> @FroggyFlox Can we have your input on this one.

@phillxnet , @mikemonkers , sorry for not getting back to you earlier... I do like the idea of having it in a separate repo too, it will keep this one as focused as we can, and yet provide an easy way to build dev env for those who prefer it that way.
I'm also OK with the name: it is consistent, and clearly describes its intent as a development tool.

Sorry again for the delay in getting back to you... I still haven't found the time to even give it a try, unfortunately.

phillxnet:
@FroggyFlox Thanks. So we have a consensus.
@mikemonkers @FroggyFlox We now have:

https://github.com/rockstor/rockstor-dev-env

If you fancy creating the first pull request there as per the current state of this one but and link it to the chat we have had here for context then maybe we just pop that in and then start on the fixes as we go via smaller issues/pull requests and hopefully this will get a life of it's own. I like @FroggyFlox have yet to test it or even look over it properly unfortunately so it is likely that you may have to baby sit this new repo given you are currently the only one with working knowledge of it's state. So if you could keep an eye on it for us and try and assess the sanity of future pull requests on it that would be great, as currently I'd quite like to get it in as it is so it can get the hoped for life of it's own via clearly defined revisions.

Quite exciting project actually and I hope soon to be able to point folks at this as the preferred way to stand up what's initially needed to get Rockstor development setup up and running quickly. But I don't expect the Tumbleweed builds to be anywhere usefully for a bit, given the python 2 libraries issue so we should probably try and make that clear, maybe in a further pr, in the associated doc. That way we set expectations for those jumping in. I'm also no longer publishing Tumbleweed rpms until we have our Python 3 move well under way. And we may even have to revise our currently Python 2 only build system before then anyway. All a bit of a challenge but hay ho.

Thanks again for you continued efforts in raising our on-boarding game here. It can only help. And I'm looking forward to familiarising myself with this process. But until then, which may be a little while I'm afraid, it would be most appreciated if you could keep an eye on the new repo as the rest of the core developers get up-to-speed on this approach. It's probably about time we had this new repo and this your hopefully pending pr is an excellent start, at least that's my hope. And well done for doing the docs as you go along also. Super helpful/essential.

New Repo is, as per this one, GPLv3 for consistency and because that is our chosen default project wide where we can. Our Rock-ons repo is however AGPL v3.

I look forward to your fresh pull request in the new repo and hopefully we can get some takers on successive improvements.

mikemonkers:
New PR created here: https://github.com/rockstor/rockstor-dev-env/pull/1

mikemonkers:
@phillxnet regarding Tumbleweed, it does actually build in this environment due to my pulling the python dependencies from pip. This env does this for Leap too as it happens which maybe be a point of contention versus using the bistro python packages. Which causes this to differ from the kiwi mechanics in https://github.com/rockstor/rockstor-installer.

Just thought I'd point that out in case you'd not spotted that.

mikemonkers:
I shall close this PR given it is not covered int he new repo at rockstor/rockstor-dev-env#1

mikemonkers:
@phillxnet I'm happy for the branch to be deleted, if you are?

PR #2213, opened 2020-09-14
[NG] - Authorized_keys doesn't allow public key ssh login. Fixes #2212

### LINKED ISSUE #2212: [NG] - Authorized_keys doesn't allow public key ssh login
Hope this is the right place to report this (let me know if not :) ).

Fresh install of 4.0.1 on Leap 15.2 using the testing packages method.

**Problem:** 
Pasting an authorised key into a user profile saves the key to /home/$user/.ssh/authorized_keys, but I believe due to the file permissions of the .ssh directory and file created it doesn't actually allow the user to log in via SSH.

**Steps to reproduce:**
Create a user and paste a public key in the profile. Try to login via SSH with that key, and you're asked for a password.

**Suspected cause:**
The key/.ssh directory is created with the following permissions:
`nas:/home/rob # ll /home/rob/ | grep .ssh
d-w-rwxr-T 1 rob users   30 Sep 13 20:53 .ssh`

And the key:
`nas:/home/rob # ll /home/rob/.ssh/
total 4
---x-wx--T 1 rob users 394 Sep 13 20:53 authorized_keys`

However, if I remove the .ssh directory and run the usual ssh-copy-id command from a client, the permissions are as follows (and public key login works):
`nas:/home/rob # ll /home/rob/ | grep .ssh
drwx------ 1 rob users   30 Sep 13 20:55 .ssh`

And the key:
`nas:/home/rob # ll /home/rob/.ssh/
total 4
-rw------- 1 rob users 394 Sep 13 20:55 authorized_keys`

https://en.opensuse.org/SDB:OpenSSH_public_key_authentication#Security explains the permissions I believe in a bit more detail, but I believe it must be at most 700 and preferably 600.

phillxnet:
Fix incorrect use of os.chmod leading to dir/file permissions based failure of public key based ssh/sftp user access.

Current Rockstor defaults only allow the root user access via ssh, however sftp access can also use public ssh key based authentication. If a non root Rockstor user is also the owner of a Rockstor share, and that share is exported via SFTP, key based authentication can be enabled by supplying a client public key to the Rockstor users profile:

System - Users - Edit user (pen icon)

The resulting sftp access is within a chroot environment wherein the owned sftp exported shares are mapped.

But this public key based authentication method failed due to safeguards in the ssh subsystem when incorrect dir/file permissions were found.

Fixes #2212 

## Security note
Only public keys entered via the Rockstor Web-UI (System - Users edit) method would have been affected by the resulting incorrect file permissions. However if a private key was generated via the command line, and the same user also had a public key entered into their user profile via the Rockstor Web-UI thereafter then the private key would also be affected by this bug.

## Resolving an existing config (via Web-UI)
Removing an existing public key, submitting the change, and then re-adding the key resolves all known directory and file permissions resulting from this bug for the given user. And by virtue of then conforming to the strict dir/file permissions enforced by the ssh subsystem, ssh based key authentication functions as expected.

### DIFF
--- a/src/rockstor/system/users.py
+++ b/src/rockstor/system/users.py
@@ -31,6 +31,7 @@ import chardet
 import random
 import string
 import crypt
+import stat
 
 import logging
 
@@ -233,7 +234,8 @@ def add_ssh_key(username, key, old_key=None):
     if not os.path.isdir(SSH_DIR):
         os.mkdir(SSH_DIR)
     run_command([CHOWN, "-R", "%s:%s" % (username, groupname), SSH_DIR])
-    os.chmod(SSH_DIR, 700)
+    # Set directory to rwx --- --- (700) via stat constants.
+    os.chmod(SSH_DIR, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
     fo, npath = mkstemp()
     exists = False
     with open(AUTH_KEYS, openmode) as afo, open(npath, "w") as tfo:
@@ -248,5 +250,6 @@ def add_ssh_key(username, key, old_key=None):
     if exists:
         return os.remove(npath)
     move(npath, AUTH_KEYS)
-    os.chmod(AUTH_KEYS, 600)
+    # Set file to rw- --- --- (600) via stat constants.
+    os.chmod(AUTH_KEYS, stat.S_IRUSR | stat.S_IWUSR)
     run_command([CHOWN, "%s:%s" % (username, groupname), AUTH_KEYS])


phillxnet:
As per the excellent report/issue opened by @rssfed23, pre pull request dir/file permissions were confirmed post the addition of a public key to a user profile via the Rockstor Web-UI. Post pull request we have:
```
drwx------ 1 philip users  30 Sep 14 12:43 .ssh
-rw------- 1 philip users 397 Sep 14 12:43 authorized_keys
```
These dir/file permissions look to be conformant with current best practice, and the original intent of the erroneous use of os.chmod, and also result in a functional key based ssh/sftp authentication.

Reference, as per issue: https://en.opensuse.org/SDB:OpenSSH_public_key_authentication#Security

phillxnet:
@FroggyFlox I've not applied our usual code house keeping of moving to string.format in this case as I thought it important to keep the changes here as clear and simple as possible. We can always revisit this file in that regard when next we do a stint of that as part of our ongoing Python 3 move.

FroggyFlox:
Thanks a lot for such a quick fix and for the notice, @phillxnet !

I'm still swamped on my end so I probably won't have time to test your PR within the next few days, unfortunately. Changes seem to be very focused and limited however, so I'm confident it's all ok! Bottom line, don't wait for me to test if before merging that one if you want it in quickly.

Cheers,

PR #2214, opened 2020-09-14
9000 user limit within NIS breaks Web-UI for larger user sets. Fixes #2211

### LINKED ISSUE #2211: 9000 user limit within NIS breaks Web-UI for larger user sets
Thanks to forum member legion411 in the following forum thread for reporting this issue. It is strongly suspected that we have a repeat of our prior user limit expansion from 5000 to 9000 in issue:

"On GUI, only 5000 user show up" #1915
addressed in pull request:
"On GUI only 5000 user show up. Fixes #1915" #1916

Only this time the reported user count suspected of triggering the indicated failure is around 30,000.

In this new assumed related context, attempting to add a Samba share thus:

Click ‘Storage’–>‘File Sharing’–>Samba’–>‘Add Samba Export’

resulted in the following:
```
“Unknown internal error doing a GET to /api/users?page=1&format=json&page_size=9000&count=”
```
Observed in 3.9.2-57 CentOS base.

Please update the following forum thread with this issues progress:
https://forum.rockstor.com/t/error-while-attempting-to-create-smb-export/7313

phillxnet:
Increase maxPageSize from 9000 to 32000 predominantly to deal with larger installs, i.e. > 9000 users / groups. System wide change so should also accommodate all other datatable displays and background functions within the Web-UI.

The observed failure in the associated issue, with an NIS user count of greater than 30,000, was triggered via:

Click ‘Storage’–>‘File Sharing’–>Samba’–>‘Add Samba Export’

Resulting in:
```
“Unknown internal error doing a GET to /api/users?page=1&format=json&page_size=9000&count=”
```

This patch is an attempt to narrow down the possible causes of this error by at least raising our known limitation of 9000 items per call. Thus:

Fixes #2211

Hopefully. But it is not unlikely that we have another issue at play here, i.e. time-out related.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/rockstor.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/rockstor.js
@@ -426,7 +426,7 @@ RockStorGlobals = {
     navbarLoaded: false,
     applianceNameSet: false,
     currentAppliance: null,
-    maxPageSize: 9000,
+    maxPageSize: 32000,
     browserChecked: false,
     kernel: null
 };


phillxnet:
For context I am linking this pull request to an essentially identical limit increase from 5000 to 9000 performed in the following pull request:

"On GUI only 5000 user show up. Fixes #1915" #1916

phillxnet:
@FroggyFlox Merging this as 'mostly harmless'. We've done this exact same thing before, over 2.5 years ago, in the references pull request.

I do think we need to re-visit this mechanism as getting 32000 items in a go is likely not quite what we want but given it previously created a 'barrier' of sorts and that we have again hit this same barrier, presumably, I'm inclined to merge as I don't think it's going to affect anything that has fewer items than 9000 anyway, and we have had no ill reports with this limit, so a 3.5 x jump will hopefully only move this issue aside and either fix the reported issue or enable another to be uncovered.

Builds with this added behave < 1s differently in our 205 unit test run so I think we are safe popping this one in.

PR #2221, opened 2020-09-30
NIS regression re Rockstor 4 release candidates. Fixes #2216

### LINKED ISSUE #2216: NIS regression re Rockstor 4 release candidates
Thanks to forum member legion411 for highlighting this issue. In Rockstor 4 release candidates we have a regression/failure in NIS configuration via the Web-UI. Initial investigation by @FroggyFlox has identified the cause to be a variation in how NIS is configured on our new disto base of openSUSE Leap15.2 to that of our prior CenOS base. Config file locations and nature have changed and authconfig, a RedHat/CentOS specific program, does not exist.

The observed user visible error in the case of attempting to configure NIS in our new 'Built on openSUSE' variant is the following:

> _Houston, we've had a problem
> NIS could not be configured. Try again
> ...
> IOError: [Errno 21] Is a directory: '/etc/sysconfig/network'_

Forum thread post reference:
https://forum.rockstor.com/t/error-while-attempting-to-create-smb-export/7313/13

phillxnet:
Modify NIS config file paths and nature of edits appropriate for the current 'Build on openSUSE' endeavour and move the use of chkconfig to systemd which chkconfig was more recently an aliased for anyway. N.B. we no longer release CentOS variants so backward compatibility was not addressed.

Includes:
- Added remarks to explain config_nis().
- Remarking out CentOS'ism (authconfig).
- Addition of executing "netconfig update -f" to instantiate yp.conf edits.

Fixes #2216 

## Testing
An NIS server was setup with around 30,000 users as detailed in the referenced issue. Post pr patch an otherwise 4.0.2-0 Rockstor instance was able to successfully register with the NIS server via Web-UI config means only. Successful NIS client config was then established via:
```
pro8:~ # ypcat passwd | wc -l
30026
```
Demonstrating yellow pages's registration of the addition 30,000 users on top of the client machines default users.

Turning off the NIS service was also observed to in turn remove the additional users. Rockstor's user related Web-UI components were also observed to reflect these consequent users changes.

### DIFF
--- a/src/rockstor/smart_manager/views/nis_service.py
+++ b/src/rockstor/smart_manager/views/nis_service.py
@@ -18,7 +18,8 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 from rest_framework.response import Response
 from storageadmin.util import handle_exception
-from system.services import init_service_op, chkconfig
+from system.osi import run_command
+from system.services import systemctl
 from system.nis import configure_nis
 from django.db import transaction
 from base_service import BaseServiceDetailView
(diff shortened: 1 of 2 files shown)


phillxnet:
@FroggyFlox I've test built with this patch included and there doesn't seem to be an blockers so I'll get this merged ready for the next release.

PR #2222, opened 2020-09-30
remove redundant chkconfig use. Fixes #1986

### LINKED ISSUE #1986: remove redundant chkconfig use

### LINKED ISSUE #2216: NIS regression re Rockstor 4 release candidates

phillxnet:
Post removal of our last remaining use of chkconfig, see commit: 508e883

"NIS regression re Rockstor 4 release candidates. Fixes #2216"

this patch removes it's binary path definition within both production and test settings files and it's associated system/services.py wrapper.

Fixes #1986

## Caveats
This pr is proposed as a follow-up to the indicated, and pending:
"NIS regression re Rockstor 4 release candidates. Fixes #2216" #2221
As such it should be merged after #2221 as it removes elements that are still dependants pre #2221

### DIFF
--- a/conf/settings.conf.in
+++ b/conf/settings.conf.in
@@ -437,7 +437,6 @@ OAUTH2_PROVIDER_APPLICATION_MODEL = 'oauth2_provider.Application'
 # The following have been tested in CentOS, openSUSE Leap15, and Tumbleweed
 UDEVADM = subprocess.check_output(["which", "udevadm"]).rstrip()
 SHUTDOWN = subprocess.check_output(["which", "shutdown"]).rstrip()
-CHKCONFIG_BIN = subprocess.check_output(["which", "chkconfig"]).rstrip()
 
 # Establish our OS base id, name, and version:
 # Use id for code path decisions. Others are for Web-UI display purposes.
--- a/conf/test-settings.conf.in
+++ b/conf/test-settings.conf.in
@@ -395,7 +395,6 @@ UPDATE_CHANNELS = {
 # The following have been tested in CentOS, openSUSE Leap15, and Tumbleweed
 UDEVADM = subprocess.check_output(["which", "udevadm"]).rstrip()
 SHUTDOWN = subprocess.check_output(["which", "shutdown"]).rstrip()
-CHKCONFIG_BIN = subprocess.check_output(["which", "chkconfig"]).rstrip()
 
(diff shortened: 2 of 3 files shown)


phillxnet:
@FroggyFlox the pre-requisite of #2221 is now merged and I've done a final test build of this pr on resulting master and again all looks to be as expected, so merging also ready for next release. This is really just a house keeping tidy up and as per test doesn't appear to have upset anything as all elements removed are now no longer referenced anywhere anyway.

PR #2225, opened 2020-10-07
minor optimisation of smb share add - "Admin user" option. Fixes #2220

### LINKED ISSUE #2220: slow admin user filter on Samba export when using Chrome
While working on an NIS config regression issue where the test case is 30,000 users, it was found that in Firefox each character entered into the Add Samba Export "Admin User" form field took approximately 1 second to resolve (show matching users in drop down). Where as the same function within Chrome resulted in a 20 second per character delay, which given the length of this delay could easily lead to a time out on a typical user name entry give the delay is accumulative; and a decidedly unpleasant/miss-leading user experience re unresponsive UI.

These measurements were observed on a 2013 Haswell desktop class machine running openSUSE Leap 15.2 Linux with relatively recent versions of both browsers. At this point the majority of Web-UI code is running browser side so the Rockstor server speed is far less relevant.

It is suspected that a difference in JS optimisation is causing this issue and the suspected JS code segment is within the "display_adminUser_options" HandlebarHelper:

https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/storageadmin/static/storageadmin/js/views/add_samba_export.js#L245-L264

And initial suspicion is the lack of loop length caching within the inner for command used to match input to each in turn of all known users, resulting in repeated redundant calls to the non trivial (at scale) _this.smbShare.get('admin_users').length

```
                    for(i=0; i< _this.smbShare.get('admin_users').length; i++){
                        if(_this.smbShare.get('admin_users')[i].username == userName){
                            html += 'selected= "selected"';
                        }
```
In order to provide the html highlight of a matching user within the options drop down.

phillxnet:
Includes:

Select2 (likely main client cpu user):
- Add min input length to avoid extraneous 1 to 3 char inputs.
- Add custom simplified filter, i.e. starts with rather than sub string.

Options highlighter pre-processor:
- Comments to indicate the observed reproducer.
- Minor inner look optimisation via caching vars.

Fixes #2220 
See issue text for context.
Fixes in this case is not quite accurate but it helps with usability a little in Chrome. Note this issue has to date only been observed as a problem in Chrome. Firefox behaves just fine both pre and post patch.

Summary of browser behaviour:
Firefox: relatively unchanged performance as usable pre and post patch. No massive delays.
Chrome: slightly improved usability re faster initial responsiveness. But still poor performance per 3rd key press there after.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/add_samba_export.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/add_samba_export.js
@@ -139,7 +139,40 @@ AddSambaExportView = RockstorLayoutView.extend({
         if(this.sambaShareId == null) {
             this.$('#shares').select2();
         }
-        this.$('#admin_users').select2();
+
+        // https://select2.org/searching
+        // https://select2.org/searching#customizing-how-results-are-matched
+        function matchCustom(params, data) {
+            // If there are no search terms, return all of the data
+            if ($.trim(params.term) === '') {
+                return data;
+            }
+
+            // Do not display the item if there is no 'text' property
+            if (typeof data.text === 'undefined') {
+                return null;
+            }
+
+            // Search area
+            // `params.term` should be the term that is used for searching
+            // `data.text` is the text that is displayed for the data object
+            // substring didn't appear any faster than indexOf:
+            // if (data.text.indexOf(params.term) === 0) {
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox As per our quick side channel discussion I'm planning to pop this one in. Helps a bit and doesn't look to be risky during our RC stage testing channel phase.

PR #2245, opened 2021-01-06
Unit test maintenance re legacy disparity. Fixes #2243

### LINKED ISSUE #2243: Unit test maintenance re legacy disparity
Prior to embarking on our Rockstor 4 release candidates the existing unit tests on our prior CentOS base were modified to have zero failures/errors in the following pull request:
"remark out all currently failing unit tests. Fixes #2137" #2153
In the mean time we have added additional unit tests and still have some existing failures in prior tests on the non legacy 'Built on openSUSE' endeavour. This issue is to address this disparity by returning our unit test status to that of our prior OS base with the intention of establishing this state prior to our final Stable 4 release.

phillxnet:
Re-establish prior CentOS baseline of all existing tests passing.

Fixes #2243
Ready for review.

The aim here was to re-establish our pre 'Built on openSUSE' endeavour baseline established in our prior CentOS base. This looks to have been achieved and although there is still much work to be done in the tests, mainly failures to mock actual system state effectively, it does look like we have at least returned to our prior 'baseline' by explainable OS differences alone.

## Summary of changes:
- Delete of existing, non excluded user 'games' fails as non existent in openSUSE base. Replace with non excluded 'lp' user that does exist.
- Creation of new user with existing uid test failed as prior uid of 99 for nobody (CentOS) is now available. Change to 65534 which is openSUSE uid for user nobody.
- Auto_update (legacy yum-based) is now defunct: we are now zypper based. Disable/Enable tests failed with "No such file or directory: '/etc/yum/yum-cron.conf'" but are intended as API level tests. Mock low level auto_update() and add TODO to test_pkg_mgmt.py for the appropriate low level test once we migrate this functionality to a zypper equivalent. This way we preserve & test the presence of the existing API.
- Group create with pre-existing group id fails as group id used did not actually exist. Substitute with gid of users (100) which does exist in our new openSUSE base.
- Prior samba test created a temporary user that relied on a real system user gid existing. This same gid does not exist in the new openSUSE base, resulting in a an exception. Fix by mocking a fake group name return via getgrgid.
- Add missing docker_status() mocking that was failing a more recently added low level network test: test_get_con_config(), due to system docker state. Thanks to @FroggyFlox on GitHub for this fix.

### Testing results:
As for git/GitHub tag 4.0.5-0 which equates to Release Candidate 6

### DIFF
--- a/src/rockstor/storageadmin/tests/test_commands.py
+++ b/src/rockstor/storageadmin/tests/test_commands.py
@@ -56,6 +56,12 @@ class CommandTests(APITestMixin, APITestCase):
         cls.mock_update_check = cls.patch_update_check.start()
         cls.mock_update_check.return_value = 1, 1, 1
 
+        # Mock src/rockstor/system/pkg_mgmt.py auto_update() in
+        # src/rockstor/storageadmin/views/command.py
+        cls.patch_auto_update = patch('storageadmin.views.command.auto_update')
+        cls.mock_auto_update = cls.patch_auto_update.start()
+        cls.mock_auto_update.return_value = True
+
         cls.patch_system_shutdown = patch(
             'storageadmin.views.command.system_shutdown')
         cls.mock_system_shutdown = cls.patch_system_shutdown.start()
--- a/src/rockstor/storageadmin/tests/test_group.py
+++ b/src/rockstor/storageadmin/tests/test_group.py
@@ -92,12 +92,12 @@ class GroupTests(APITestMixin, APITestCase):
 
         # invalid gid
         data = {'groupname': 'ngroup2',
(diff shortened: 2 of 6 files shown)


FroggyFlox (approved):
@phillxnet, thanks a lot for taking care of this; it all works as expected on my end as well:
```
# ./bin/test -v 2
(...)
Ran 212 tests in 60.203s

OK
```

phillxnet:
@FroggyFlox Thanks for the double check. Much appreciated. Given this pr only makes changes to the testing code I'm going to merge so we can have this testing set as the new normal. I'll update the forum dev wiki accordingly post merge.

PR #2247, opened 2021-01-07
Correct typo in field label (#2244)

### LINKED ISSUE #2244: Inaccurate field label for Certificate path in LDAP service configuration form
Following our rework of LDAP implementation (#2235), we now require the user to provide a **path** to TLS certificate rather than a **URL** to the certificate. Unfortunately, the corresponding form label was not updated accordingly in #2235 (my fault). See example below:
image

This can be rather misleading to the user and should thus be changed to **Certificate path** rather than "Certificate URL".

FroggyFlox:
Fixes #2244.
@phillxnet, ready for review.

Very simple commit to fix a typo in field label for the "Certificate path" in the LDAP service configuration form.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/services/configure_ldap.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/services/configure_ldap.jst
@@ -39,7 +39,7 @@
                         </div>
                     </div>
                     <div class="form-group" id="cert-ph">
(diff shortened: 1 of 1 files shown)


phillxnet (approved):
@FroggyFlox Thanks for sorting this out. Adds clarity and nice to see an included label / input association fix along the way via the "for=" element fix.

ldap-path-label

Merging as works for me via above screen grab.

PR #2262, opened 2021-01-24
Remove donate menu item and update legacy paypal mechanism (#2260)

### LINKED ISSUE #2260: Remove donate menu item and update legacy paypal mechanism
Given "Update Readme.md and add paypal.me donate link (#2250)" #2257 has now been merged, it is proposed that we similarly adopt this non legacy donate mechanism within our Web-UI. Further to this, it is also proposed that we remove the top menu "Donate" entry given its single recorded use in multiple years of 'service'. This would also serve to remove an undesirable external link, the paypal button gif, from our Web-UI and likely the entire donate dialog mechanism. The intention is to keep the existing donate references within for example the updates page but transition these to the non legacy paypal.me link from the proposed for removal main menu item initiated donate dialog.

Linking to another prior resolved and related issue for context:
revise donate menu item within Web-UI. Fixes #2104 #2105

phillxnet:
See referenced issue for context. The paypal button we have used for our donation link is now regarded as legacy. This pr moves our donate Web-UI elements to the newer paypal.me mechanism and centres our donate related text on the update page while simultaneously removing the main Donate menu item and it's associated dedicated dialog.

Fixes #2260 
Ready for review.

Summary of changes:
- Remove Donate menu entry.
- Remove Donate dialog.
- Move prior Donate dialog main text re donations etc to update page.
- Add commercial support links to subscription dependant update page text.
- Move remaining donate links from legacy paypal button to paypal.me.
- Updated relevant copy write notices date element.

### Top menu (no Donate entry)
rhs-menu-post-pr
### No subscription activated
no-subscription-top-banner
and
commercial-support-options-in-table
### Stable updates enabled
stable-enabled-post-pr
### Testing updates enabled
testing-enabled-post-pr

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/router.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/router.js
@@ -3,7 +3,7 @@
  * @licstart  The following is the entire license notice for the
  * JavaScript code in this page.
  *
- * Copyright (c) 2012-2017 RockStor, Inc. <http://rockstor.com>
+ * Copyright (c) 2012-2021 RockStor, Inc. <http://rockstor.com>
  * This file is part of RockStor.
  *
  * RockStor is free software; you can redistribute it and/or modify
@@ -957,19 +957,6 @@ $(document).ready(function() {
             $(target).removeClass('in');
             $(target).css('height', '0');
         }
-    });
-
-    // donate menu item handler
-    $('#donate_nav').click(function(event) {
-        if (event) {
-            event.preventDefault();
-        }
-        $('#donate-modal').modal('show');
-    });
-
-    // donate modal paypal donate button handler
-    $('#donate-modal #paypal_donate_button').click(function(event) {
-        $('#donate-modal').modal('hide');
     });
 
     /********** Websockets **************/
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst
@@ -56,7 +56,6 @@
       <li><a href="mailto:support@rockstor.com" target="_blank">Team email</a></li>
     </ul>
   </li>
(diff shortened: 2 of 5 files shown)


FroggyFlox (approved):
Looks all good to me, @phillxnet!

I could verify the absence of the "Donate" entry in the top menu. For the various changes in the "Software Updates" page, however, I couldn't see the new text upon activating the Testing channel. The activation went well, but I just cannot see the green div box. I'm not sure why, but it's not related to your PR as I can't see the pre-PR box on a master build either. My guess is that I'm not looking at that the right way.

phillxnet:
@FroggyFlox Re:
> My guess is that I'm not looking at that the right way.
That's a good one. Made me chuckle.

So the 'story' on this one is that I've updated the existing text but made no changes to the logic of when and where it's displayed. The main change I made only shows up when there are updates. This could be improved upon but that can be in another followup pr. To test this pr, if you build from source and subscribe to any channel, then any rpm and it's associated changelog will be considered an update. This is due to source builds being considered unknown. And rpm releases not so. Hence any rpm is an 'update' to any source build and so should trigger the display of the sited green background text.

Sorry I should have put this in a caveate to reproduce the test scenario I worked under here.

We should definitely also surface donate/community nature also, with links, pre-update availability. But we do this, without links, in our initial popup on first login. And I believe also in the subsequent dashboard prompts to update when no update channel has been selected.

Thanks for taking an initial look at this. And again apologies for my not specifying a reproducer for the indicated changes.

phillxnet:
@FroggyFlox I'm going to merge this as is since it at least addresses the removal of the now legacy paypal button legacy and updates our existing text. I think we could further surface the additional info/lins added in systems that have yet to have a pending update but that can be done in a follow-up pr.

Thanks again for the check. This looks to at least be a safe set of changes at least and does remove an external call which is always nice and reduces the browser code loaded for every page, if only slightly.

PR #2265, opened 2021-01-26
Specify --lines=0 when running systemctl status (#2267)

### LINKED ISSUE #2267: Long delay loading web UI pages due to systemctl status usage
As proposed in #2265:

> I am trying out Rockstor on a Raspberry Pi 3 running openSUSE Leap 15.2. The system root is installed to an old USB hard drive, and there is some noticeably long delays here and there, especially in the web UI.
> 
> Running `iotop`, some culprits I spot are the invocations of `systemctl status` that take several seconds to complete. As described by systemd/systemd#2460 and systemd/systemd#8864, the delay appears to be from waiting for journal lines to be retrieved and cached, and can be avoided by specifying `--lines=0`. The journal lines seem to be provided as a convenience for human inspection, and are not too useful to scripts; I suspect (but am not certain) that Rockstor makes no use of them.
> 
> With this change, "Services", "Samba", "NFS", and other service-related pages appear to load much faster—going from around 10 to 30 seconds (on first load) to 1 to 4 seconds.

chrstphrchvz:
Fixes #2267

I am trying out Rockstor on a Raspberry Pi 3 running openSUSE Leap 15.2. The system root is installed to an old USB hard drive, and there is some noticeably long delays here and there, especially in the web UI.

Running `iotop`, some culprits I spot are the invocations of `systemctl status` that take several seconds to complete. As described by systemd/systemd#2460 and systemd/systemd#8864, the delay appears to be from waiting for journal lines to be retrieved and cached, and can be avoided by specifying `--lines=0`. The journal lines seem to be provided as a convenience for human inspection, and are not too useful to scripts; I suspect (but am not certain) that Rockstor makes no use of them.

With this change, "Services", "Samba", "NFS", and other service-related pages appear to load much faster—going from around 10 to 30 seconds (on first load) to 1 to 4 seconds.

### DIFF
--- a/src/rockstor/system/services.py
+++ b/src/rockstor/system/services.py
@@ -66,11 +66,19 @@ def init_service_op(service_name, command, throw=True):
     if service_name not in supported_services:
         raise Exception("unknown service: {}".format(service_name))
 
-    return run_command([SYSTEMCTL_BIN, command, service_name], throw=throw)
+    arg_list = [SYSTEMCTL_BIN, command, service_name]
+    if command == "status":
+        arg_list.append("--lines=0")
+
+    return run_command(arg_list, throw=throw)
 
 
 def systemctl(service_name, switch):
-    return run_command([SYSTEMCTL_BIN, switch, service_name], log=True)
+    arg_list = [SYSTEMCTL_BIN, switch, service_name]
+    if switch == "status":
+        arg_list.append("--lines=0")
+
+    return run_command(arg_list, log=True)
 
 
 def set_autostart(service, switch):
@@ -169,15 +177,15 @@ def service_status(service_name, config=None):
     elif service_name in ("replication", "data-collector", "ztask-daemon"):
         return superctl(service_name, "status")
     elif service_name == "smb":
-        out, err, rc = run_command([SYSTEMCTL_BIN, "status", "smb"], throw=False)
+        out, err, rc = run_command([SYSTEMCTL_BIN, "--lines=0", "status", "smb"], throw=False)
         if rc != 0:
             return out, err, rc
-        return run_command([SYSTEMCTL_BIN, "status", "nmb"], throw=False)
+        return run_command([SYSTEMCTL_BIN, "--lines=0", "status", "nmb"], throw=False)
     elif service_name == "nut":
         # Establish if nut is running by lowest common denominator nut-monitor
         # In netclient mode it is all that is required, however we don't then
         # reflect the state of the other services of nut-server and nut-driver.
-        return run_command([SYSTEMCTL_BIN, "status", "nut-monitor"], throw=False)
+        return run_command([SYSTEMCTL_BIN, "--lines=0", "status", "nut-monitor"], throw=False)
     elif service_name == "active-directory":
         if config is not None:
             active_directory_rc = 1


phillxnet:
@chrstphrchvz This is a nice find. And you are a brave/patient person to be attempting our currently rather heavy weight code on a Pi3. But all the same this kind of enhancement can only help across the board. As per your suspicion I too don't think Rockstor uses these log entries so this could be a really nice fix.

Before a proper review, could you please make an associated issue, as per my comments in your other parallel pr, so we have the attribution mechanism in place.

Interesting find, I'd not appreciated this and the slower systems do help to find these slow spots.
@FroggyFlox Can you see an issue with this. I'll have a proper look myself soon hopefully.
If it's a free win it's a most welcome one as we are slowing down a tad.

Incidentally the Pi3 is not officially supported but mainly for performance reasons given the Pi4 is so much faster, as we are still rather heavy weight with hopefully many cheap fixes such as this to come in time. Plus we have the pending Python 3 move which should also help with speed/efficiency.

chrstphrchvz:
Fixes #2267

chrstphrchvz:
At the moment I think my use of an old 20GB 4200RPM hard drive (rather than SD card/flash drive/SSD) means disk I/O is the more important performance bottleneck: `iostat -x` often shows high %util for the disk, while CPU usage is rather low (definitely never stuck at 100%) and RAM usage is also low (about 300MB in use).

phillxnet:
@chrstphrchvz We now have quite a few Pi4 users on our forum and although this is a supported platform and significantly less 'strangled' than the Pi3, I think an open discussion of your findings thus far on the forum would be very interesting.
Please consider starting a new post on the forum with your findings thus far as it may well attract some additional input on 'stuff' we can do to improve the main performance bottleneck on lower end hardware. Which of course helps across the board anyway.
https://forum.rockstor.com/

FroggyFlox:
Very nice find indeed, @chrstphrchvz!

It remains to be verified, but all the calls to `systemctl status` I know simply use the resulting return code to decide whether or not the service is running so we indeed do not need the logs lines.

That actually made me look further into the `systemctl` `man` page and noticed the following:
> This function is intended to generate human-readable output. If you are looking for computer-parsable output, use show instead.

https://www.freedesktop.org/software/systemd/man/systemctl.html#status%20PATTERN%E2%80%A6%7CPID%E2%80%A6%5D

The problem with `systemctl show` is that is doesn't return a return code depending on the service's status, so we would need to search specifically for its *ActiveState* property and test against it, which might not be faster than @chrstphrchvz's solution here. It might be worth investigating further in the future, however.

All of that to say that this solution seems very good and likely to improve performances across the board, especially considering I don't think we're losing anything by implementing it.

phillxnet:
@chrstphrchvz & @FroggyFlox 
Having a look at this pr currently.

Post enabling all run_command logging via:

>         if log:

to

>         if True:

In a source build with this pr pre-applied, so debug logging enabled; we have the following on a service page refresh:

```
rleap15-2:~ # cat /opt/rockstor-dev/var/log/rockstor.log | grep -E "systemctl.*status"
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status rockstor --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status smartd --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl --lines=0 status smb
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status ntpd --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl --lines=0 status nut-monitor
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status snmpd --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status sshd --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status docker --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status shellinaboxd --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status rpcbind --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status sssd --lines=0
[28/Jan/2021 12:37:38] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status rpcbind --lines=0
[28/Jan/2021 12:37:39] DEBUG [system.osi:179] Running command: /usr/bin/systemctl status rockstor-bootstrap --lines=0
```
and via a non exhaustive exploration of the return code of the with and without '--lines=0' systemctl commands via:

> echo $?

I've not noticed any behaviour change from our code perspective, rc wise that is. So this is so far looking like a win/win, which is nice.

Also a slow but workable near real time output of these command logs was also observed via:

> watch -d 'tail -n 100 /opt/rockstor-dev/var/log/rockstor.log | grep -E "systemctl.*status"'

phillxnet:
@chrstphrchvz OK, this is looking very much like we should just pop it in. The variable parameter ordering bothers me a little for no good reason but it allows for simpler code which is always favoured.

About to merging as-is given it looks to be pretty safe and there is no functional change. Bar of course the noted, and most welcome performance improvements. This is a nice find and thanks again for taking the care and time to submit it; much appreciated. And thanks also for the excellent upstream issue links and reproducer info. Most helpful.

Gunni:
Hey, sorry for the necro, but I bumped into this report from the systemd [report linked above](https://github.com/systemd/systemd/issues/2460), I just wanted to suggest using the dbus interface to get service information, it is much faster in my opinion.

I made a gist [showing how this is done](https://gist.github.com/Gunni/b38ccefbf6beca26fe76a7ca5abcd08c) in case you are interested.

chrstphrchvz:
Thanks for the suggestion @Gunni, I do like the idea of retrieving things via Python APIs (especially official ones available from OS package managers) rather than parsing shell command output as Rockstor tends to.

It has been a while since I worked on (still-unfinished) contributions to Rockstor, though, so I do not know if I will get to try it anytime soon.

PR #2269, opened 2021-01-28
Fix forum title encoding re auto support topic creation link (#2150)

### LINKED ISSUE #2150: Create a forum support topic link has insufficient character escapes for title
On occasions, certain bug reports presented via our embedded "Houston, we've had a problem' page link (step 3) that normally helps in the creation of a pre-titled forum post results in a json output of the forum front page. On initial inspection it seems we are failing to adequately quote / escape Rockstor API error reports such that they do not translate, within the resulting URL, to forum URL requests. Hence if a Rockstor error contains for example:
```
&format=json
```
which is often the case, this is in turn interpreted by the forum to display in json.

And example of such a problematic report would be:
```
 Unknown internal error doing a GET to /api/netatalk?page=1&format=json&page_size=9000&count= 
```
or
```
Unknown internal error doing a GET to /api/users?page=1&format=json&page_size=9000&count=
```
We may well simply need to escape chars within suggested new topic titles thus:
```
?	%3F
=	%3D
&	%26
```

phillxnet:
Add forum title encoding handlebars helper to properly encode the auto generated support forum topic creation link. This avoids the possible prior behaviour of inadvertent forum api activation via url API crossover between quoted Rockstor API error messages, used in the title, and actual forum apis. Resulting in failed forum topic creation.

Fixes #2150
Ready for review.

N.B. Please note that during testing a forum account will be required. Ideally one without elevated privileges due to the nature of the failure state (pre-pr).

## Testing
Please see issue text comment https://github.com/rockstor/rockstor-core/issues/2150#issuecomment-769324695 for the testing method/reproducer and proof of fix.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/router.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/router.js
@@ -834,6 +834,15 @@ Handlebars.registerHelper('sidenav', function(condition) {
     }
     return new Handlebars.SafeString(html);
 });
+
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
+// attempt to conform to https://tools.ietf.org/html/rfc3986
+Handlebars.registerHelper('clean_forum_title', function(err_detail) {
+    return encodeURIComponent(err_detail).replace(/[!'()*]/g, function(c) {
+        return '%' + c.charCodeAt(0).toString(16);
+    });
+});
+
 //Initiate the router
 var app_router = new AppRouter;
 //###Render the view###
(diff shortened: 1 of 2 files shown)


FroggyFlox (approved):
@phillxnet,

I can build from source with this branch without problem, and I could reproduce the fix you detailed in the related comment. Upon opening the link in a new **private** browser tab, I was offered with sign-in/up dialog window from Discourse. I believe that would have failed prior to this PR...

The changes are small and thus very likely to be safe, despite fixing the encoding issue.
Thanks a lot!

phillxnet:
@FroggyFlox Thanks for the review. Much appreciated.

I also think this is a safe and useful addition and at least helps. I'll merge as-is since we can always re-visit. I think for example we could improve the user facing text here as well but all in good time.

PR #2271, opened 2021-01-30
repeated log errors on disabled quotas (#2236)

### LINKED ISSUE #2236: Repeated Log Errors for shares with disabled quotas
Probably just a bit of LOG housekeeping really, but currently shares with quotas disabled are being repeatedly logged as Errors.  Example extract from Rockstor Log:

[16/Dec/2020 18:27:33] ERROR [system.osi:174] non-zero code(1) returned by command: ['/usr/sbin/btrfs', 'qgroup', 'show', '/mnt2/ROOT/home']. output: [''] error: ["ERROR: can't list qgroups: quotas not enabled", '']

Logs are repeatedly filled with this 'ERROR', which isn't an error as ROOT pool share quotas are disabled by default, and other non-ROOT share quotas might also be disabled by user.  This information could be of use at times when tracking down other errors, so could these items be managed separately as DEBUG to avoid unnecessary filling of logs?

phillxnet:
Remove explicit logging of key 'btrfs qgroup show' commands to avoid excessive log spamming. As we have now supported the pool quota disabled state mostly successfully for some time we can remove the initial default of always logging these commands which are reported as errors by the underlying btrfs commands when quotas are disabled on the parent pool. And hence are at odds with their supported state within the Rockstor Web-UI. In most common cases we catch and log as info in addition to the prior 'raw' command log.

Prior explicit log parameters have been kept as explicit to ease reversing this change.

Includes minor comment typo.

Fixes #2236 
Ready for review.

### Pre pr
With a ROOT pool, quota disabled, a Shares refresh page produces the following log entry:

> [30/Jan/2021 18:33:29] ERROR [system.osi:196] non-zero code(1) returned by command: ['/usr/sbin/btrfs', 'qgroup', 'show', '/mnt2/ROOT/home']. output: [''] error: ["ERROR: can't list qgroups: quotas not enabled", '']

### Post pr
There is no ERROR message produced.

N.B. there are still error states that we wish to surface and they should still be logged or surfaced. But for the time being this patch reduces our mostly redundant a and potentially misleading 'ERROR' log spamming for known common btrfs quota disabled states.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -1151,7 +1151,7 @@ def qgroup_max(mnt_pt):
     :return: -1 if quotas not enabled, else highest 2015/* qgroup found or 0
     """
     try:
-        o, e, rc = run_command([BTRFS, "qgroup", "show", mnt_pt], log=True)
+        o, e, rc = run_command([BTRFS, "qgroup", "show", mnt_pt], log=False)
     except CommandException as e:
         # disabled quotas can result in o = [''], rc = 1 and e[0] =
         emsg = "ERROR: can't list qgroups: quotas not enabled"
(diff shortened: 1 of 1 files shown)


FroggyFlox (approved):
@phillxnet ,

Looks all good on my end as well! It builds without error, and I do not see the repeated errors related to quota disabled in the logs anymore.

Thanks a lot!

phillxnet:
@FroggyFlox Thanks for the double check/review. Much appreciated.
I'll go ahead and merge.

PR #2283, opened 2021-03-07
Replace orphaned django-ztask with Huey (#2276)

### LINKED ISSUE #2276: Replace orphaned django-ztask with Huey
Our current background process scheduler django-ztaskd has been an orphaned project for some time. This in turn hampers our Django and Python update plans, detailed in the following issues and associated comments:
### "Update Django within current Python 2 constrains" #2254

https://github.com/rockstor/rockstor-core/issues/2254#issuecomment-771667824
https://github.com/rockstor/rockstor-core/issues/2254#issuecomment-771823905
https://github.com/rockstor/rockstor-core/issues/2254#issuecomment-771945822

### "move to python 3" #1877
https://github.com/rockstor/rockstor-core/issues/1877#issuecomment-619563667

This issue proposes Huey as a hopefully near fit, well maintained, Python 2.7/3, and Django native without plugins replacement:
https://github.com/coleifer/huey
https://pypi.org/project/huey/
Latest doc links: https://huey.readthedocs.io/en/latest/index.html
v1.10.1 (july 2018) started auto testing on Python 3.7: main readme only mentions 3.4. So good on newer Python.
v2.1.0 (June 2019) "Fix gzip compatibility issue when using Python 2.x."
https://github.com/coleifer/huey/blob/master/CHANGELOG.md#210
So presumably still Python 2.7 at that point at least.

This issue was originally envisaged as a sub part of the above referenced Django update. That way we start out with a far newer version, given the project states supporting/testing only on currently supported versions of Django. So if done before our update we may end up implementing a considerably younger, and possibly more problematic version. With the added work of then possibly having to migrate to newer versions once we update our Django. However we have a strict co-dependence chicken-egg situation here. And as proposed in the original text of the above Django update, we may very well require a testing branch within which we partially complete the Django update with significant breakage, and then follow-up with this issue to complete our ability to achieve even running rockstor services.

phillxnet:
As django-ztask is now orphaned and no longer compatible with newer Django or Python, this commit proposes replacing django-ztask, project wide, with Huey tasks which is both python 2 and 3 compatible as well as apparently functional with our django version and far newer variants.

Fixes #2276 

Ready for review.

As a rather large, but required, change set it is appreciated that a full review is non trivial. However a quick functional double-check of pool balance/add/remove/re-raid and basic rock-on functionality would be appreciated as bar the added balance end time there is no functional change within this pr. Note the new /opt/rockstor/var/log/huey.log file which give live details of Huey task start/end/name/worker.

## Includes:
- Initial Huey instantiation and configuration within Django.
- Maintain existing 'program'/service name within supervisord config and within smart-manager db: to minimise required changes.
- Use rockon task callers result handler to set rockon.taskid.
- Remove storageadmin django_ztasks model migration.
- Transition task() decorators & invocation to Huey's task/db_task.
- Added Huey task signals/logging.
- Includes some black and string.format() improvements.
- Add comments re future Huey pipeline use within config restore.
- Remove defunct service-manager and job-dispatcher cmd vars. In commit 36eeee8514dfe32cc002d71dc3101bcc61d1674d service-manager was removed. And in a9c36db2971f48c33df0652fe90060122b3d5e6a job-dispatcher was added unused when django-ztask was introduced.
- Explicitly specify worker-type for run_huey to aid future re-config.
- Development debug logging
- Update unit tests for Huey task result handler related changes.

### Pool Balance specific:
- Move pool balance and resize from django_ztask to huey task.
- Move to Python 2 & 3 compatible unicode - pool_balance status. When we save byte code strings (Python 2 default) to Django db filed, they are stored and retrieved as unicode strings. Move to explicit unicode to ease Python 3 transition where string literals are unicode by default.
- Minor refactor to ease state change requirement to help reduced redundant db access in pool_balance status update.
- Add clause to not update state if proposed and current are identical.
- Refactor pool balance status processing - reduces db writes.
- Reduce db access - pool_balance status update via update_fields.
- Add/re-enable previously dis-functional balance end time via django timezone.now() and Huey task signals.
- Skip costly pool balance model update if it is not required.
- Includes additional debug logging for balance status processing.

### Rock-on specific:
- In task calls we were calling apis just to update status, these calls would in turn invoke db calls/locks that we already had. This could lead to un-resolvable circular locking issues so use our existing task rockon db handles where every possible first.
- Move to huey task invocation syntax in rockon_id.py in turn removing the Python 3.7 reserved keyword of 'async'.
- Do rock-on config updates within a single huey task.
- Add comments to RockOn model to clarify use of state/status.
- Clarify via comments our slightly obscure use of global() re fail over for generic within rockons.
- Avoid redundant rockon db object retrieval in rockon_status().
- We previously retrieved a rock-on name from a current rockon db object then used this to re-lookup the rock-on db object. Given rockon_status is not run within it's own task we can pass the original object and save on db work.
- Add Huey taskid field to RockOn model to ease sanity checks. Previously our django_ztask uuid index db field was set to it's associated rockon db id. As Huey tasks assign their own uuid we add a field to our RockOn model to aid sanity checks re state/status/task and centralise these three indicators in the same model. Add associated db migration file.
- Manage new Huey taskid field by setting & nulling/assigning None in all existing 'finally' clauses in @db_task decorated rockon functions to aid in this fields management.
- Pass parent huey.task to rockon call_local() instances to have consistent and meaningful Huey logging via signals etc. By passing our parent task through we ensure the task association to the child functions which would otherwise label their associated rockon db objects with a None task: miss-representing the parent tasks invocation as local within the related rockon model.
- Preserve and log rockon.taskid - non live rockon config update. We currently call uninstall install via call_local when doing a rockon non live config update. For consistency / predictability, do not remove the taskid from the rockon model when it is in pending_update. This way we can tell that the rockon is mid update task.
- Add single instant task lock/limiter to restore_config(). As this can be a long running task, especially when re-installing / restoring rock-ons, we ensure that only a single instance of the overall controlling task is active at any one time.
- Fix bug re rockon_transition_checker() only waiting for 2 secs.
- Make rockon_transition_checker aware of pending rockon tasks.
- Increase rock-on restore per instance timeout from 30 to 46 seconds. In many instance with common download and cpu speeds this wait time is more appropriate as the prior 30 seconds timeout would not infrequently abandon early otherwise successful rock-on restores.

#### Minor refactor re rock-ons restore.
Move away from dict parameter modification to explicit returns with intergrated shallow copies of the associated rockons dict within helper functions to aid in isolating their actions. Also note an attempt to remove modifications to the base of iterations in for loops involving this dict. Precautionary moves to help improve data integrity and clarify operations in this complex area of the code. It may be that we need deep copies in some instance where we as yet only have shallow copies.

We also move our rock-on restore tasks 'up-one' to the caller, avoiding the use of Huey pipelines/chains and hopefully simplifying overall. This also helps to accommodate the explicit return values now employed in our rock-on restore sub-processes.

## General notes:
Given Huey non scheduled tasks are still subject to a minor delays of around a second or so within the runner it is best we take the earliest opportunity to stamp our rockon model with it's taskid. This may make the later task code update of this taskid element redundant but it is believed that updating this property at both opportunities should make for a more robust approach: i.e. task fails before scheduled task code is executed. We still have evidence of the prior event. But that info will in turn be overwritten upon any future task code execution, thus self repairing. This may come into play as we move towards automated
task retry capability within Huey.

If we only change a subset of model fields, explicitly update only those fields via model.save()'s update_fields parameter. This helps to reduce db contention as we more more towards proper multi-threaded functioning.

Post the introduction of Huey the data-collector, also run via supervisord, failed all Django model access due to threading constrains.

> "DatabaseError: DatabaseWrapper objects created in a thread can only be used in that same thread. ... "

in supervisord_data-collector_stderr.log. The fix was to apply the greenlet required monkey patching prior to the existing, and previously functional, fs.btrfs import. It is hypothesized that the newer wrapping of the @task Huey decorator within fs.btrfs
invoked this otherwise potential incompatibility.

Traceback was used from Huey TaskException in balances Errors column because initially the 'error' key was chosen and worked well for some issues but for the more common 'real' btrfs errors 'error' simply surfaces: "CommandException()" Where was a full traceback, although messy, does report the actual cause of the CommandException(), i.e. as part of our sterr element:

> "stderr = ["ERROR: error removing device '/dev/disk/by-id/ata-QEMU_HARDDISK_QM00003': No space left on device", '']"

Along with a full traceback, involving the huey egg version, this looks to serve our needs better given we are on new turf here with Huey.

A pending Huey task, once execution begins, is no longer pending. So this does not necessarily mean we have a failed state. Use info to help with user diagnosis.

## Testing
All task related functions re pool balance and rock-ons were tested to behave comparably or better than with our prior django-ztask implementation. Note that the rockstor restart facility within services was not tested as is not considered a show-stopper.

Config restore was tested with the inclusion of 3 typical rock-ons, Plex, Duplicati - by Linuxserver.io, and Gollum. A freshly build source rockstor instal system had this config uploaded post pool import with docker in a disabled state. Post config play-back the rock-on system was re-enabled and configured and all 3 rock-ons were installed and enabled. N.B. under some circumstances an existing issue was observed where docker would report unknown subvolumes. The resolution was to force delete the imported rock-ons-root and re-create a subvol by the same name and re-play the config-backup. Note also that with our current 2 Huey workers there is a practical limit of 2 concurrent docker tasks and attempting a 3rd results in an explanatory Web-UI error message to try again later. This is by design and moving to say 3 concurrent Huey workers was considered overkill. Also note that a single Huey worker is not sufficient for a restore as we depend upon a certain amount of concurrency with our current config restore setup: i.e. the restore process is a task that initiates the rock-on restore as it's last element. As Huey tasks return immediately this is not a problem, however during rock-on restore (itself a Huey task) we in turn require another Huey worker to execute each rock-on install. Hence the minimum requirement of 2 Huey workers.

Unit tests were also trivially modified and thereafter gave the expected result of:

```
./bin/test -v 2 -p test_*

Ran 216 tests in 16.376s

OK
```

## Caveats

This like-for-like replacement of django-ztask by Huey only involved minor modifications to our Web-UI update mechanism; mainly concerned with db optimisations and robustness. And so shares some of the prior hick-ups. However given we intend to move to Django Streams 2, which will involve significant changes to our Web-UI update nature, trivial bug issues remaining in Web-UI update for pool balance and rock-on functionality are also not considered show stoppers here. The main aim is to free our task requirement from the technical dept of django-ztask and allow us to proceed with dependant issues such as are detailed in the referenced above. Allowing us to kick-off our 4 stable variant with more maintainable task technologies.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -101,9 +101,10 @@ gunicorn_cmd = ${buildout:directory}/bin/gunicorn --bind=${init-gunicorn:bind}:$
 smart_manager_cmd = ${buildout:directory}/bin/sm
 replicad_cmd = ${buildout:directory}/bin/replicad
 dc_cmd = ${buildout:directory}/bin/data-collector
-sm_cmd = ${buildout:directory}/bin/service-monitor
-jd_cmd = ${buildout:directory}/bin/job-dispatcher
-ztask_cmd = ${buildout:directory}/bin/django ztaskd --noreload -l DEBUG -f ${supervisord-conf:logdir}/ztask.log
+# Huey ommandline options take precedence over those in settings.py.
+# https://huey.readthedocs.io/en/latest/django.html#running-the-consumer
+# "... Worker type, must be “thread”, “process” or “greenlet”. The default is thread, ..."
+huey_cmd = ${buildout:directory}/bin/django run_huey --workers 2 --worker-type thread --logfile ${supervisord-conf:logdir}/huey.log
 
 [django-settings-conf]
 rootdir = ${buildout:directory}/src/rockstor
--- a/conf/django-hack
+++ b/conf/django-hack
@@ -16,7 +16,7 @@ sys.path[0:0] = [
     join(base, 'eggs/django_braces-1.8.0-py2.7.egg'),
     join(base, 'eggs/django_oauth_toolkit-0.9.0-py2.7.egg'),
     join(base, 'eggs/django_pipeline-1.6.9-py2.7.egg'),
-    join(base, 'eggs/django_ztask-0.1.5-py2.7.egg'),
+    join(base, 'eggs/huey-2.3.0-py2.7.egg'),
     join(base, 'eggs/djangorecipe-1.9-py2.7.egg'),
     join(base, 'eggs/djangorestframework-3.1.1-py2.7.egg'),
     join(base, 'eggs/gevent-1.1.2-py2.7-linux-x86_64.egg'),
--- a/conf/settings.conf.in
+++ b/conf/settings.conf.in
@@ -20,6 +20,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 # Django settings for Rockstor project.
 import os
 import subprocess, distro
+from huey import SqliteHuey
 
 DEBUG = ${django-settings-conf:debug}
 TEMPLATE_DEBUG = DEBUG
@@ -158,7 +159,7 @@ INSTALLED_APPS = (
     'rest_framework',
     'smart_manager',
     'oauth2_provider',
-    'django_ztask',
+    'huey.contrib.djhuey',
 )
 
 STATICFILES_STORAGE = 'pipeline.storage.PipelineCachedStorage'
@@ -419,7 +420,7 @@ ROCKONS = {
 	'local_metastore': '${buildout:depdir}/rockons-metastore',
 }
 
-ZTASKD_URL = 'ipc:///var/run/rockon-ztaskd'
+HUEY = SqliteHuey(filename='${buildout:depdir}/rockstor-tasks-huey.db')
 
 TASK_SCHEDULER = {
 	       'max_log': 100 #max number of task log entries to keep
--- a/conf/start_huey.bash.in
+++ b/conf/start_huey.bash.in
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# incomplete shell wrapper for potential suture invocation of huey:
+# https://github.com/coleifer/huey/issues/88
+
+echo "Starting Rockstor-huey as `whoami`"
+
+# Start Huey
+exec python2 ${buildout:directory}/manage.py run_huey
--- a/conf/supervisord-prod.conf.in
+++ b/conf/supervisord-prod.conf.in
@@ -122,16 +122,17 @@ stderr_logfile=${supervisord-conf:logdir}/supervisord_%(program_name)s_stderr.lo
 stderr_logfile_maxbytes=1MB   ; max # logfile bytes b4 rotation (default 50MB)
 stderr_logfile_backups=10     ; # of stderr logfile backups (default 10)
 
-; ztask daemon
+; huey daemon - program name maintained as legacy to match existing service name
 [program:ztask-daemon]
 environment=DJANGO_SETTINGS_MODULE=settings
-command=${supervisord-conf:ztask_cmd} ; the program (relative uses PATH, can take args)
+command=${supervisord-conf:huey_cmd} ; the program (relative uses PATH, can take args)
 process_name=%(program_name)s ; process_name expr (default %(program_name)s)
 numprocs=1                    ; number of processes copies to start (def 1)
 priority=200
 autostart=true                ; start at supervisord start (default: true)
-autorestart=unexpected        ; whether/when to restart (default: unexpected)
-startsecs=2                   ; number of secs prog must stay running (def. 1)
+; autorestart=unexpected        ; whether/when to restart (default: unexpected)
+autorestart=true
+startsecs=10                   ; number of secs prog must stay running (def. 1)
 startretries=3                ; max # of serial start failures (default 3)
 exitcodes=0,2                 ; 'expected' exit codes for process (default 0,2)
 stopsignal=TERM               ; signal used to kill process (default TERM)
@@ -142,3 +143,5 @@ stdout_logfile_backups=10     ; # of stdout logfile backups (default 10)
 stderr_logfile=${supervisord-conf:logdir}/supervisord_%(program_name)s_stderr.log        ; stderr log path, NONE for none; default AUTO
 stderr_logfile_maxbytes=1MB   ; max # logfile bytes b4 rotation (default 50MB)
 stderr_logfile_backups=10     ; # of stderr logfile backups (default 10)
+; Causes supervisor to send the termination signal (SIGTERM) to the whole process group.
+;stopasgroup=true
--- a/conf/supervisord.conf.in
+++ b/conf/supervisord.conf.in
@@ -121,16 +121,17 @@ stderr_logfile=${supervisord-conf:logdir}/supervisord_%(program_name)s_stderr.lo
 stderr_logfile_maxbytes=1MB   ; max # logfile bytes b4 rotation (default 50MB)
 stderr_logfile_backups=10     ; # of stderr logfile backups (default 10)
 
-; ztask daemon
+; huey daemon - program name maintained as legacy to match existing service name
 [program:ztask-daemon]
 environment=DJANGO_SETTINGS_MODULE=settings
-command=${supervisord-conf:ztask_cmd} ; the program (relative uses PATH, can take args)
+command=${supervisord-conf:huey_cmd} ; the program (relative uses PATH, can take args)
 process_name=%(program_name)s ; process_name expr (default %(program_name)s)
 numprocs=1                    ; number of processes copies to start (def 1)
 priority=200
 autostart=true                ; start at supervisord start (default: true)
-autorestart=unexpected        ; whether/when to restart (default: unexpected)
-startsecs=2                   ; number of secs prog must stay running (def. 1)
+; autorestart=unexpected        ; whether/when to restart (default: unexpected)
+autorestart=true
+startsecs=10                   ; number of secs prog must stay running (def. 1)
 startretries=3                ; max # of serial start failures (default 3)
 exitcodes=0,2                 ; 'expected' exit codes for process (default 0,2)
 stopsignal=TERM               ; signal used to kill process (default TERM)
@@ -141,3 +142,5 @@ stdout_logfile_backups=10     ; # of stdout logfile backups (default 10)
 stderr_logfile=${supervisord-conf:logdir}/supervisord_%(program_name)s_stderr.log        ; stderr log path, NONE for none; default AUTO
 stderr_logfile_maxbytes=1MB   ; max # logfile bytes b4 rotation (default 50MB)
 stderr_logfile_backups=10     ; # of stderr logfile backups (default 10)
+; Causes supervisor to send the termination signal (SIGTERM) to the whole process group.
+;stopasgroup=true
--- a/prod-buildout.cfg
+++ b/prod-buildout.cfg
@@ -55,8 +55,10 @@ gunicorn_cmd = ${buildout:depdir}/bin/gunicorn --bind=${init-gunicorn:bind}:${in
 smart_manager_cmd = ${buildout:depdir}/bin/sm
 replicad_cmd = ${buildout:depdir}/bin/replicad
 dc_cmd = ${buildout:depdir}/bin/data-collector
-sm_cmd = ${buildout:depdir}/bin/service-monitor
-ztask_cmd = ${buildout:depdir}/bin/django ztaskd --noreload --replayfailed -f ${supervisord-conf:logdir}/ztask.log
+# Huey ommandline options take precedence over those in settings.py.
+# https://huey.readthedocs.io/en/latest/django.html#running-the-consumer
+# "... Worker type, must be “thread”, “process” or “greenlet”. The default is thread, ..."
+huey_cmd = ${buildout:directory}/bin/django run_huey --workers 2 --worker-type thread --logfile ${supervisord-conf:logdir}/huey.log
 input = ${buildout:directory}/conf/supervisord-prod.conf.in
 
 [django-settings-conf]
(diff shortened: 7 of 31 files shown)


phillxnet:
@FroggyFlox Re production build path error, post fix commit we now have:
```
ps -aef --forest
...
root      3629     1  0 10:13 ?        00:00:00 /usr/bin/python2 /opt/rockstor/bin/supervisord -c /opt/rockstor/etc/supervisord.conf
root      3674  3629  0 10:13 ?        00:00:00  \_ nginx: master process /usr/sbin/nginx -c /opt/rockstor/etc/nginx/nginx.conf
nginx     3679  3674  0 10:13 ?        00:00:00  |   \_ nginx: worker process
nginx     3680  3674  0 10:13 ?        00:00:00  |   \_ nginx: worker process
root      3675  3629  0 10:13 ?        00:00:00  \_ /usr/bin/python2 /opt/rockstor/bin/gunicorn --bind=127.0.0.1:8000 --pid=/run/gunicorn.pid --workers=2 --log-file=/opt/rockstor/var/log/gunicorn.log --pythonpath=/opt/rockstor/src/rockstor --timeout=120 --graceful-timeout=120 wsgi:application
root      3699  3675  0 10:13 ?        00:00:00  |   \_ /usr/bin/python2 /opt/rockstor/bin/gunicorn --bind=127.0.0.1:8000 --pid=/run/gunicorn.pid --workers=2 --log-file=/opt/rockstor/var/log/gunicorn.log --pythonpath=/opt/rockstor/src/rockstor --timeout=120 --graceful-timeout=120 wsgi:application
root      3700  3675  0 10:13 ?        00:00:00  |   \_ /usr/bin/python2 /opt/rockstor/bin/gunicorn --bind=127.0.0.1:8000 --pid=/run/gunicorn.pid --workers=2 --log-file=/opt/rockstor/var/log/gunicorn.log --pythonpath=/opt/rockstor/src/rockstor --timeout=120 --graceful-timeout=120 wsgi:application
root      3677  3629  0 10:13 ?        00:00:00  \_ /usr/bin/python2 /opt/rockstor/bin/data-collector
root      3678  3629  0 10:13 ?        00:00:03  \_ /usr/bin/python2.7 /opt/rockstor/bin/django run_huey --workers 2 --worker-type thread --logfile /opt/rockstor/var/log/huey.log
```
Which is looking as intended now. Previously there was no run_huey entry on the production build due to my cut-paste error.

This highlights yet more kipple we should tend to as we work our way through the remaining technical dept. I.e. redundant configuration files/profiles etc.

FroggyFlox:
@phillxnet ,

I can confirm that it builds successfully on Leap 15-2 fully updated and applies the migrations:
```
rockdev:~ # /opt/build/bin/django showmigrations
admin
 [ ] 0001_initial
auth
 [X] 0001_initial
 [X] 0002_alter_permission_name_max_length
 [X] 0003_alter_user_email_max_length
 [X] 0004_alter_user_username_opts
 [X] 0005_alter_user_last_login_null
 [X] 0006_require_contenttypes_0002
contenttypes
 [X] 0001_initial
 [X] 0002_remove_content_type_name
oauth2_provider
 [X] 0001_initial
 [ ] 0002_08_updates
sessions
 [ ] 0001_initial
sites
 [ ] 0001_initial
smart_manager
 [ ] 0001_initial
 [ ] 0002_auto_20170216_1212
storageadmin
 [X] 0001_initial
 [X] 0002_auto_20161125_0051
 [X] 0003_auto_20170114_1332
 [X] 0004_auto_20170523_1140
 [X] 0005_auto_20180913_0923
 [X] 0006_dcontainerargs
 [X] 0007_auto_20181210_0740
 [X] 0008_auto_20190115_1637
 [X] 0009_auto_20200210_1948
 [X] 0010_sambashare_time_machine
 [X] 0011_auto_20200314_1207
 [X] 0012_auto_20200429_1428
 [X] 0013_auto_20200815_2004
 [X] 0014_rockon_taskid
```

`run_huey` is indeed triggered and running:
```
rockdev:~ # ps -aef --forest
UID        PID  PPID  C STIME TTY          TIME CMD
(...)
root     22605     1  0 10:23 ?        00:00:00 /usr/bin/python /opt/build/bin/supervisord -c /opt/build/etc/supervisord.conf
root     22611 22605  0 10:23 ?        00:00:00  \_ nginx: master process /usr/sbin/nginx -c /opt/build/etc/nginx/nginx.conf
nginx    22615 22611  0 10:23 ?        00:00:00  |   \_ nginx: worker process
nginx    22616 22611  0 10:23 ?        00:00:00  |   \_ nginx: worker process
root     22612 22605  0 10:23 ?        00:00:00  \_ /usr/bin/python /opt/build/bin/gunicorn --bind=127.0.0.1:8000 --pid=/run/gunicorn.pid --workers=2 --log-file=/opt/build/var/log/gunicorn.log --pythonpath=/opt/build/src/rockstor --timeout=120 --graceful-timeout=120 wsgi
root     22631 22612  0 10:23 ?        00:00:00  |   \_ /usr/bin/python /opt/build/bin/gunicorn --bind=127.0.0.1:8000 --pid=/run/gunicorn.pid --workers=2 --log-file=/opt/build/var/log/gunicorn.log --pythonpath=/opt/build/src/rockstor --timeout=120 --graceful-timeout=120 
root     22632 22612  0 10:23 ?        00:00:00  |   \_ /usr/bin/python /opt/build/bin/gunicorn --bind=127.0.0.1:8000 --pid=/run/gunicorn.pid --workers=2 --log-file=/opt/build/var/log/gunicorn.log --pythonpath=/opt/build/src/rockstor --timeout=120 --graceful-timeout=120 
root     22613 22605  0 10:23 ?        00:00:00  \_ /usr/bin/python /opt/build/bin/data-collector
root     22614 22605  1 10:23 ?        00:00:02  \_ /usr/bin/python2.7 /opt/build/bin/django run_huey --workers 2 --worker-type thread --logfile /opt/build/var/log/huey.log
```
and shows as such in the logs:
```
rockdev:~ # cat /opt/build/var/log/huey.log 
[2021-03-08 10:24:01,284] INFO:huey.consumer:MainThread:Huey consumer started with 2 thread, PID 22614 at 2021-03-08 15:24:01.284023
[2021-03-08 10:24:01,284] INFO:huey.consumer:MainThread:Scheduler runs every 1 second(s).
[2021-03-08 10:24:01,285] INFO:huey.consumer:MainThread:Periodic tasks are enabled.
[2021-03-08 10:24:01,285] INFO:huey.consumer:MainThread:The following commands are available:
+ storageadmin.views.rockon_helpers.uninstall
+ storageadmin.views.rockon_helpers.update
+ smart_manager.views.ztask_helpers.ztask_helpers.restart_rockstor
+ storageadmin.views.config_backup.restore_rockons
+ fs.btrfs.start_resize_pool
+ storageadmin.views.rockon_helpers.install
+ storageadmin.views.rockon_helpers.stop
+ storageadmin.views.config_backup.restore_config
+ fs.btrfs.start_balance
+ storageadmin.views.rockon_helpers.start
```

I'm now going to try testing a few things...

phillxnet:
@FroggyFlox Thanks, much appreciated.

FroggyFlox:
@phillxnet ,

I've now tested a few things:

#### Rock-ons
Install/Start/Stop/Uninstall... all works as intended. I've also tried to "push the system" by starting several STOP tasks one after the other without waiting for the first one to finish and it was all handled beautifully in the back-end.
Restore config with 3 rock-ons installed: all works beautifully as well. Nothing to report there!

#### Pool
I've tested adding 2 disks, removing 1, changing raid level, all worked nicely! I've only found one quirk, but I don't think it's related to this PR:
1. One pool with 3 disks, in raid1.
2. Resize/ReRaid to change raid1 --> single
3. Right away (do not wait for the previous step to finish), Resize/Reraid to remove 2 of the 3 disks. The intent was to test queuing of the tasks. I have no idea of this behaves prior to this PR, unfortunately, but I was expecting this to struggle (understandably).
4. Step 3 fails: 
```
[2021-03-08 11:53:08,642] ERROR:huey:Worker-2:Unhandled exception in task 537be736-bc43-47fd-bb47-772e2ac5b55b.
  File "/opt/build/eggs/huey-2.3.0-py2.7.egg/huey/api.py", line 360, in _execute
    task_value = task.execute()
  File "/opt/build/eggs/huey-2.3.0-py2.7.egg/huey/api.py", line 724, in execute
    return func(*args, **kwargs)
  File "/opt/build/src/rockstor/fs/btrfs.py", line 1787, in start_resize_pool
    raise e
CommandException: Error running a command. cmd = /usr/sbin/btrfs device delete /dev/disk/by-id/ata-QEMU_HARDDISK_cccData2 /dev/disk/by-id/ata-QEMU_HARDDISK_eeeData3 /mnt2/data_pool. rc = 1. stdout = ['']. stderr = ["ERROR: error removing device '/dev/disk/by-id/ata-QEMU_HARDDISK_eeeData3': unable to go below two devices on raid1", '']
```
All of this makes sense to me, and I just had to start a balance manually on the webUI (forced, as otherwise I couldn't due to another balance being paused or in process), which fixed the raid status. I could then repeat step 3, which proceeded without error.

Bottom line: it does look great to me, and I do feel a lot more confident in our task management now!

phillxnet:
@FroggyFlox That's excellent. Thanks for the double check.

Re the disk/pool re-raid thing there I think we can approach that in another pr. And yes, we may well have a little race thing going on as Huey tasks can take up to 3 seconds or so to actually start. But the commands that initiated them return immediately. So we may well have a pending task that we don't recognise there. I'd has something similar with the Rock-ons until I added pending Huey task awareness to them. We will likely have to do the same for the pool tasks.

Cheers. I'll merge this as-is and get to the release then.

Many thanks for the double check. This one was way larger than I had imagined but needs must and at least we have shed one of the most out dated elements of the project which has to be good. It's served us well but it's time we moved on to allow the Django/Python updates to continue.

PR #2295, opened 2021-04-09
Fetch Group management setting and add it to payload (#2294).

### LINKED ISSUE #2294: Not putting a group under rockstor management is not respected
Following [a report](https://forum.rockstor.com/t/deleting-a-user-newbie-question/7756?u=flox) by forum user *wdc* and subsequent attempts at reproducing the issue described therein, I noticed a quirk in the creation of a new group:

Indeed, the setting of the checkbox *Put this group under Rockstor management?* is **not** respected and the new group created ends up being managed by Rockstor regardless of the selected status of this checkbox.

After a brief investigation, it seems we simply fail to pull the value of this checkbox and include it in the payload; combined to the fact that we default this value to `True` in our `Group` model, we end up with all groups created from the webUI as being placed under Rockstor management. It seems we lost this setting somewhere in #530, fixing #401.

FroggyFlox:
Fixes #2294.
@phillxnet, ready for review.

As mentioned in #2294, all groups created from the webUI end up listed as "managed by Rosktor" regardless of the user choice in the group creation form. This turns out to be due to the fact that we do not actually fetch and send this info as part of the payload on form submission.
This pull request (PR) thus simply proposes to add this information to the payload:
1. We fetch the user's choice by getting the `"checked"` status of the corresponding checkbox (`id="admin"`).
2. We add it to the payload.

As this functionality was implemented before (we seem to have lost it during various changes), we already have all the backend logic in place and working so the changes included in this PR are all we need.

For demonstration, two groups were created from the webUI:
- **ManagedGroup**: the checkbox "Put this group under Rockstor management?" was **checked**
- **UnmanagedGroup**: the checkbox "Put this group under Rockstor management?" was **unchecked**.

Both are listed in their respective table:
image

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/add_group.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/add_group.js
@@ -72,6 +72,7 @@ AddGroupView = RockstorLayoutView.extend({
             submitHandler: function() {
                 var groupname = _this.$('#groupname').val();
                 var gid = _this.$('#gid').val() || null;
+                var admin = _this.$('#admin').prop("checked");
                 var group;
                 if (_this.groupname != null && _this.group != null) {
                     group = new Group({
@@ -92,7 +93,8 @@ AddGroupView = RockstorLayoutView.extend({
                     group = new tmpGroupModel();
                     group.save({
                         groupname: groupname,
-                        gid: gid
+                        gid: gid,
+                        admin: admin
                     }, {
                         success: function(model, response, options) {
                             _this.$('#group-create-form :input').tooltip('hide');


phillxnet:
@FroggyFlox Again, sorry for the hold up here on my part. I will take a look at this shortly.

Thanks for yet another excellent pull request presentation. Much appreciated.

phillxnet (approved):
@FroggyFlox Nice find and thanks for the fix. Looks like this, as you researched, was lost in a large rework some time ago. Much appreciated.

Reproduction of issue pre-fix:
Pre-fix-groups-created

Post fix proof:
Post-fix-groups-created

phillxnet:
@FroggyFlox Well done spotting this and thanks for the rapid fix in returning this long lost 'feature'.

Again, apologies for the hold up on my part here.

PR #2300, opened 2021-05-06
Fake Product UUID Case Sensitivity Incorrect on Rockstor 4 (#2290)

### LINKED ISSUE #2290: Fake Product UUID Case Sensitivity
It seems Rockstor 4, or rather OpenSUSE generates the same UUID but in a different case to Centos.

The file /opt/rockstor/src/rockstor/system/osi.py detects the fake UUID but in a case sensitive way.  This means that proper appliance IDs are not generated in RS4 for systems that present with a fake UUID.

The proper fix should make the Fake UUID detection process case insensitive.  An alternative fix would be to list fake uuids in both upper and lower cases.

phillxnet:
As Rockstor 4 is associated with newer kernel versions by default, which in turn now produce lowercase product uuid's, we fail in some cases to match against our mixed case tuple of know fake product uuids.

Fixes #2290 
@FroggyFlox Ready for review.

Fix by normalising on lowercase in line with the sited uuid standard of: rfc4122 as per the kernel.

Includes formatting clean-ups of the known fake product uuids.

N.B. no functional change is made to how an existing install, of the same major version, generates it's Appliance ID.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -2265,35 +2265,28 @@ def hostid():
     doesn't exist for any reason, generate a uuid like we used to prior to this
     change.
 
-    There's a lazy vendor problem where uuid is not set and defaults to
-    03000200-0400-0500-0006-000700080009 or
-    00020003-0004-0005-0006-000700080009. I don't know how these values are
-    populated, but empirically it seems to be just these
-    two. A non-persistent uuid is also generated in this case.
-
-    Observed motherboards with these fake non unique puuids:
-    - Examples of the first listed fake_puuids
-    ASRock N3700-ITX
-    ASRock C2550D4I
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox I'm inclined to merge this one now ready for our next release. It's, in fact, only a case change of our know fake product uuids and nothing more, and so makes not functional change to an existing systems Appliance ID: on re-install given the same hardware. That's was important.

We also now have Appman so the case change from 3 to 4 is not a show stopper. And we might indicate this requirement within the activation code dialog (which already now links to Appman); but I see this as a related but independent issue to what is addressed in this pull request.

FroggyFlox:
Sounds good! Sorry for not getting to it; I got distracted by our other matters at hand.

PR #2303, opened 2021-07-19
change rockons default url to https #2304

### LINKED ISSUE #2304: rockon url is http but rockstor.com is now https
Fix for default url for rockon pull url

#2303 

Hopefully even i cant mess that pull request up. haha.

korg:
As rockstor.com is now https only, changed the config setting for remote_metastore to https to save redirects

Fixes #2304

### DIFF
--- a/conf/settings.conf.in
+++ b/conf/settings.conf.in
@@ -415,7 +415,7 @@ UPDATE_CHANNELS = {
 }
 
 ROCKONS = {
-	'remote_metastore': 'http://rockstor.com/rockons',
+	'remote_metastore': 'https://rockstor.com/rockons',
 	'remote_root': 'root.json',
 	'local_metastore': '${buildout:depdir}/rockons-metastore',
 }


phillxnet:
@korg Nice. That was quick work.

I've added a link back to the issue you opened via the "Fixes" edit as then the associated issue is auto closed when we merge this pull request. I've also edited this pull request's title to be a little simpler. I'll do the same for the issue. This is nice as we use these issue/pr titles in our changelog and they need a certain format in order to display correctly in the Web-UI. That way we tend to the all important attribution.

Thanks again for stepping up to this task. We can test the resulting build soon hopefully; just in case.

It will be nice to not see all those redirects go whizzing by. But I guess we may have to wait a few years for this to take-hold; better sooner than later I guess. I'm popping this one into the pre Rockstor 4 release as I think it would be good to start out on the right foot with regard to our re-release as it were.

Marking as needs review. @FroggyFlox can you think of anything that might trip us up here?

phillxnet:
@korg Note that you have entered this pull request against your master branch korg:master. It is recommended that you always submit pull requests from a issue specific branch. But no worries for now as long as you don't submit any more pull requests or make any more changes to your master branch. Otherwise a review here would be invalidated as you master takes on a life of it's own. Hopefully we can get this one checked out quick so we can avoid the ago of starting a fresh.

Gain much appreciated for stepping up to this one.

FroggyFlox:
> Marking as needs review. @FroggyFlox can you think of anything that might trip us up here?

I can't see anything wrong at first; I'll try to find some time to test build that branch unless you beat me to it.

The one limitation I see is actually not related to this PR itself, but rather revealed by it. Indeed, if I'm correct, as this fix in included in `settings.conf.in`, it will only be "applied"/used at build time, isn't it? In this case, we should definitely include it now while Rockstor 4 is still in "testing" channel only.

We could think of improving our resilience against this sort of things by only specifying the server in settings.conf (`rockstor.com`) but validate the protocol (if `http`, then correct to `https`; if `https`, then leave as it) in `rockon.py`. Do you think it would be worth it or would it be just unnecessary complications?

The more I think about it, the more these kind of settings do not really belong in django settings, do they? Things like that are susceptible to change after build time after all...

phillxnet:
@FroggyFlox Re:

>  ... as this fix in included in `settings.conf.in`, it will only be "applied"/used at build time, isn't it?

Yes, it will only take effect post an rpm update with the fix rolled in.

> we should definitely include it now while Rockstor 4 is still in "testing" channel only.

Agreed. I've now popped the issue in the ISO milestone. I'll add this pull request also to be clearer.



> We could think of improving our resilience against this sort of things by only specifying the server in settings.conf (`rockstor.com`) but validate the protocol (if `http`, then correct to `https`; if `https`, then leave as it) in `rockon.py`. Do you think it would be worth it or would it be just unnecessary complications?

This sounds like overkill to me, but I like the idea of moving this definition to a runtime, such as what @korg did to prove the 'fix'/performance-improvement in the linked forum thread post here:
https://forum.rockstor.com/t/rockon-update-fails/7903/14
i.e. we move it to a 'constant' where it's needed so if need be folks can simply edit the top-of-file constant and restart the rockstor service and be away with say a mirror of the repo. Very much like we currently do via a sed to test the post pull request merge of the GitHub repo prior to publishing to production. This change to a constant may also help make that sed more robust actually.



> The more I think about it, the more these kind of settings do not really belong in django settings, do they? Things like that are susceptible to change after build time after all...

Yes, also agreed. So I propose that we either:
1:
- merge this pull request to get attribution and at least move what we have to https, and
- open a new issue closely defining our move from Django settings to constant for say the remote and local Rock-on repos (if the latter is not too much of a scope creep). This way we also allow folks to more easily use say an NFS mount to store their local rock-ons. This has been requested and moving to a runtime hard-coding rather than a build-time hard-coding facilitates that. We can then more easily extend later to allow for such things as Web-UI override of these so that our constant is the default unless a db entry is found for example.

2: 
- extend this pull request to include the above changes.

I don't like 2: as we are merging from korg:master which is in itself fragile. So I vote for 1: That way we can get this in with attribution and effect from effort and then start a new issue with an appropriate issue specific branch pull request.

Option 1 also means we have the fix in place right away, ready for the next rpm; in case we don't get any takers for the proposed move of definition from build time to run time.

FroggyFlox:
@phillxnet , I agree with option 1 as well.

phillxnet (approved):
@korg Thanks for diagnosing and stepping up to the fix for this one.

Only really a functional/real-life-performance review as we are just changing from http/https here and there-by avoiding the server-side redirect.

I built 2 roughly equal KVM Rockstor instances running on the same host. One with this fix/improvement and one without:

I then did an initial and 2 subsequent Rock-ons update button presses on each Rockstor instance. Waiting for each to complete in turn. The following timings were taken from the production server logs and the tests were done at a relatively quite time:

rpm 4.0.7-2303
```
26 seconds - initial update button
25 seconds - second update button
26 seconds - third update button
```
rpm 4.0.7-0
```
43 seconds - initial update button
43 seconds second update button
42 seconds third update button
```
And in an effort to control for network/load variations I did a fourth simultaneous update:
```
rpm 4.0.7-2303
26 seconds fourth simultaneous

rpm 4.0.7-0
43 seconds fourth simultaneous
```
The difference between http (with consequent redirect) and https was quite surprising to me so thanks again for looking into this. I had thought on our initial server side change to https recently, with consequent redirect, that the vast majority of the time was in the payload transit, and almost none in the chit-chat. Apparently this is not the case.

Obviously this is a significant difference and in cases where the round trip, even for the initial redirect, is larger we are getting into potential timeout scenarios; especially taking into account the additional seconds required for the acquired rock-ons info to then be processed, especially on slower machines and as our Rock-ons library grows.

As discussed in the comments we have more improvements to be done in this area but I'll merge this as it's an easy win for now.

PR #2307, opened 2021-07-26
Pass string with pkg to logger.info() #2306

### LINKED ISSUE #2306: rpm_build_info: 'rockstor' hardcoded in error
In the `rpm_build_info` function, the `rockstor` package hardcoded in one of the possible errors:

https://github.com/rockstor/rockstor-core/blob/f170ee56ce129d12e3b4c81534a39f72a767e4f0/src/rockstor/system/pkg_mgmt.py#L118

Since the function is getting the package name from the `pkg` argument, `pkg` should be used in the error instead, even though the function is currently only used to check for the rockstor package.

chrstphrchvz:
…rather than a string with "rockstor" hardcoded

Fixes #2306

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -115,7 +115,7 @@ def rpm_build_info(pkg):
         # By checking both the first error element and the second to last we
         # catch one yum waiting for another to release yum lock.
         if e.err[0] == emsg or e.err[-2] == emsg:
-            logger.info('No "rockstor" package found: source install?')
+            logger.info('No "{}" package found: source install?'.format(pkg))
             return version, date
         # otherwise we raise an exception as normal.
         raise e


chrstphrchvz:
Fixes #2306

phillxnet:
@chrstphrchvz As per my comment on your linked issue: Nice.
Thanks. Will take a proper look soon. Cheers.

phillxnet:
As expected, this looks to be a drop in replacement for what we already have. A source build, with pr pre-merged, results in the expected:
```
[17/Aug/2021 15:55:40] INFO [system.pkg_mgmt:118] No "rockstor" package found: source install?
[17/Aug/2021 15:55:53] INFO [system.pkg_mgmt:118] No "rockstor" package found: source install?
```
So I'll go ahead and merge this as part of your ongoing improvements in this area. Currently the test/text only relates to the specific instance of the rockstor package but as you say it can't harm / can only help to line this us for more general use.

Thanks for your ongoing work in this area. This should be included in our pending 4.0.8-0 testing channel rpm release.

PR #2320, opened 2021-09-26
Add another JMS567 controller serial obfuscation detection #2318

### LINKED ISSUE #2318: Add another JMS567 controller serial obfuscation detection
Thanks to @azilber in #2316 for reporting another incompatible multi-drive enclosure. We already have, in-place, detection for a number of 4, 5, and 8 bay ORICO USB 3.0 external enclosures that assign "000000000000" to all devices. In this latest report we have:

"Mediasonic 4 Bay Dock for 2.5' / 3.5' SATA HDD / SSD - USB 3.0 & eSATA (HFD1-SU3S2)"
which assignes "152D00539000" as serial for all of its member devices. 

From lsusb:
```
Bus 002 Device 002: ID 152d:0567 JMicron Technology Corp. / JMicron USA Technology Corp. JMS567 SATA 6Gb/s bridge
idVendor 0x152d JMicron Technology Corp. / JMicron USA Technology Corp.
idProduct 0x0567 JMS567 SATA 6Gb/s bridge
iProduct 11 USB to ATA/ATAPI Bridge
iSerial 5 152D00539000
```

See the following forum thread for some background on this drive serial obfuscation:
https://forum.rockstor.com/t/usb-disks-with-no-serials/4365/4

And the following source files:
Handlebars.registerHelper('checkSerialStatus', function (serial, diskName, opts):
https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/storageadmin/static/storageadmin/js/views/disks.js
https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/storageadmin/static/storageadmin/js/views/add_pool.js
https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/storageadmin/static/storageadmin/js/models/models.js

phillxnet:
"Mediasonic 4 Bay Dock for 2.5' / 3.5' SATA HDD / SSD - USB 3.0 & eSATA (HFD1-SU3S2)" reported as obfuscating contained drives serials and supplanting them all with 152D00539000. Blocked as per similarly observed behaviour in a number of 4, 5, and 8
bay ORICO USB 3.0 external enclosures which did the same but using the already detected "000000000000".

Note our report only relates to this device's use via USB 3.0. It's eSATA connection may not behave similarly.

Fixes #2318


phillxnet:
@FroggyFlox I'm intending to roll this into our pending 4.0.9-0 release given it may save some folks from using this devices first drive entry and ending up in a spot when they add another. So planning on merging this directly.

Bit of a force fix but until we know otherwise I think it's best.
Also no additional mechanisms added so not adding anything new bar an additional serial check. Syntax of addition is also relatively simple.

FroggyFlox:
@phillxnet , I completely agree and, most of all, trust you on that.
That makes me wonder: do we have a warning/note in our docs for people using such USB drive bays? Not necessarily regarding these specific models but something warning users about special requirements or precautions needed with these. Sorry I can't check at the moment.

phillxnet:
@FroggyFlox Re:
> do we have a warning/note in our docs for people using such USB drive bays?

No, and @azilber linked report of another one is what prompted me to collect the reports thus far in the issue text.

> Not necessarily regarding these specific models but something warning users about special requirements or precautions needed with these

It' a tricky one really as this same device may work jut fine under eSATA connection. But then if it presents the original serials then this 'block' will not enter into effect anyway. It's really only the multi disk enclosures that supplant one serial for all their hosted member disks. Currently, if we don't catch it then the first listed device is not 'noticed' but all subsequent ones are 'flagged' as repeat serial and given a "fake-serial-*". And if we are more sophisticated and mark them all as fake we will inadvertently catch single disk enclosures that use the same chip/approach. But in that case there is no issue: unless of course there are multiple instances of these single enclosures attached!

I'll get this merged for now. We have more work to do in this area, as always, but bit-by-bit.

PR #2326, opened 2021-11-18
Remove coverage library as unused and now requires Python 3 #2325

### LINKED ISSUE #2325: Pin version, or drop 'coverage' re Python 2-3 move
Thanks to @FroggyFlox for highlighting this issue. We currently depend upon coverage and do not specify a version. However the more recent versions of coverage have dropped Python 2.7 compatibility.

See our own issue regarding doing the same here: "move to python 3" #1877

References:
Coverage changelog: https://coverage.readthedocs.io/en/latest/changes.html#version-6-0b1-2021-07-18

It is proposed that we, for the time being, either:
1. Pin our dependency to our last know working version of 5.5 (2021-02-28). Which equates to the last 'stable' version released given that from the above 5.6b1 release notes we have: "These changes are part of 6.0."
2. Drop this dependency until we are the other side of our own Python 2 to 3 transition #1877.

phillxnet:
Although version 5.5 still has Python 2.7 compatibility we are not currently making use of coverage's facilities as we once did a few years ago. So remove for the time being and consider re-adding post our own Python 2 to 3 migration.

Fixes #2325 

@FroggyFlox Creating for test build purposes. I'll update shortly once I have more info post test builds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,6 @@ setup(
     install_requires=[
         'URLObject == 2.1.1',
         'chardet == 2.3.0',
-        'coverage',
         'distribute >= 0.6.35',
         'django == 1.8.16',
         'django-oauth-toolkit == 0.9.0',


phillxnet:
Bar the unrelated test error detailed in #2327 we have a successful source build and consequent service start.
And all bar the one indicated unit tests pass.

phillxnet:
@FroggyFlox I'm going to go ahead and merge this as it looks pretty safe to me. We know the last known working version so can return it with that pin if required. But this way we restore the build capability and reduce our dependencies.

Issue #2033, opened 2019-04-13
remove custom discourse docker treatment


phillxnet:
More recent versions of discourse, whose preferred install method is via docker, have moved to using a custom setup script (discourse-setup) which establishes a containers/app.yml file. There is also now a custom launcher script (launcher) which uses the app.yml file and applies the config expressed there to reconfigure and start the associated containers that constitute discourse.

The launcher script also has a number of other capabilities, ie start, stop, bootstrap, rebuild etc.

We already have existing 'special' treatment for the openvpn and owncloud Rock-ons but only for their install. It is proposed that we add discourse to this list of 'specials' and attempt also to accommodate the additional treatment that starting stopping and re configuring now requires, ie incorporate a wrapper for their launcher and use that as a proxy for our existing start, stop, re-config methods, ie add an additional mechanism akin to our existing custom install capability but for starting and stopping 'special' docker based systems such as discourse now is.

References:
discourse_docker on GitHub: https://github.com/discourse/discourse_docker
docker install guide: https://github.com/discourse/discourse/blob/master/docs/INSTALL-cloud.md

The existing Rock-on docker image
discourse/discourse : https://hub.docker.com/r/discourse/discourse
is now deprecated in favour of
discourse/base: https://hub.docker.com/r/discourse/base/

An associated issue in rockon-registry will link back to this issue as a dependency.

phillxnet:
Noting that we have the following out-dated special treatment for our now removed discourse rock-on:

src/rockstor/storageadmin/views/rockon_discourse.py

phillxnet:
I am currently working on removing the outdated and now redundant special treatment for the now removed discourse Rock-on.

phillxnet:
Closing as
Fixed by #2329

PR #2329, opened 2021-11-23
Remove special treatment for the deprecated discourse rock-on #2033

phillxnet:
If anyone wishes to restore our discourse rock-on and use this removed code as a template then we have it preserved in git so we can safely remove for now to avoid future confusion and avoid carrying dead code.

## Testing:
No additional test failures.
Source build (Leap 15.3 JeOS base) was able to establish the rock-on system then install-&-run/stop/start/stop/uninstall the netdata rock-on. Then stop and start the Rock-on service successfully there after.


phillxnet:
@FroggyFlox Although this may affect existing users of our discourse rock-on I'm not keen to keep otherwise dead code around. Especially give the almost complete lack of feedback on this particular issue. And that given we are moving to the next stable release we need to remove code that is not longer supported. I'm guessing that if we do have any long term users of our discourse rock-on,having installed it long ago when  the image was retrievable, that they have similarly done their own management of that container.

It's also a burden for us to carry too many 'special' cases and I think it would be good to remove all the cases of such special treatment that we can. Especially as we move to having to migrate all our code to Python 3.

phillxnet:
A source build post this change was also tested against our rock-on registry post discourse removal and no errors were observed.

See partner issue: https://github.com/rockstor/rockon-registry/pull/299

@FroggyFlox given this is essentially removing dead code I think we can safely merge.

Caveate: Existing users of our depricated/removed discourse rock-on should still have the ability to remove the rock-on as our fail through functionality should allow for this. Enabling those users to 'clean-up' and move to fully a self managed discourse arrangement outside of the Rockstor Web-UI. With the remnants of it's install also removed. No share content will have been affected by this, only the image removal.

phillxnet:
@FroggyFlox I'll go ahead an merge this to avoid it growing too old to merge. Testing looks to have been fairly comprehensive.

FroggyFlox:
Sorry for dropping the ball on this one and never getting back to you.

PR #2350, opened 2022-01-20
Update luks.py

zjzh:
refactoring code with chained comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

### DIFF
--- a/src/rockstor/system/luks.py
+++ b/src/rockstor/system/luks.py
@@ -201,7 +201,7 @@ def get_unlocked_luks_containers_uuids():
             "crypt",
         ]
     )
-    if len(out) > 0 and rc == 0:
+    if len(out) > 0 == rc:
         # The output has at least one line and our dmsetup executed OK.
         for each_line in out:
             if each_line == "":


phillxnet:
@zjzh Thanks for the suggestion / pull request. And in short I think I do prefer this actually; as was a little surprised to find this form of dual comparison. 

You submission as rather caught us in-transition. We are soon to create a testing branch as we have just finished our years long move to releasing our first v4 stable "Built on openSUSE" due mainly to the team size and number of bugs we uncovered and sorted in our move from our old CentOS OS base to the new openSUSE one.

As per: https://rockstor.com/docs/contribute_section.html and more specifically the Developer subsection of:
Could you please create an issue for his pull request. I know this is a pain but it's how we surface the all important attribution in the project. Whenever anyone is presented with a packaged update within our Web-UI we provide a changelog that links to GitHub issues and we then have our attribution route as folks can then see where the change was contributed from. We then have a "Fixes #issue-number" in the pull request text you've submitted here and we have an auto close link to the actual code change.

Sorry to have to ask more of you on what is a little change but attribution in such a project is important our our Web-UI expects to find an issue for each change we add.

Also in a few days I'm going to kick off the next testing channel and this time in a git branch so this change belong in that channel but we can cross that bridge when it exists.

Incidentally it looks like I'm to blame for this non Pythoneskness:
https://github.com/rockstor/rockstor-core/pull/1716
You should find this sort of comparison less common as time progresses and that was from 5 years ago no.

Thanks again for the submission and I'm all in favour of having code speak more clearly to a wider audience so do consider similar contributions of code clarity improvements as you find them. But we do need the issues to attach them to. And in this case for instance it need be no more than "Make luks.py more pythonic" or the like. Then add a fixes #issue-num to this pr.

Apologies for the timing our part here, but as stated we are about to start a testing channel and I'd like to keep all master contributions, regardless of complexity, strictly to serious bug fixes or the like. So if you could rebase this code on our testing channel when it arrives that would be great.

Great to have another set of eyes on the code by they way. I've not looked at that particular code segment for a number of years now actually and I'm surprised to find that style was of my own doing!! I've hopefully got a little more pythonic as I've staggered on. Keep an eye our for more of this sort and note that in our soon to emerge testing channel we are to commit the on-gong django migration to a Python 3 compatible version and then update our Python and associated build tools. So we have a lot of code mutations/churn to come shortly. See the following issue for this now ongoing work:
https://github.com/rockstor/rockstor-core/issues/1877

Also given my suggested issue title you may well fine other such issues in this code area that are similarly in need of some attention. But try and keep to a "no functional change" type arrangement and if we have unit testing in this area ensure we are still goon on their output. The above doc link indicates who to run these.

Sorry for the delay and further requests here. I'll try and get that testing channel setup now directly. As mentioned we've only just reached one of our years long goals so having a little break currently on the intense development before we dive in again with a new testing channel phase: but this time in it's own branch due to the large changes we have to carry forward to the next stable release.

phillxnet:
@zjzh Hello again. I'v now established our testing branch. It may well be that you have to re-submit this pr to rebase your change on this new testing branch. You are currently the first to be submitting to this new branch. We are in need of contributor doc changes to indicate this new arrangement as those here:
mention only master currently.

I see from the branch name that you likely submitted this change via GitHub Web-UI. You may find it interesting to do a full build from our source. Currently master and testing are identical so if you fancy doing a walk through of those docs and submitting suggestion that would be appreciated. They have been re-written relatively recently so more eyes on that also would be good.

I've created the following issue in docs to cover this doc requirement and new contributor information:
https://github.com/rockstor/rockstor-doc/issues/357

Thanks again for prompting this change via your most welcome contribution.

zjzh:
@phillxnet Thanks for your comments and I pull a request to the testing branch #2351

phillxnet:
Closing as we now have the 'against testing' as requested counterpart here:
"Make luks.py more pythonic" #2351

PR #2351, opened 2022-01-21
Make luks.py more pythonic

zjzh:
refactoring code with chained comparison which is more pythonic, concise, readable and efficient

### DIFF
--- a/src/rockstor/system/luks.py
+++ b/src/rockstor/system/luks.py
@@ -201,7 +201,7 @@ def get_unlocked_luks_containers_uuids():
             "crypt",
         ]
     )
-    if len(out) > 0 and rc == 0:
+    if len(out) > 0 == rc:
         # The output has at least one line and our dmsetup executed OK.
         for each_line in out:
             if each_line == "":


phillxnet:
@zjzh Thanks for opening a new pr against the testing channel as requested.
I've closed your prior master channel pr given this one now exists.

> and efficient
I'm not sure I buy the "more ...concise [or] efficient" side of this argument but I like all the others :) .

See: https://docs.python.org/2.7/reference/expressions.html#comparisons:

> Comparisons can be chained arbitrarily, e.g., x < y <= z is equivalent to x < y and y <= z, except that y is evaluated only once (but in both cases z is not evaluated at all when x < y is found to be false).

Where with a chained comparison, what we had before, there is only one evaluation needed of the common variable.
And the prior chained arrangement is more concise. But readability/clarity trumps all in non time critical stuff such as we have.
So in short: agreed. Lets keep our stuff as clear and open/explicit as we can.

Could you please open a matching issue to keep our attribution system happy, as requested in #2350 
Then I think we are set for this one.

Sorry for all the messing about but this way we get attribution within the Web-UI via our eventual changelog which is nice.

Did you find any other chained comparisons in this file that can similarly go under the same issue/pr set?

Thanks again for your efforts here.

phillxnet:
@zjzh I'm afraid I've miss read your changes. My apologies: you've caught me on kind of a break post our first v4 stable release. I thought we had a chained comparison before. I'm actually partial to not chaining them. Oh well. Lets go with it for now.
@FroggyFlox You input on this one and if we should adopt such things in the simpler multiple if comparisons.
I don't want this type of thing in our more complex stuff as I personally find it too mind bendy and I'm not convinced of it's Pythonic-ness given readability is very high on the list.

phillxnet:
@zjzh hello again.

I am sorry for the time wasted all around on your contribution. Initially by my own lack of the same in my rushed first look at your first pr against master due to our then pending developer 'tack' re the now opened testing channel. And in assuming you had found an anomalous chained comparison when in fact you were adding one: out of context. Our project, from memory, has hundreds of compound conditionals and I'm hoping they are all non chained (for readability / clarity) as unless performance is paramount, which it is not here, and in almost all other projects I'd hazard a guess, it's a retrograde step to add such a change. The Pythonic claim is a stretch and that's all we have left on this one. Plus there is the concern of code consistency. As stated we have hundreds of such comparisons and all hopefully are not chained so why choose this one which is one of our least critical performance wise as well. An issue opened first to propose a change is part of our attempt to provide good practice and why we have written our contributor guidelines docs. That way the maintainers can assess the rational/reasoning presented for such a change and not have to work backwards with every pr to see the purpose. Other folks code, or your own a few months later, is not always as obvious. But that pertains to more complex pull requests admittedly.

In short this doesn't help us or you in GitHub actually as your prior commit record of very many such code editor suggestions without any justification other than:

"... more pythonic, concise, readable and efficient" this issue.

"refactoring code with chained comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？" Update luks.py #2350

From your GitHub profile "Opened 88 commits in 93 repositories" entry here:
https://github.com/zjzh
we have:
Update test_imperative_auto_prune.py 

> "refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？"

refactoring code with For Else 

> "refactoring code with For Else which is more pythonic, concise and efficient; how do you think this change which has practical value？"

refactoring code with chained comparison which is more pythonic, concise, readable and efficient 

> "refactoring code with chained comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？"

refactoring code with For Else 

> "refactoring code with For Else which is more pythonic, concise and efficient than break; how do you think this change which has practical value？"

refactoring code with Dictionary Comprehension 

> "refactoring code with Dictionary Comprehension which is more pythonic, concise and efficient; how do you think this change which has practical value？"

refactoring code with List Comprehension

> "refactoring code with List Comprehension which is more pythonic, concise and efficient; how do you think this change which has practical value？"

refactoring code with Dictionary Comprehension 

> "refactoring code with Dictionary Comprehension which is more pythonic, concise and efficient;how do think this change which has practical value？"

refactoring code with Assign Multiple Targets 

> "refactoring code with Assign Multiple Targets which is more pythonic, concise, readable and efficient; how do think this change which has practical value？"

refactoring code with Assign Multiple Targets 

> refactoring code with Assign Multiple Targets which is more pythonic, concise, readable and efficient;how do think this change which has practical value？

And so on for he next few day up until yet another almost identical "code contribution" here in our own repository.
List above was an excerpt from:

projects-with-randome-editor-code-suggestions

Which is just a few days worth!

Some pr links for concrete context for my reasoning (all available from your GitHub profile, at least currently):

https://github.com/django-extensions/django-extensions/pull/1702
https://github.com/facebookincubator/OnlineSchemaChange/pull/33
https://github.com/yahoo/TensorFlowOnSpark/pull/582
https://github.com/getsentry/sentry/pull/31245
https://github.com/getsentry/sentry/pull/31248
https://github.com/sympy/sympy/pull/22891
https://github.com/sympy/sympy/pull/22896
https://github.com/PaddlePaddle/Paddle/pull/39089
https://github.com/PaddlePaddle/Paddle/pull/39119


So as per my proper reading of your commit re:

> I'm afraid I've miss read your changes. My apologies: you've caught me on kind of a break post our first v4 stable release. I thought we had a chained comparison before. I'm actually partial to not chaining them. Oh well.

And in the context that our hundreds of other compound comparisons are non chained, hopefully, and that this would only pollute the code base with what is essentially a code editor suggestion I'm going to have to politely turn down you contribution and save you time on creating the issue I requested for attribution purpose in you original pr a few days ago.

Also note that such repeated commits, now around 88 on at least as many different open source projects, without concern for code consistency, or any provided pr template from what I can tell, etc; could constitute a net time waste across very many well meaning projects. All modern open source developers have such tools available to them also. A random code editor generated refactor with the exact same commit message over dozens of commits/projects is not necessarily doing good; for you or those projects.

If you fancy proposing a genuine issue, or picking one of our several hundred open across our multiple open source repositories and gaining some programming experience or mentor-ship then fantastic. We have welcomed many first time GitHub contributors in several of our projects now but code editor suggestions out-of-context that reduce readability is not the way to go, especially so many times across so many projects distracting so many maintainers who almost exclusively have very little time. It's like we are being made the subject of an experiment of sorts!!!

Please observer some respect for the often very little available time of others in this arena when contributing in such a random fashion with machine generated code suggestions. We can all see these also in our editors. Your current most likely robot driven behaviour, with no concern for the project itself, is at least not welcome here. And once your current behaviour is more well known, in many other places also I suspect.

Closing as not helpful to either party.

My apologies if I have miss-read a good intention. Language barriers are a thing. But so it appropriate behaviour.

Issue #2254, opened 2021-01-10
Update Django within current Python 2 constrains


phillxnet:
With reference to "move to python 3" #1877 it is proposed that we begin our Django update 'train' from the current just prior LTS version (1.8.16) to the next proper LTS version of 1.8.19 and then in turn to the last Python 2 supported version of 1.11.29.

This issue it intended to cover this proposed update prior to our Python 3 move as newer versions of Django than 1.11.29 will have to wait until our base code is transitioned to Python 3.

In summary:
1.8.16 -> 1.8.19 so we are on LTS.
Then:
1.8.19 -> 1.11.29 to reach the latest available for Python2 via an LTS to LTS update.

A future issue/pr can then take us to Django 2 or 3 once we have #1877 sorted.

Due to the potentially deep and broad changes that may be required in this update it is further proposed that all work take place in parallel to our current master branch by way of a new testing branch. Once all our tests pass and all looks to be good the work can then be rebased on master and merged accordingly. This should help with avoiding breakage in what is currently our Release Candidate branch of master. In the future, given the scale of our technical debt, it may be that we should move to using the proposed testing branch to do all of the larger pull requests to come and release our testing rpms of the future from this testing branch. Leaving the possibility to have master as the Stable release branch. This arrangement is currently only a proposal however.

phillxnet:
I have begun looking into this issue. It remains to be seen if this is an appropriate issue for the allocated milestone.
Fingers crossed.

phillxnet:
It would appear that in moving to 1.11.29 we are also loosing compatibility with our current django-rest-framework version 3.1.1.

> jangorestframework-3.1.1-py2.7.egg/rest_framework/response.py", line 8, in <module>', '    from django.core.handlers.wsgi import STATUS_CODE_TEXT', 'ImportError: cannot import name STATUS_CODE_TEXT', '']

https://docs.djangoproject.com/en/1.11/releases/1.9/#miscellaneous

> django.http.responses.REASON_PHRASES and django.core.handlers.wsgi.STATUS_CODE_TEXT have been removed. Use Python’s stdlib instead: http.client.responses for Python 3 and httplib.responses for Python 2.

https://docs.python.org/2/library/httplib.html#httplib.responses

phillxnet:
## django-rest-framework
https://www.django-rest-framework.org/#requirements
states Python 3 and Django 2.2 + only for current versions!

Last version to support Python 2 was 3.9.3:
[...]

Issue #2362, opened 2022-03-10
WARNING -- EARLY TESTING CHANNEL -- KNOWN BROKEN (DEVELOPERS ONLY)


phillxnet:
Our Web-UI parses the rockstor package update changelog and displays those entries it understands, ignoring others.

This issue enables us to display, within the Web-UI, the above capitalised warning. And via the associated issue link, this text.

We are embarking on a massive set of code changes that will initially leave the **testing rpms broken** and so of **little use to non-developers**. 

We would like to inform non-developer testers of the following status:

## Known issues:
- [x] More recent oathlib version is Python 3 only (sorted by pinning in additional commit to #2365 ).
- [x] On submitting initial setup / admin user screen we have a variable type error, See Issue #2366 .
- [x] On reboot there is a catastrophic database migration failure that has yet to be resolved (#2376).
- [ ] 14 of 208 unit tests indicate multiple failures/errors likely associated with our in-progress Django update (see #2254). Issue #2371 
- [x] Mostly dysfunctional Dashboard (issue to follow). See /var/log/nginx/error.log for socket.io failed connections. Issue #2368
- [x] Inaccessible Replication pages; Send, Receive, likely pending Django url changes #2370.
- [x] Inaccessible Services page. Issue: #2370 
- [x] Inaccessible Task page, thanks @Hooverdan96. Issue: #2370
- [x] All System pages inaccessible intermittently (Services always fails). Post visiting Rock-ons then network some are then available!! Mostly related to #2368.
- [x] We don't appear to be starting all our apps: See:  /opt/rockstor-dev/var/log/supervisord_data-collector_stderr.log #2368 
- [x] The associated update of djangorestframework in #2365 has resulting in a generic requirement to move from request.DATA to request.data. See Issue #2379 
- [x] Django management commands are failing to run. Issue: #2377

More details to follow as we establish the new testing branch/phase and it's associated build system changes.

phillxnet:
Closing this meta-issue now as much water has passed under the theoretical bridge. We now have entirely working testing and have more than restored our past test coverage - sorting the last outstanding tick in this issue.

The Modernisation of source code issue linked here has received some attention also give we are not Poetry based. But that issue will likely be closed as unattainable given there is always ongoing progress in all our many upstreams.

Closing as fixed given we are now longer in early testing, but in Release Candidate phase.

PR #2365, opened 2022-04-08
Update Django within current Python 2 constrains #2254

phillxnet:
With reference to "move to python 3" GitHub #1877 it is proposed that we begin our Django update 'train' from our just prior to LTS version of (1.8.16) to the last Python 2 supported version of (1.11.29). These changes are intended to address the majority of updates required prior to our Python 2 -> 3 move as newer versions of Django than (1.11.29) will have to wait until our base code is transitioned to Python 3.

N.B. these changes are not sufficient for continued function across the board but are intended as a stop-off point to encourage further developer interest and help to promote co-operation in this address of our existing technical dept.

See: #2254
For development notes.

## Includes
- Remove defunct patterns module: deprecated (1.8) removed (1.10).
- Url updates.
-- Move to using plain list for urlpatters construct.
-- Update appliances url to list format.
-- Remove defunct patterns use on storageadmin/smart_manager urls.
-- Remove defunct leading empty string in urlpatterns lists.
-- Move remaining urlpattern list elements to url() type.
Django 2.0 introduces path() to remove/replace regex requirement, with re-path() as a direct drop-in replacement. But we are not there yet.
-- Normalise on url formatting, at least for now pre path() re_path().
Moved all root api page url definitions to top url file. If only one url is then left in any include, remove that include. This allows for common leading "/" prior to include without breaking the top api call, and reduces the number of files to view in order to manage api reformatting.
-- Remove redundant leading "/" from all url elements. This tends to numerous (urls.W002) warnings.
-- add now missing trailing '/' to include sub url lists.
These now replace the prior leading '/' within each include. But they were removed due to the following warnings:
""?: (urls.W002) Your URL pattern '^/(?P<did>\d+)$' has a regex beginning with a '/'."
"If this pattern is targeted in an include(), ensure the include() pattern has a trailing '/'."
- Update Django requires/version specification.
- Move db_router.py allow_migrate to post 1.10 format. Our prior syntax was deprecated in 1.8 and removed in 1.10.
- Upgrade django-rest-framework - 3.1.1 to 3.9.3 for Django compat. This is the last version of django-rest-framework to support Python 2.
- Upgrade django-oauth-toolkit and it's dependency 'Requests'.
- Requests in turn required a chardet update. Moved to latest and last to support Python 2.7.
- Update import for newer django-oauth-toolkit.
- Highlight/todo django-ztask to Huey transition points #2276. These have now been merged out by the subsequent Huey transition work.
- Django migrate --list deprecated (1.8), use showmigrations --list.
- Update to last release of djangorecipe from 1.9. Includes removal of now redundant/deprecated "projectegg" option.
- Move to render() from removed render_to_response() in home.py.
- Remove now defunct TEMPLATE_* settings - set TEMPLATES equivalent.
- Apply remaining non-hard-coded oauth2_provider migrations.
Our prior move to Django 1.8 incorporated all migrations to that date, including those for our prior django-oauth-toolkit == 0.9.0. This ended up causing a column/field conflict when applying the full migrations post updating to django-oauth-toolkit == 1.1.2. Resolve by fake applying those we already have pre-applied.
- Update paths in django-hack for updated eggs.
- Update django-hack template to account for djangorecipe updates.
Djangorecipe no longer contains the manage unit so we revert to using a mashup of path pre-definition and an upstream python 2 template for django command line management.
- Update test-settings.conf.in re Django update - TEMPLATE settings.
Includes some outstanding huey migrations that were missed in the django-ztask to huey move. Only affects the testing environment.
- Initial modification to serializers.py - djangorestframework update.
As from djangorestframework error we have:
"Creating a ModelSerializer without either the 'fields' attribute or the 'exclude' attribute has been deprecated since 3.3.0, and is now disallowed."
Add required fields param serializers.py - djangorestframework update.
- Fix Django 1.11.29 errors re insufficient Disk Obj save() calls.
From ""./bin/test -v 3 -p test_commands.py" we previously received:
"ValueError: <Disk: Disk object> instance isn't saved. Use bulk=False or save the object first."
in subtests of: test_bootstrap_command, and test_refresh_disk_state.
Both subtests related to _update_disk_state()
N.B. we may now over save in the earlier stages of _update_disk_state().
- Fix newer Django-rest-framwork warning re unordered object list.
test_get (storageadmin.tests.test_appliances.AppliancesTests):
"UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list:
<class 'storageadmin.models.appliance.Appliance'> QuerySet.
- Fix another disk instance isn't saved test failure.
test_compression (storageadmin.tests.test_pools.PoolTests):
ValueError: <Disk: Disk object> instance isn\'t saved. Use bulk=False or save the object first.\n']
In this case we already proceed to save each disk object in turn but this newer version of Django (1.11.29) responded to bulk=False option which in turn calls each associated objects save() method.
- Another Django-rest-framwork warning re unordered object list.
unordered object_list: <class 'storageadmin.models.update_subscription.
UpdateSubscription'> QuerySet.
paginator = self.django_paginator_class(queryset, page_size)
Both reported within supervisord_gunicorn_stderr.log
- Add default ordering to some more models. Addresses more test provoked warnings re:
"UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list"
- Restore prior default of APPEND_SLASH = False.
It is suspected that a newer Django began enabling this option if not found.
- Remove CommonMiddleware and APPEND_SLASH setting for now.
- Add TODO's to appliances.py re exception handling.
- Order pool and network device model outputs by default.
- Add development logging for ongoing concerns.
- Modify imports re Django parse_header and rest_framework media.
- Add follow=true and APIClient to test_appliances.py.
- In progress modifications to explore some ongoing test failures.
- Add auto generated migration file related to Django update.
It seems we have a new requirement in newer django to explicitly state our disk attached manager config.


FroggyFlox:
@phillxnet ... I was just looking at it... Could the following be a clue?
https://github.com/oauthlib/oauthlib/commit/b41d4801dec588d82e6633869ac541e207598297

We might need to pin oauthlib version as well... although I'm not sure which module is listing it as a dependency.

phillxnet:
@FroggyFlox Yes, this is what I was thinking. It's been a while since I've visited this code. I'll have a loot at additional pinning as it does very much look that way. Once sorted I'll push another commit to this branch as it's something that's changed since I was here last.

FroggyFlox:
Sounds good!
If I understand correctly, it seems that django-oauth_toolkit-1.1.2 has no "upper limit" on the `oauth` version so that might be why it imports the latest version it can:
https://github.com/jazzband/django-oauth-toolkit/blob/096ed111dc36d59c3d0281a249b596c180d220ec/setup.cfg#L30-L34

phillxnet:
From oauthlib changelog we have in version 3.1.1 (2021-05-31):
https://github.com/oauthlib/oauthlib/blob/master/CHANGELOG.rst#311-2021-05-31
734: python2 code removal

So: oauthlib 3.1.0 (2019-08-06) looks to be the last to support Python 2.7
https://github.com/oauthlib/oauthlib/blob/master/CHANGELOG.rst#310-2019-08-06
and Python 3.7 was added in 3.0.0:
https://github.com/oauthlib/oauthlib/blob/master/CHANGELOG.rst#300-2019-01-01

phillxnet:
We are now at:
```
Ran 208 tests in 15.300s
FAILED (failures=5, errors=13)
```
which is up from when this code was last visited: https://github.com/rockstor/rockstor-core/issues/2254#issuecomment-844348638
```
Ran 207 tests in 15.257s
FAILED (failures=4, errors=10)
```
However we  have had some test changes in the interim.

## New problem
We also appear to have another issue likely related to newer network code and it's interaction with the now updated Django:
Upon first Web-UI login we get a:

> Houston, we've had a problem.
> [u"'true' value must be either True or False."] 

```
[08/Apr/2022 19:02:42] DEBUG [storageadmin.views.network:82] The function _update_or_create_ctype has been called
[08/Apr/2022 19:02:42] ERROR [storageadmin.util:45] Exception: [u"'true' value must be either True or False."]
  File "/opt/rockstor-dev/src/rockstor/rest_framework_custom/generic_view.py", line 41, in _handle_exception
    yield
  File "/opt/rockstor-dev/src/rockstor/storageadmin/views/network.py", line 600, in post
    self._refresh_connections()
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/utils/decorators.py", line 185, in inner
    return func(*args, **kwargs)
  File "/opt/rockstor-dev/src/rockstor/storageadmin/views/network.py", line 184, in _refresh_connections
    cls._update_or_create_ctype(nco, ctype, ctype_d)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/utils/decorators.py", line 185, in inner
    return func(*args, **kwargs)
  File "/opt/rockstor-dev/src/rockstor/storageadmin/views/network.py", line 129, in _update_or_create_ctype
    connection=co, usercon=usercon, **config
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/query.py", line 394, in create
    obj.save(force_insert=True, using=self.db)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/base.py", line 808, in save
    force_update=force_update, update_fields=update_fields)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/base.py", line 838, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/base.py", line 924, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/base.py", line 963, in _do_insert
    using=using, raw=raw)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/query.py", line 1079, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/sql/compiler.py", line 1111, in execute_sql
    for sql, params in self.as_sql():
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/sql/compiler.py", line 1064, in as_sql
    for obj in self.query.objs
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/sql/compiler.py", line 1003, in prepare_value
    value = field.get_db_prep_save(value, connection=self.connection)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/fields/__init__.py", line 770, in get_db_prep_save
    prepared=False)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/fields/__init__.py", line 762, in get_db_prep_value
    value = self.get_prep_value(value)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/fields/__init__.py", line 1043, in get_prep_value
    return self.to_python(value)
  File "/opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/models/fields/__init__.py", line 1036, in to_python
    params={'value': value},
ValidationError: [u"'true' value must be either True or False."]
[08/Apr/2022 19:02:42] DEBUG [storageadmin.util:46] Current Rockstor version: Unknown Version
```
@FroggyFlox @Hooverdan96 This does look to be a differing treatment of this more recent network code by the Django update within this pr. Having a little look now. It may well account for the larger than expected test result differences. But I do see lots of object saves already which was a think I had to do through the code in this pr for many of the other tests to pass.
I'l like to get this sorted before merging this pull request as I remember login was a problem area and want to make sure this at least is working again.  Curios that the error seems to suggest a string value i.e. unicode "true" being saved to a boolean model/db field within _update_or_create_ctype().

FroggyFlox:
@phillxnet ... that's a bit I wrote for the implementation of rocknets. It's curious that it's a string, indeed, but that seems to be my fault.
We'll probably need to dig in on the output of the `docker` call there.

FroggyFlox:
the `icc` key is fetched by the following:
https://github.com/rockstor/rockstor-core/blob/eeb0084a2ea6f6bdb1cdbe35428aa5021ddadc72/src/rockstor/system/network.py#L234-L237

This `dtmap` dict is more or less the output of `docker network inspect bridge` in this case. In this get_con_config() call, it is (as obtained using the Pycharm python console:
```python
{u'Ingress': False, u'Name': u'bridge', u'Created': u'2022-04-08T15:00:13.833568712-04:00', u'EnableIPv6': False,
 u'Labels': {}, u'Driver': u'bridge', u'Attachable': False, u'ConfigOnly': False, u'Internal': False,
 u'ConfigFrom': {u'Network': u''},
 u'Options': {u'com.docker.network.bridge.name': u'docker0', u'com.docker.network.bridge.default_bridge': u'true',
              u'com.docker.network.bridge.enable_ip_masquerade': u'true', u'com.docker.network.driver.mtu': u'1500',
              u'com.docker.network.bridge.host_binding_ipv4': u'0.0.0.0',
              u'com.docker.network.bridge.enable_icc': u'true'},
 u'IPAM': {u'Config': [{u'Subnet': u'172.17.0.0/16', u'Gateway': u'172.17.0.1'}], u'Driver': u'default',
           u'Options': None}, u'Scope': u'local',
 u'Id': u'e4691c34b67c4a98332f1e484f6d36cdae00ba4e3ab48f8286a0004d4c6098ee', u'Containers': {}}
```

So it seems it's coming from there as it is returned as a string and not a boolean as the other True/False values outside of the "Options" key.

I see two options there to fix this:
1. transform this into boolean (hacky but we could do some sort of `if 'true': True else False`
2. change this model's field as a text input rather than BooleanField, but that might require some changes where that is used.

What do you think? I feel I'm creeping outside the scope of this PR, though, so I can create an issue if you'd like.

FroggyFlox:
Note that the `ip_masquerade` field should also be in the same situation if I'm correct:
https://github.com/rockstor/rockstor-core/blob/eeb0084a2ea6f6bdb1cdbe35428aa5021ddadc72/src/rockstor/system/network.py#L239-L244

phillxnet:
@FroggyFlox Agreed. I'm just having a look now. I suspect this is a 'thing' where we got invisible string 'true' to boolean True conversion that has since been tighten-up as I've not seen this before here. I've been creeping up on it as it goes.

Re:

> What do you think? I feel I'm creeping outside the scope of this PR, though, so I can create an issue if you'd like.

It may not be actually as it may be an artifact of the Django db population. Either way we can put it in a commit of it's own so if need be it could be cherry picked for stable if it comes up.

> * transform this into boolean (hacky but we could do some sort of `if 'true': True else False`

I much prefer this as we are already doing a tone of pre-processing in order to preset sanity and simplicity from the models to everything else. I'm almost there on this one I think. I'll post my findings shortly.

phillxnet:
OK, we have our culprit as you suspected:
```
[08/Apr/2022 20:35:54] DEBUG [system.network:238] dtmap['Options']={u'com.docker.network.bridge.name': u'docker0', u'com.docker.network.bridge.default_bridge': u'true', u'com.docker.network.bridge.enable_ip_masquerade': u'true', u'com.docker.network.driver.mtu': u'1500', u'com.docker.network.bridge.host_binding_ipv4': u'0.0.0.0', u'com.docker.network.bridge.enable_icc': u'true'}
[08/Apr/2022 20:35:54] DEBUG [storageadmin.views.network:176] 'ctype' was found in config
[08/Apr/2022 20:35:54] DEBUG [storageadmin.views.network:179] we have extracted a ctype_d={'aux_address': None, 'subnet': u'172.17.0.0/16', 'internal': False, 'host_binding': u'0.0.0.0', 'docker_name': 'docker0', 'icc': u'true', 'ip_masquerade': False, 'dgateway': u'172.17.0.1', 'ip_range': None}
[08/Apr/2022 20:35:54] DEBUG [storageadmin.views.network:82] The function _update_or_create_ctype has been called
[08/Apr/2022 20:35:54] DEBUG [storageadmin.views.network:83] co=NetworkConnection object, ctype=bridge, config={'aux_address': None, 'subnet': u'172.17.0.0/16', 'internal': False, 'host_binding': u'0.0.0.0', 'docker_name': 'docker0', 'icc': u'true', 'ip_masquerade': False, 'dgateway': u'172.17.0.1', 'ip_range': None}
[08/Apr/2022 20:35:54] ERROR [storageadmin.util:45] Exception: [u"'true' value must be either True or False."]
```

Note also that as you say we have the same issue with "ip_masquerade", but  we also have it with:

> u'com.docker.network.bridge.default_bridge': u'true'

FroggyFlox:
Thanks for the confirmation!

> u'com.docker.network.bridge.default_bridge': u'true'

I believe we do not use that "option" so it's not a problem (to the best of my knowledge).

phillxnet:
@FroggyFlox Funny you should say that. I was just looking for it and noticed it's lack of existence. We can always pop in a comment for future folk in case we end up parsing  default_bridge as well. I'll sleep on this and hopefully fix tomorrow. I may end up popping this in an issue/pr set of it's own as it may have further ramifications so would be easier if we had a dedicated issue to reference. But I'd first like to prove the fix allows basic function for this pr before merging this and splitting the network fix out to it's own pr built on the then newly updated testing.

FroggyFlox:
Sounds good!
Thanks a lot for the massive work there!

FroggyFlox:
> That is the network stuff here I think should go only to testing. It's too deep to go to master I think.

Agreed... especially given it doesn't seem to cause a problem in the Django version of the `master` branch anyway.

phillxnet:
@FroggyFlox & @Hooverdan96
Having now sorted the last known pinning requirement (broken build) via an additional commit, and spun out the network 'true' to Boolean issue in #2366, for which I have a fix lined up which goes on top of this pull request. I'm going to merge this into testing so we can more easily engage with the numerous failures that still exist down to the Django update and related library updates. I have update the meta issue for this new testing channel with some of the fields of failures still remaining:

WARNING -- EARLY TESTING CHANNEL -- KNOWN BROKEN (DEVELOPERS ONLY) #2362

As we go along we can link the specific GitHub issues and pull request fixes to that overall warning issue and keep that in play  until we once again reach a mostly functional system. The we can begin on the Python 3 move hopefully.

phillxnet:
N.B. there are currently unmet migration requirements added by this now merged in testing branch pr that will be addressed in the ongoing and previously referenced:
#2400 
Where an additional db migration is also required but his time concerning a refinement to the "poolbalance" model.

PR #2374, opened 2022-05-25
Prevent duplicate localdomain entry in generic map #2373

### LINKED ISSUE #2373: Unable to configure email alerts if the local network domain is localdomain
Because of a quirk of the rockstor Django application, if your local network has its default domain configured as `localdomain`, then it is not possible to configure email alerts. This is because Rockstor automatically adds an entry to the `/etc/default/generic` postfix lookup table with the domain name `localdomain` and then tries to add another entry for the default domain of the host. This leads to a duplicate entry being added in the lookup table with the `postmap` command failing with the following error:

```
CommandException: Error running a command. cmd = /usr/sbin/postmap /etc/postfix/generic. rc = 1. stdout = ['']. stderr = ['postmap: warning: lmdb:/etc/postfix/generic: duplicate entry: "@hostxxxx.localdomain"', 'postmap: fatal: error updating lmdb:/etc/postfix/generic: Invalid argument', '']
```

This can be fixed trivially by not adding the extra entry in this case.

parthjoshi-pc:
Prevent creation of a duplicate entry if the local network domain is localdomain in the /etc/postfix/generic lookup table when configuring email alerts

[Maintainer edit]
Fixes #2373

### DIFF
--- a/src/rockstor/storageadmin/views/email_client.py
+++ b/src/rockstor/storageadmin/views/email_client.py
@@ -164,7 +164,7 @@ def update_generic(sender, revert=False):
             fo.write("@{} {}\n".format(hostname, sender))
             fo.write("@{}.localdomain {}\n".format(hostname, sender))
             # add @<hostname>.<domain> if we can get a dnsdomain:
-            if dnsdomain != "":
+            if dnsdomain not in ("", "localdomain"):
                 fo.write("@{}.{} {}\n".format(hostname, dnsdomain, sender))
             # Add fall through entries for when the sending agent uses localhost.
             # This avoids some bounce scenarios when sender/from is root@localhost


parthjoshi-pc:
Fixes #2373

phillxnet:
@parthjoshi-pc Thanks for reporting and fixing this. Nice find, and fix.

We are currently in the throws of getting our next testing channel of the ground and into a semi working state. And have the following proposed doc change for contributions:

https://github.com/rockstor/rockstor-doc/issues/357

However this patch/fix looks exactly like what we were considering as "minor hot fixes" to our current stable release, with a view to cherry picking it into our existing currently dysfunctional testing branch there-after. See:
https://forum.rockstor.com/t/v4-built-on-opensuse-stable-channel-changelog/8212
and
https://github.com/rockstor/rockstor-core/issues/2362
for context.

@FroggyFlox & @Hooverdan96 assuming this passes review, do you concur?

Hooverdan96:
Makes sense to me, from what I understand about it.

phillxnet:
@parthjoshi-pc & @Hooverdan96 apologies for the delay on this one. As mentioned earlier I think this one looks like a key candidate for direct master merge; where our testing branch can pick it up via the next testing branch re-base on master once I've done a little juggling with the ongoing major testing changes etc.

phillxnet:
@parthjoshi-pc @Hooverdan96 @FroggyFlox 
Sorry folks but a slight change of plan here. I'd like to get this out, finally, and really our testing branch is the place for that: at least initially. Especially given the current shop "hold-up". And the delay here is indicative of our need to get the testing channel releases rolling again. So I'm going to change this over to apply to testing and merge shortly after I've merged my currently overly large testing update. This way we get it out to all folks who want to test it and due to it's specificity (unlike my mentioned testing pr I'm afraid) we can cherry pick later for master if that branch/release hasn't already moved over to a snapshot of testing. All a bit too in-flux currently so I'll just move this over to testing and take it from there.

Apologies for the mixed messages here, just trying to find the fastest way out of our current backlog and pending work.

phillxnet:
@parthjoshi-pc Thanks again for contributing this fix and again my apologies for the confusion and delay around it's being merged. It's one of the first merges after our branch split and things haven't worked out quite as neatly as I'd hoped, mainly due to me overstepping my own advise re focused pull requests as it goes. We should be back on a more productive footing once our currently major ongoing technical dept issues are well under-way.

This fix should now appear in our hopefully soon to be released new testing channel update run first. And likely later as a cherry pick back-port to our master branch based stable release if the master plan (pun) goes as intended.

@Hooverdan96 & @FroggyFlox My apologies on my slow handling here. Given this is going first to testing so I can in-turn re-base the current hold-up testing work I've popped it in as-is.

Hooverdan96:
@phillxnet, thanks for handling this. Obviously I am of no help here right now :)

Issue #2371, opened 2022-04-22
(t) Failing unit tests


phillxnet:
(t) = testing branch. Prior to our recent large Django version update we had all unite tests passing.
We now have:
```
Ran 208 tests in 19.686s

FAILED (failures=4, errors=10)
```

Noting the following failure/Error overview captured via:

> ./bin/test > test-results.txt 2>&1

>.E..............................F....................................E...E...............E.......E.E..E...................E....E............E...FFF......................................................................

## Error results:

- [x] ERROR: test_get (storageadmin.tests.test_appliances.AppliancesTests) -
- TypeError: assertEqual() got an unexpected keyword argument 'follow'
- [ ] ERROR: test_post_requests_2 (storageadmin.tests.test_pool_balance.PoolBalanceTests)
- IntegrityError: insert or update on table "storageadmin_poolbalance" violates foreign key constraint "storageadmin_poolbal_pool_id_f6813dfc_fk_storagead"
DETAIL:  Key (pool_id)=(2) is not present in table "storageadmin_pool".
- [ ] ERROR: test_post_requests_2 (storageadmin.tests.test_pool_scrub.PoolScrubTests)
- IntegrityError: insert or update on table "storageadmin_poolscrub" violates foreign key constraint "storageadmin_poolscrub_pool_id_9910cdd4_fk_storageadmin_pool_id"
DETAIL:  Key (pool_id)=(2) is not present in table "storageadmin_pool".
- [ ] ERROR: test_create_samba_share (storageadmin.tests.test_samba.SambaTests)
- IntegrityError: insert or update on table "storageadmin_sambashare" violates foreign key constraint "storageadmin_sambash_share_id_e755b23e_fk_storagead"
DETAIL:  Key (share_id)=(23) is not present in table "storageadmin_share".
- [ ] ERROR: test_post_requests_2 (storageadmin.tests.test_samba.SambaTests)
- IntegrityError: insert or update on table "storageadmin_sambashare" violates foreign key constraint "storageadmin_sambash_share_id_e755b23e_fk_storagead"
DETAIL:  Key (share_id)=(24) is not present in table "storageadmin_share".
- [ ] ERROR: test_post_requests_no_admin (storageadmin.tests.test_samba.SambaTests)
- IntegrityError: insert or update on table "storageadmin_sambashare" violates foreign key constraint "storageadmin_sambash_share_id_e755b23e_fk_storagead"
[...]

PR #2383, opened 2022-06-21
(t) Test fixtures non functional #2382

### LINKED ISSUE #2382: Test fixtures non functional
When our unit tests were last re-vamped (read updated to work again), post some major API and error reporting changes in the interim at that time, we have had a completely non functional fixture arrangement. See roughly 4 years ago:

"Standardise error text format and move to string.format part-1. Fixes #1913" #1914

Fixtures are the fake db content used to present tests with the desired database context in which the test is to be set. This short coming (know bug) was circumvented at the time, and since, by what may be considered excessive object mocking. This is less than desirable (read more 'artificial') than presenting tests with actual database dumps, or parts there of: a common origin of test fixtures.

Since our recent Django update via:
"Update Django within current Python 2 constrains #2254" #2365
our current 'broken fixtures' arrangement has now run it's course due to the increased sanity checks imposed on database / object interconnects; in turn increasing the complexity of mocking objects for our tests. See: (t) Failing unit tests #2371 for which this issue is a spin-off to address individually the failing fixtures issue now holding back our required test maintenance.

Thanks to @FroggyFlox for doing the vast majority of the 'solving' on this one.
Pull request to follow with fixture instantiation fix and small sample of adapted unit tests as proof of fix.

phillxnet:
Thanks to @FroggyFlox here for the fix to our long standing fixtures load failure. Separate commits are associated with this pr to help with attribution.

Fixes #2382 
See issue text for more context.

## Fixtures load fix
- Modify APITestMixin to
  - not skip `setUpClass()`
  - fetch User from fixture
- Do not inherit from APITestCase in `PoolScrubTests()` as APITestMixin already inherits from it.
- Create new fixture to included, (and later had removed), newly-added oauth2_provider migrations. This db element was later found not to be required.
- Keep fixture for test_pool_scrub.py to minimal models needed.
- Keep models for API auth in their own fixture

Includes:
- Removal of prior incompatible model mocks  in test_pool_scrub.py, so it can be used as a proof of,fix for the outstanding fixtures issue addressed above by @FroggyFlox .
- Minor test fix re inadvertent follow= parameter in test_appliances.py (see tick element in ongoing #2371)
- Notes on creating appropriate test_pool_scrub_balance and test_pools fixtures/db-dumps.
- Logging comments in pool_scrub.py _validate_pool() to ascertain current pool objects.
- Minor Black formatting fixes in  test_pool_scrub.py.

## Caveat
Note, this patch is intended to addresses fixture load failure in all tests. However only the one test, test_pool_scrub.py, has been modified accordingly as a proof of fix in this patches case. Further test development/modification to use the post-patch functional fixtures will be done in a/other dedicated issue/s.

## Testing
```
bin/test -v 2 -p "test_pool_scrub.py"
...
test_get (storageadmin.tests.test_pool_scrub.PoolScrubTests) ... ok
test_post_requests_1 (storageadmin.tests.test_pool_scrub.PoolScrubTests) ... ok
test_post_requests_2 (storageadmin.tests.test_pool_scrub.PoolScrubTests) ... ok

Ran 3 tests in 0.221s

OK
```


phillxnet:
@FroggyFlox Thanks for the double check. I've now just run this through another clean source build and test run and I think we are good so I'll merge to press on with addressing the fixtures updates across the entire test suite and likely test updates accordingly.

Issue #2384, opened 2022-06-22
(t) Refresh or remove as required all test fixtures


phillxnet:
(t) = testing branch. Prior to the work begun in '(t) Failing unit tests' #2371 as a result of major Django changes enacted in 'Update Django within current Python 2 constrains #2254' #2365, and following on from the fixture re-enablement spin-off of the former: '(t) Test fixtures non functional #2382' #2383 we have a test wide requirement to update or remove our now mostly outdated and newely re-enabled fixtures to address our current test situation of:

```
Ran 86 tests in 4.514s

FAILED (failures=3, errors=29)
```

in comparison to our current stable channel result of:
```
Ran 217 tests in 21.718s

OK
```

It is proposed that this issue be at minimum an address of the fixtures removal/update. Some test editing and mock object removal will inevitably be required especially where we don't just remove the fixtures that were not 'in-play' anyway in our last stable, and end up replacing the mocks and hand-created objects with the more robust fixture arrangement anyway. #2371 is evidence of a requirement to do this in some areas.

## Design goals
1. Fixtures to have an absolute minimum content.
2. Fixture reuse acceptable, thought discouraged as it can often challenging 1.
3. Always provide exact and proven command line instructions to reproduce any fixture used.
4. Always provide exact and proven command line instructions to run each test in isolation.

Where 3. & 4. are within the respective test files as comments towards the top of the file.

phillxnet:
@FroggyFlox Wow thanks; that's a nice break down. I'll get chipping away.

FroggyFlox:
> `AssertionError`: raw HTML returned as error message, test update likely.

These are all occurring in `test_task_scheduler.py` and very likely relate to #2370.
Removing the trailing slash in the URL of the `post` tests in this class fixes the errors:
```
buildvm:/opt/rockstor-dev # bin/test -p "test_task_scheduler.py"
Creating test database for alias 'default'...
Creating test database for alias 'smart_manager'...
System check identified no issues (0 silenced).
........
----------------------------------------------------------------------
Ran 8 tests in 0.983s

OK
Destroying test database for alias 'default'...
Destroying test database for alias 'smart_manager'...
```

phillxnet:
Re:

> These are all occurring in `test_task_scheduler.py` and very likely relate to #2370.
> Removing the trailing slash in the URL of the `post` tests in this class fixes the errors:

Pre:
```
[...]

PR #2387, opened 2022-06-28
(t) avoid recalculating pool free #2386

### LINKED ISSUE #2386: (t) avoid recalculating pool free
(t) = testing branch. While working on #2384 and revising mocking of such things as pool free space, we see a redundant re-calculation of the same in our PUT (remove) pool command when the exact same calculation has already been made via an earlier Pool object @property instantiation still held.

Pool "free" property definition:
https://github.com/rockstor/rockstor-core/blob/afa3e413a0202042fcabdc7c7e69b42d9042fe4f/src/rockstor/storageadmin/models/pool.py#L97-L104

storageadmin.views.pool.PoolDetailView -put()
https://github.com/rockstor/rockstor-core/blob/afa3e413a0202042fcabdc7c7e69b42d9042fe4f/src/rockstor/storageadmin/views/pool.py#L700-L705

This is obviously redundant. And given the second in-transaction re-use/repeat-call is the only other user of fs.btrfs.pool_usage() we can improve/ease our traceability and mocking by centralising this methods use.

phillxnet:
Remove a redundant re-calculation of pool.free in PUT (remove) pool command when the exact same calculation has already been made via an earlier and current Pool object @Property instantiation.

Fixes #2386

### DIFF
--- a/src/rockstor/storageadmin/views/pool.py
+++ b/src/rockstor/storageadmin/views/pool.py
@@ -26,7 +26,6 @@ from storageadmin.serializers import PoolInfoSerializer
 from storageadmin.models import Disk, Pool, Share, PoolBalance
 from fs.btrfs import (
     add_pool,
-    pool_usage,
     resize_pool_cmd,
     umount_root,
     btrfs_uuid,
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox I'm inclined to pop this one into testing branch given its trivial nature, and so that I can re-base the ongoing work related to #2384, which should simplify some mocking I need there. Just wanted to spin this one out given it is design related regarding our use of properties on the pool object. And the dual calls of pool_usage in the view (locally and via the Pool object) were confusing and redundant.

Issue #1611, opened 2017-01-10
Implement a disk replacement UI


phillxnet:
At times it is desired to replace an existing disk with another one, ie the function of:

    btrfs replace start devid /dev/sdX /mnt2/pool-name

A user level interface for this process would make for a nice improvement. We should ensure to allow for the -r switch where the source device is considered read only (recommended for when there are many read errors on that device). The status of this or other current replace operations could be shown in a tab on the same page, akin to the balance tab in pools maybe.
The current status can be had via:

    btrfs replace status /mnt2/pool-name

N.B. it is generally considered to be a longer process to use replace rather than:
"btrfs dev add" and then "btrfs dev delete", it might make sense to suggest this course of action in the same UI.

phillxnet:
Please also update the following forum thread with development in this area:
https://forum.rockstor.com/t/bad-drive-add-new-drive-problem/2798

phillxnet:
Linking to prior duplicate issue. Suggest we maintain this issue as prior issue has less info and unsure of "...much faster..." claim.
Linking to #593
https://github.com/rockstor/rockstor-core/issues/593

phillxnet:
Please also update the following forum thread with this issues resolution:
https://forum.rockstor.com/t/defective-disk-any-recommendations-regarding-replacement/3037

phillxnet:
Please also update the following forum thread with this issues resolution:
https://forum.rockstor.com/t/btrfs-replace-and-possibility-of-custom-dashboard-widgets/3775

phillxnet:
Linking to another more recent forum thread concerning the utility of a replace Web-UI element:
https://forum.rockstor.com/t/replacing-disks-in-a-raid10-pool/5130

phillxnet:
It is proposed that we extend the existing 'pool_resize_wizard' to include this feature.

While working on the related issue:
"Implement a delete missing disk in pool UI" #1700
the following UI entry point is to be added as a place holder:

![proposed-disk-replace-ui-entry-point](https://user-images.githubusercontent.com/2521585/46344376-d5121380-c638-11e8-9106-bc825777c434.png)

This should help with those looking for this feature within the UI and identify it, for the time being, as a 'planned option'.

Hooverdan96:
@phillxnet if I am not mistaken this can be closed now? Or do we have any outstanding UI work for this particular piece?

lxwulf:
When can we count approximately with the release of this function? Else great work! 😊

[...]

PR #2391, opened 2022-07-04
Abstract pool balance command construction #2390

### LINKED ISSUE #2390: Normalise on btrfs Huey task procedures having minimum content
(t) = testing channel. As with our existing low level fs.btrfs.start_resize_pool() it is proposed that we, likewise, minimise it's counterpart fs.btrfs.start_balance() Huey tasks (@task() decorator) to contain only logging and exception handling; abstracting the command construction to a more easily testable and mockable (as non threaded) discreet balance_pool_cmd() akin to our existing resize_pool_cmd(). Hence normalising also on discreet btrfs command generation.

During developments under #2384 a mocking limitation was observed due to a lack of btrfs command abstraction for post disk add balance invocations. Having the ability to mock for example a bash null of "true" for these commands let us preserve the existing code paths during testing down to and including the final Huey call, which also allows for testing of the Huey db object intervention: i.e having Huey 'fill in' end times for tasks pool tasks (pending issue already in TODO code comments).

phillxnet:
Reduces code run under Huey and enables easier testing of balance command construction procedure which will be needed as we expand btrfs-raid support to for example raid1c3 and raid1c4.

Fixes #2390 

Also aligns with format used in pool resize regarding abstracting the command generation and doing only logging and exception handling in btrfs Huey tasks.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -1786,9 +1786,7 @@ def start_resize_pool(cmd):
             )
         raise e
 
-
-@task()
-def start_balance(mnt_pt, force=False, convert=None):
+def balance_pool_cmd(mnt_pt, force=False, convert=None):
     cmd = ["btrfs", "balance", "start", mnt_pt]
     # With no filters we also get a warning that block some balances due to
     # expected long execution time, in this case "--full-balance" is required.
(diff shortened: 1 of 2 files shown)


phillxnet:
@FroggyFlox A little rough on the comments still but needed for some ongoing work in #2384 .
Merging as as spin off dependency for #2384 but useful as an indicator of how I think we should go as we move towards supporting such things as disk replace for example:
"Implement a disk replacement UI" #1611

PR #2420, opened 2022-11-08
Improve initrock re db setup/tune & systemd service setup #2076

### LINKED ISSUE #2076: initrock should drop rockstor specific databases rather than del dir
During a source build, or a source to rpm transition, initrock does the python equivalent of 'rm -rf /var/lib/pgsql/data' this has been known to cause postgresql to coredump. Restart was auto enacted but a more 'measured' approach of dropping our known databases only would be more appropriate and likely more predictable.

### LINKED ISSUE #2361: Improve initrock robustness re db delete
Under certain manual rockstor service modifications including by-hand delete of the /opt/rockstor directory we see a likely avoidable initrock script fragility:
```
Mar 10 17:14:17 rleap15-3 initrock[21233]: Traceback (most recent call last):
Mar 10 17:14:17 rleap15-3 initrock[21233]:   File "/opt/rockstor/bin/initrock", line 40, in <module>
Mar 10 17:14:17 rleap15-3 initrock[21233]:     sys.exit(scripts.initrock.main())
Mar 10 17:14:17 rleap15-3 initrock[21233]:   File "/opt/rockstor/src/rockstor/scripts/initrock.py", line 387, in main
Mar 10 17:14:17 rleap15-3 initrock[21233]:     shutil.rmtree("/var/lib/pgsql/data")
Mar 10 17:14:17 rleap15-3 initrock[21233]:   File "/usr/lib64/python2.7/shutil.py", line 270, in rmtree
Mar 10 17:14:17 rleap15-3 initrock[21233]:     rmtree(fullname, ignore_errors, onerror)
Mar 10 17:14:17 rleap15-3 initrock[21233]:   File "/usr/lib64/python2.7/shutil.py", line 275, in rmtree
Mar 10 17:14:17 rleap15-3 initrock[21233]:     onerror(os.remove, fullname, sys.exc_info())
Mar 10 17:14:17 rleap15-3 initrock[21233]:   File "/usr/lib64/python2.7/shutil.py", line 273, in rmtree
Mar 10 17:14:17 rleap15-3 initrock[21233]:     os.remove(fullname)
Mar 10 17:14:17 rleap15-3 initrock[21233]: OSError: [Errno 2] No such file or directory: '/var/lib/pgsql/data/pg_stat_tmp/global.tmp'
Mar 10 17:14:17 rleap15-3 systemd[1]: rockstor-pre.service: Main process exited, code=exited, status=1/FAILURE
Mar 10 17:14:17 rleap15-3 systemd[1]: rockstor-pre.service: Failed with result 'exit-code'.
Mar 10 17:14:17 rleap15-3 systemd[1]: Failed to start Tasks required prior to starting Rockstor.
```
With the suspicious line as follows:

> OSError: [Errno 2] No such file or directory: '/var/lib/pgsql/data/pg_stat_tmp/global.tmp'

We use shutil.rmtree("/var/lib/pgsql/data") which is wipe any prior database and the underlying os.remove fails us.

The core of the issue here is that we should talk to db itself to request it clears house, not go in and do this for ourselves at the file level. As such this issue is likely a duplicate of:

"initrock should drop rockstor specific databases rather than del dir" #2076

phillxnet:
Rationalises/clarifies existing processes, including:
- Request databases/role drops during set/reset, rather than prior brute-force database dir delete approach.
- Add Postgres tune step, performed on every initrock run. All tuning exposed in a single postgresql_tune.sql file.
- New more specific host based authentication (pg_hba.conf) approach: explicit users and database names. Template based on OS default postgresql13. Performed on every initrock run.
- Move towards greater use of explicit sql files during tune/setup. Eases user intervention.
- Add generalised rockstor-pre.service update capability.
- Added systemd service removal process, triggered by removal of template in conf directory.
- Removed now redundant hard-coded postgresql.conf, allowing more flexibility regarding tracking our upstream by making only the changes/tunes we require via "ALTER SYSTEM SET" and it's interaction with postgresql.auto.conf.
- Minor log and other variable refactoring.
- Black formatting.

Fixes #2076 
Fixes #2361


phillxnet:
During end-to-end testing of this pr, it turns out our new more constrained host bus authentication via pg_hba.conf was not allowing our Django user access to the "test_..." databases temporarily created when running our test. It was also observed that this same test setup expects a "postgres" database what was no longer part of our newer setup. This in turn caused a fail-through that then ended up in the test setup being unable to delete their own temporary databases (even after running all testts successfully):

> /opt/rockstor-dev/eggs/Django-1.11.29-py2.7.egg/django/db/backends/postgresql/base.py:267: RuntimeWarning: Normally Django will use a connection to the 'postgres' database to avoid running initialization queries against the production database when it's not needed (for example, when running tests). Django was unable to create a connection to the 'postgres' database and will use the default database instead.

> django.db.utils.OperationalError: cannot drop the currently open database


A commit is to follow that addresses these issues.

PR #2422, opened 2022-11-08
Update psutil to the latest version #2421

### LINKED ISSUE #2421: Update psutil to the latest version
We use psutils (https://github.com/giampaolo/psutil) predominantly for our dashboard as a way, from python, to query running processes and system utilization. As this is a Python library we have take into account our underlying (currently outdated) Python version. However this utilities current version looks to still be compatible with our current python and so it is proposed that we update sooner rather than later, and pin our new version to aid in our pending Python 2 to 3 move.

Current version 3.3.0, proposed version 5.9.4.

phillxnet:
It appears that the latest psutil version of 5.9.4 is still Python 2.7 compatible.
Upgrade our reference in setup.py and django-hack.py.

Fixes #2421

We use this library/utility to retrieve CPU % for
our Dashboard CPU widget.


phillxnet:
Builds on both architectures succeeded with all tests passing and with working CPU widgets in the Dashboard.
Merging ready for next release.

PR #2427, opened 2022-11-11
Embed and update more dependencies #2424

### LINKED ISSUE #2424: Move to embed and update more dependencies
During the development involved in moving us from our v3 (CentOS base) to our v4 "Built on openSUSE" we partly transitioned to using some OS system python libraries. It is now required that we transition back to including as many dependencies as is practical.

Three reasons for this proposed direction/requirement:
1. It is the industry standard for Python apps - however we are a little more than an app given our close integration with the OS, but still.
2. We are still dependant on some older legacy Python modules, at least until: #1877 is resolved. See also https://github.com/rockstor/rockstor-installer/issues/123.
3. We plan to carry our own Python interpreter in newer major releases so that we are no longer bound by our base OS and it's official Python version.

Current package (rpm) defined dependencies directly related to Python; in the context of the above referenced issues:
```
via:
rpm -qpR rockstor-4.5.0-0.x86_64.rpm | grep "python"
```
- [X] python --- Initially, pre #1877, we source our OS
- [x] python-dbus-python  --- May require additional repo in 15.4, see comments below re PyPi dbus-python v1.2.18.
- [X] python-xml --- available in 15.4 default repos.
- [x] python2-chardet --- Suspected redundant as already build from source and package included.
- [x] python2-distro -- Not available in 15.4, so we need to re-include for the latter issue in 2. above. Or remove our use as it was predominantly for our transition days between CentOS to openSUSE. But was also used to adapt behaviour between Leap versions and Tumbleweed.
- [X] python2-psycopg2 --- redundant as we now build our own from source for both architectures after 4.5.0-0 and pr #2407 & #2403
- [x] python2-pyzmq -- Built & embedded via egg for some time, but there are some questions regarding duplication that may need addressing, at least before #1877 is resolved.
- [x] python2-requests -- Likely redundant as package dependency. Recently updated via #2365, but we may still have elements of our code that resources the system instance.
- [x] python2-setuptools  -- Likely required during our package install - could be tricky as we require an old version as well.
- [x] python2-six -- Used in preperation for our Python  2 -> 3 transition - could be tricky in Leap 15.4, likely we need to add this into our package for now via building the egg. Or like python-dbus-python it may have to be resourced via an addition repo in 15.4.
- [X] python3-distro  -- Available in 15.4 also.
- [X] python3-python-dateutil  -- Available in 15.4 also.
```

phillxnet:
Update setup.py and our django-hack.py to accommodate embedding and updating:
- pytz to latest version.
- six to latest version.
- distro to latest version with prior expected behaviour.

The following were all moved to the last Python 2
compatible versions, prior to our move to Python 3:
- pyzmq
- requests
- dbus-python

Fixes #2424


phillxnet:
superseded by #2431

PR #2433, opened 2022-11-20
Address buildout replacement #2432

phillxnet:
Entirely untested and incomplete -- save point.

Created to encourage community input on development in this direction re:
Fixing #2432 

Progress to date:
- Most of the "Required" fields with some: nice to haves-
- Plus the beginnings of grouped dependencies.
- 
- [x] Remove all buildout, bootstrap, zc.sourcerelease etc dependencies and special-case. Our proposed Poetry build system should replace all the associated processes as any transition period where one is doing some things and another some other (and possible by accident more of the same) will be way more confusing that a clean start on Poetry.
- [x] Remove djangorecipe (https://pypi.org/project/djangorecipe/) last updated **Djangorecipe: easy install of Django with buildout** Last updated to 2.2.1 Released: Jun 29, 2016.
- [x] Hard-wire hard-code all paths associated with our current buildout bootstrap steps. this can be un-done at a later date but should make for a speedier transition (new path to be existing "/opt/rockstor" as per existing rpm installs)
- [x] Hard-wire hard-code all paths associated with our conf pre-processors within buildout.
- [x] Establish /opt/rockstor/etc/nginx/mime.types (from conf).
- [x] Establish /opt/rockstor/etc/nginx/nginx.conf  (from conf).
- [ ] ~~Create Poetry plugin to download our jslibs.tar.gz and unzip into /opt/rockstor/static/js directory such that there is then an /opt/rockstor/static/js/dev directory containing all our backbone, bootstrap etc js libs, see: https://github.com/rockstor/rockstor-jslibs. N.B. this need be no more than running a curl tar de-archive command via a Poetry hook or the like.~~
- [x] The above is problematic given poetry's maturity and on-top of this our need, currently, to use an older version to empower us in our move from Python 2->3. Instead we can, for now, achieve our goals by carrying a build.sh script to assist in preparing our source prior to a poetry build.
- [x] Instantiate directly our supervisord.conf for our rockstor.service "/opt/rockstor/bin/supervisord -c /opt/rockstor/etc/supervisord.conf"
- [x] Instantiate a hardwired /opt/rockstor/etc/init.d/gunicorn
- [x] Run collectstatic during our poetry build, possibly via plugin.
[...]

PR #2434, opened 2022-12-04
Replace Buildout with Poetry #2432

### LINKED ISSUE #2432: Replace Buildout with Poetry
Place-holder issue, for now, intended to express and address our eventual move to Poetry for our build system.
Our existing buildout http://www.buildout.org/en/latest/ based approach is cumbersome and based on a project receiving progressively less attention as the years move on. It was once the go-to method however.

Unfortunately we are now fairly heavily invested in this approach but also depend on now very old and failing plugins for our rpm preperation.

It is proposed that we move to the currently modern https://python-poetry.org/
However it is, as yet, unclear as to our options on this front regarding our current Python 2.7 base. However poetry version [1.1.15](https://pypi.org/project/poetry/1.1.15/) (August 2022) was python 2.7 compatible. Where-as our current oldest buildout plugin is from December 2012 !!

The hope is that we can transition from buildout to an older version of Poetry initially, and then leverage it's capabilities to assist us in our move to Python 3  (see #1877) Given our existing buildout plugin dependency is python 2 only which presents a particular challenge regarding moving our base language version and associated dependencies over.

phillxnet:
Our prior build system "Buildout" has now been superseded by more modern standards based approaches. Where we remove our buildbot artifacts and replace them with an initial Poetry config.

Fixes #2432 
Supersedes #2433 
@FroggyFlox & @Hooverdan96 Ready for review.

## Includes:
- pyproject.toml
- poetry.toml - to have poetry create/use .venv dir. Thanks to FroggyFlox for this gem.
- use poetry created .venv/bin in initrock.
- poetry.lock - via dependency constraints in pyproject.toml
- move nginx service management: supervisord -> systemd and configure via a systemd override file instantiated/updated via initrock. Also required/includes some Web-UI log-manager modifications.
- delete legacy manage.py, 64-btrfs.rules, smb.service, old db file, unused huey shell script file, and various bootstrap config files.
- delete now legacy setup.py file. Poetry can generate this for us but we aim to go the more modern root of being entirely pyproject.toml based.
- delete decade old selenium and bash tests. These tests have received zero attention in 10 years.
- remove our prior django-hack.py (modified manage.py) This file is now redundant as it was associated with our prior buildout mechanism. As we have now moved to using a Poetry build/venv system we can simply use the vanilla django-admin from within the custom venv.
- move nginx.conf, mime.types, gunicorn, & supervisord.conf to ./etc.
- hard-wire rockstor service files - install dir. N.B. hard-coded directory paths in above 2 items.
- update supervisord and gunicorn to latest Python 2.7 compat versions.
- Pre-create top-level var/log dir with placeholder file.
- remove long redundant docker conf file.
- pg_dump notes from old buildbout file to initrock.
- Use regular django-admin in initrock.
- activate default static files FileSystemFinder to pick-up our jslibs top directory and incorporate into the generic static dir.
- Simple script to prepare our source for a poetry build, includes the official poetry installer.
- update supervisord.conf re upstream advice re user and dummpy pass.
- Add note on moving gunicorn to systemd.
- Move remaining systemd setup from rockstor-bootstrap rockstor-pre (initrock). We already to the vast majority of our initial systemd service config in initrock.py (rockstor-pre.service), so move the remaining duplicate firewalld and unique atd systemd commands from the 'bootstrap' command in command.py of storageadmin invoked by our scripts/bootstrap which in turn is started by our rockstor-bootstrap; to our rockstor-pre initrock.py setup.
- modernise settings.py re dynamically establishing our run-time path and settings BASE_DIR accordingly.
- configure pyproject.toml to have poetry dist tar include self build capability.
- add comment on future build.py pre-build capability. Our current Python 2.7 tether prevents this capability.

Co-authored-by: FroggyFlox <30297881+FroggyFlox@users.noreply.github.com>


FroggyFlox (approved):
@phillxnet, Thanks a lot for such a massive PR!

I've tested on a Leap 15.3 VM used for previous builds (build dependencies already included, then):
```
mkdir -p /opt/rockstor
cd /opt/rockstor/
git clone https://github.com/rockstor/rockstor-core.git .
git fetch origin pull/2434/head:2432_Replace_Buildout_with_Poetry
git checkout 2432_Replace_Buildout_with_Poetry
chmod +x build.sh 
./build.sh 
systemctl start postgresql
poetry run initrock
systemctl start rockstor-bootstrap
```

All was successful and I could login to the webUI.

Similarly, all unit tests pass:
```
Ran 236 tests in 31.290s

OK
```

phillxnet:
@FroggyFlox Thanks for the review/double-check much appreciated.

I'll leave this open for a tad longer in case anything comes up in the interim, but otherwise I think we move to merge and do new issue/pull-requests for fixes/improvements as they crop-up. That way we at least have this newer build arrangement  in-situ ready.

Once merged I'll open, and hopefully address, doc issues to update our developer contribution doc section here:
with the new Poetry based approach to building.

phillxnet:
@FroggyFlox 

> build dependencies already included

Yes we have some work to do in that direction also. But again this can be address via more specific issue-pr sets later on.

FroggyFlox:
> But again this can be address via more specific issue-pr sets later on.

Agreed. This PR is already substantial enough as it is so it'll keep future adjustments a lot more clear and better documented.

Hooverdan96:
@phillxnet quick question on the hard-coding in the nginx and rockstor service files, does this not pose a problem for the development contribution context? I am probably missing something here, but if I remember correctly, a dev build can be placed in a differently named directory. Would that not break those hard-coded paths?

phillxnet:
@Hooverdan96 Thanks for taking a look at this.

> quick question on the hard-coding in the nginx and rockstor service files, does this not pose a problem for the development contribution context?

Short answer, yes. We no longer have package relocation as we have not, in this pr, directly replaced our buildout pre-processor. A Poetry plugin may do this for us in the future. But as-is we are forced to use the older 1.1.15 Poetry as it (luckily for us) still supports Python 2.7. That was a relief actually. But note we do use the Python3 only Poetry installer as it, again fortunately for us, can install the older poetry versions. Anyway the plugin system in poetry is relatively young and given we are, until we move whole-sale to Python 3, caught between a rock and a hard place. So I went with slimming down what we do and hard coding. At least until we can use upstream latest Poetry and hopefully at that point Poetry will also be available as an official openSUSE package. They have one for Tumbleweed already.

We also have a few anomalies regarding our systemd units and their naming (the new nginx override) and location, all the rest. But I'm about to open a couple of issues for those. But I don't thing there is time to deal with those just yet and still get our more recent code out for testing. We shall see.

> ... this is a rather cool approach to simplify or at least streamline the builds

Agreed. And we are now going, within our limits currently, all out (or should that be in) on the pyproject.toml standard. So yes, definitely a timely move but it has put rather a spanner in the works of the 4.5.1-0 release schedule. But still, this move was pending all along so best we are now making it, finally.

We will have to update the developer docs to state our hard-wired hard-coded paths for now. It's a pain but we can cross the relocation bridge when we come to it. We also now have hard-coded .venv entries too - they are typically themselves hard-coded, i.e. within the rpm build process that I'm currently working on we have:
```
cat /usr/src/packages/BUILD/rockstor-core-4.5.1-0/.venv/bin/bootstrap

#!/usr/src/packages/BUILD/rockstor-core-4.5.1-0/.venv/bin/python
from scripts.bootstrap import main

if __name__ == '__main__':
    main()
```
Our prior bootstrap approach to this was the equivalent of running a sed on all such wrappers (eggs as they were then). Super not keen on machine editing stuff after Poetry has made it for us 'special' as it were. So we are going with having the rpm install the .venv post install.

So that is another concern, but again, all in good time. Many packages are not in-fact relocatable, and we managed it before but it was at the cost of way too much magic and old magic at that. So I'm trying here to get us back to basics and rebuild up again using newer tools to enable our transition to Python 3. I believe we can also have Poetry, in time, do builds against both python 2 and 3. But that is getting a little ahead of ourselves, but only a little :). This will be a nice trick during our transition, there is no intention to continue the Python 2 base a day longer that we have to. And I for one will be super relieved to be into the future for once.

Again, thanks for taking a look at this. I'll create the two systemd issues I had in mind.

PR #2438, opened 2022-12-07
Rename our nginx drop-in override systemd file #2436

### LINKED ISSUE #2436: Rename our nginx drop-in override systemd file
It has come to light that our pending change, as part of #2434, involving a systemd drop-in file to override the default OS nginx config, is not named in compliance with the recommendations expressed here:

https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Shipping_unit_file_drop-ins

It would appear that our new override file named:
> rockstor-nginx-override.conf

to be placed, as recommended into a new dir named after the relevant service (nginx) of:
- nginx.service.d
-- within /etc/systemd/system/

by "def establish_rockstor_nginx_overide_conf()" in the modified initrock script. See #2432.

Should in fact be named starting with the number "30-" i.e.

> 30-rockstor-nginx-override.conf

However given our explicit naming, rather than the generic "override" we may be OK on this front, but our current naming does not address the opaque ordering issue (of additional drop-ins) that may result from us not abiding by this recommendation.

Depends upon the merging of currently pending pr #2434

phillxnet:
To comply with upstream guidance we add "30-" to our recently added rockstor-nginx-override.conf systemd override file.

We do not clean up our prior un-numbered variant as it was not included in any official released.
See: https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Shipping_unit_file_drop-ins

Includes black formatting: missed in recent commits.

Fixes #2436

### DIFF
--- a/conf/30-rockstor-nginx-override.conf
+++ b/conf/30-rockstor-nginx-override.conf
@@ -1,5 +1,5 @@
 [Unit]
-Description=The nginx HTTP and reverse proxy server - rockstor-nginx-override.conf
+Description=The nginx HTTP and reverse proxy server - 30-rockstor-nginx-override.conf
 After=network-online.target remote-fs.target nss-lookup.target
 Wants=network-online.target
 
--- a/src/rockstor/scripts/initrock.py
+++ b/src/rockstor/scripts/initrock.py
@@ -277,6 +277,7 @@ def establish_shellinaboxd_service():
         logger.info("- established shellinaboxd.service file")
         return True
 
+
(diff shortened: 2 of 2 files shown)


phillxnet:
@FroggyFlox and @Hooverdan96
I'll just pop this one in as it's only a file name change, and should help with starting out in compliance with the referenced upstream advice, at least where this file is concerned anyway.

As noted, we don't clear up the old, un-numbered file as it was not part of any prior release.

PR #2439, opened 2022-12-09
relocate our systemd service files #2437

### LINKED ISSUE #2437: relocate our systemd service files
Currently, and for the lifetime of the project's use of systemd (the entire lifetime of systemd) we have used the following directory for our rockstor* service files (not including overrides which are considered differently in this regard, see #2436 ):

```
rleap15-4:~ # ls -la /etc/systemd/system/rockstor*
-rw-r--r-- 1 root root 300 Nov 28 18:51 /etc/systemd/system/rockstor-bootstrap.service
-rw-r--r-- 1 root root 738 Nov 28 18:51 /etc/systemd/system/rockstor-ipv6check.service
-rw-r--r-- 1 root root 319 Nov 28 18:51 /etc/systemd/system/rockstor-pre.service
-rw-r--r-- 1 root root 314 Nov 28 18:51 /etc/systemd/system/rockstor.service
```

According to: https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Unit_files

> Systemd unit files, including but not limited to service, socket, timer and path files should always be installed in %_unitdir, i.e. /usr/lib/systemd/system, and never in /etc/systemd/system, so they can be overridden by users without conflicting with packaging.

We should instead be using:
```
/usr/lib/systemd/system
```
This modification is dependant upon the merging of pending pr #2434 

and should ease our ability to use official rpm macros to establish and manage our systemd configuration. Further enhancing our package robustness re systemd modifications and our compliance with upstream standards.

phillxnet:
Use initrock.py to move existing non override rockstor services to the preferred location of /usr/lib/systemd/system. There-by allowing for admin override of our own services. Our existing mechanism to update these same service files is then run directly there-after.

Fixes #2437 

Includes:
- modifying scripts/flash_optimize.py to comply.
- modifying system/osi to comply.
- added additional clean-up mechanism to remove known legacy service files. Currently only used for the short lived rockstor-ipv6check.service, which was used to remove a previous non upstream default of disabling ipv6.
- general code tidies including:
-- cleaner abstraction of systemd paths.
-- update string formatting mechanism in flash_optimize.py. -- removal of incompatible/ineffectual tmpfs setup for /tmp 
-- removal of automated install of hdparm by YUM, now legacy in this context.
-- Avoid noatime fstab editing of anything other than "/" as we now have a different upstream system drive arrangement.


FroggyFlox:
@phillxnet, it all looks good on my end after building from source, in case it helps. I did notice an oddity, though, regarding the `rockstor-ipv6check.service`. Indeed, it seems to still be somewhat around:
```
buildvm:~ # systemctl list-units --all rockstor*
  UNIT                       LOAD      ACTIVE   SUB     DESCRIPTION                              
  rockstor-bootstrap.service loaded    active   exited  Rockstor bootstrapping tasks             
● rockstor-ipv6check.service not-found inactive dead    rockstor-ipv6check.service               
  rockstor-pre.service       loaded    active   exited  Tasks required prior to starting Rockstor
  rockstor.service           loaded    active   running Rockstor startup script                  

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.

4 loaded units listed.
To show all installed unit files use 'systemctl list-unit-files'.
```

This was after multiple `systemctl daemon-reload` and a reboot, just in case. Do you have any idea where this is coming from?

phillxnet:
@FroggyFlox Thanks, yes that does help. I've seen similar here:
```
rleap15-4:~ # systemctl list-units --all rockstor*
  UNIT                       LOAD      ACTIVE   SUB  DESCRIPTION               
● rockstor-hdparm.service    not-found inactive dead rockstor-hdparm.service
● rockstor-ipv6check.service not-found inactive dead rockstor-ipv6check.service
```
But the same services don't show up in the list-unit-files:
```
rleap15-4:~ # systemctl list-unit-files --no-pager | grep rockstor
rockstor-bootstrap.service                   disabled        disabled
rockstor-pre.service                         disabled        disabled
rockstor.service                             disabled        disabled
```
Bit different here as I'm mid rpm development in the above instance, but still, it is strange. But I think mostly-harmless.

```
rleap15-4:/usr/lib/systemd/system # rpm -qf rockstor.service rockstor-pre.service rockstor-bootstrap.service
rockstor-4.5.1-0.x86_64
rockstor-4.5.1-0.x86_64
rockstor-4.5.1-0.x86_64
```
Best if we can find the clean-up command to remove these ghost instances and add to initrock and rpm %pre scriptlet.

But I think we will be OK to pop that in a follow-up pull request specifically addressing these ghosts. I had tried to stop and disable first on all rockstor* unique (non override) services but it got a bit messy as initrock is run from within one of those services so it ended up terminating itself half-way-through. So to avoid a reboot (or rockstor-pre service restart) on upgrade from pre 4.5.1 I ended up just moving then ensuring the newly located service files were updated.

So I'm favouring going with what we have here as it does look to be working as intended and feeds into my current work on the rpm preparation which has equally to fit in with our updated systemd location. And likewise we can add maybe the following to the %prep on a quick test here:

# With ghost legacy units:
```
rleap15-4:~ # systemctl list-units --all rockstor*
  UNIT                       LOAD      ACTIVE   SUB  DESCRIPTION               
● rockstor-hdparm.service    not-found inactive dead rockstor-hdparm.service
● rockstor-ipv6check.service not-found inactive dead rockstor-ipv6check.service
```
rleap15-4:/opt/rockstor # systemctl disable rockstor-hdparm.service
Removed /etc/systemd/system/basic.target.wants/rockstor-hdparm.service.
Removed /etc/systemd/system/suspend.target.wants/rockstor-hdparm.service.
## Proposed clean-up
```
rleap15-4:/opt/rockstor # systemctl disable rockstor-ipv6check.service
Removed /etc/systemd/system/multi-user.target.wants/rockstor-ipv6check.service.
```
There-after we have:
```
rleap15-4:/opt/rockstor # systemctl list-units --all rockstor*
  UNIT LOAD ACTIVE SUB DESCRIPTION
0 loaded units listed.
To show all installed unit files use 'systemctl list-unit-files'.
```

But I'm unsure how this plays with the move of rockstor-hdparm.service, which contains previous Web-UI entered settings. As just moving it didn't upset things.

Maybe I can add a simple disable on at least the legacy rockstor-ipv6-check.service unit removal bit.

phillxnet:
I'm preparing a little commit to add the ghost clear-up at least for the legacy units. It will help to leave at least the majority of installs without these service ghosts.

FroggyFlox:
Oh good find with regards to cleaning this up with a disable call. I agree with you that we should move with what you have in the PR here and adjust as/if necessary later on.

phillxnet:
@FroggyFlox I'm just about to pop-in a little commit doing a stop & disable before the delete on the ipv6 only.

phillxnet:
## Before service disable before delete patch:
We have a hand installed (old directory) ipv6 servcie:
```
rleap15-4:/opt # systemctl list-unit-files --no-pager | grep rockstor
rockstor-ipv6check.service                   enabled         disabled
```
A regular source install build.sh run was followed, along with post build developer suggestions:
```
2022-12-10 16:36:44,732: Atd enabled and started
2022-12-10 16:36:44,878: Normalising on shellinaboxd service file
2022-12-10 16:36:44,879: - established shellinaboxd.service file
2022-12-10 16:36:44,879: /etc/systemd/system/rockstor-ipv6check.service stop/disable/remove (LEGACY).
2022-12-10 16:36:45,305: Establishing Rockstor nginx service override file
2022-12-10 16:36:45,306: /etc/systemd/system/nginx.service.d/30-rockstor-nginx-override.conf updated.
2022-12-10 16:36:45,640: /usr/lib/systemd/system/rockstor-pre.service updated.
2022-12-10 16:36:46,020: /usr/lib/systemd/system/rockstor.service updated.
2022-12-10 16:36:46,411: /usr/lib/systemd/system/rockstor-bootstrap.service updated.
2022-12-10 16:36:46,927: Systemd config altered, running daemon-reload
```
, and we have:
```
rleap15-4:/opt # systemctl list-unit-files --no-pager | grep rockstor
rockstor-bootstrap.service                   enabled         disabled
rockstor-pre.service                         enabled         disabled
rockstor.service                             enabled         disabled
```
No ghost.

phillxnet:
The last test re ipv6 legacy service was repeated but with the service disabled. Just in case.
```
2022-12-10 16:44:43,296: - shellinaboxd.service already exists
2022-12-10 16:44:43,297: /etc/systemd/system/rockstor-ipv6check.service stop/disable/remove (LEGACY).
2022-12-10 16:44:43,696: Establishing Rockstor nginx service override file
2022-12-10 16:44:43,696: /etc/systemd/system/nginx.service.d/30-rockstor-nginx-override.conf up-to-date.
2022-12-10 16:44:43,696: /usr/lib/systemd/system/rockstor-pre.service up-to-date.
2022-12-10 16:44:43,696: /usr/lib/systemd/system/rockstor.service up-to-date.
2022-12-10 16:44:43,696: /usr/lib/systemd/system/rockstor-bootstrap.service up-to-date.
2022-12-10 16:44:43,697: Systemd config altered, running daemon-reload
```
And a repeat run, once there is no legacy service, of "poetry run initrock" to test idempotency:
```
2022-12-10 16:46:58,153: - shellinaboxd.service already exists
2022-12-10 16:46:58,153: Establishing Rockstor nginx service override file
2022-12-10 16:46:58,154: /etc/systemd/system/nginx.service.d/30-rockstor-nginx-override.conf up-to-date.
2022-12-10 16:46:58,154: /usr/lib/systemd/system/rockstor-pre.service up-to-date.
2022-12-10 16:46:58,154: /usr/lib/systemd/system/rockstor.service up-to-date.
2022-12-10 16:46:58,154: /usr/lib/systemd/system/rockstor-bootstrap.service up-to-date.
```
No changes this last time so no daemon-reload and we have no mention of the ipv6 legacy servcie.

phillxnet:
@FroggyFlox OK, so that looks to have addressed at least the ghost service creation of rockstor-ipv6check.service during our removal of it's unit file. The rockstor-hdparm.service is a little more nuanced as moving it's unit file doesn't seem to change it's state, plus I want to preserve the settings in that case. Hence only approaching the ghost issue for the legacy unit file.

FroggyFlox:
Nice! Thanks for taking care of that one.

phillxnet:
@FroggyFlox Thanks for the review and prompting the improvement. I'll pop this pull request in as it essentially does what it set out to do and we can always revisit.

Issue #2474, opened 2023-01-19
Spelling typo in password reset text


phillxnet:
In the following user facing text we have "occured" which should be "occurred":

https://github.com/rockstor/rockstor-core/blob/0249f5283203461d71093447611c762901f6f378/src/rockstor/scripts/pwreset.py#L50

This would make for a good 'practice pull request' for anyone interested in contributing to Rockstor.

phillxnet:
Closing as:
Fixed by #2479

Thanks to @slowhand93 for stepping up to this one. Much appreciated.

PR #2479, opened 2023-01-19
Typo in pwreset.py: occured -> occurred

slowhand93:
Next shot for: https://github.com/rockstor/rockstor-core/issues/2474
Thanks for your patience and help @phillxnet :)

### DIFF
--- a/src/rockstor/scripts/pwreset.py
+++ b/src/rockstor/scripts/pwreset.py
@@ -47,7 +47,7 @@ def change_password(username, password):
         users.smbpasswd(username, password)
     except:
         sys.exit(
-            "Low level error occured while changing password of user: %s" % username
+            "Low level error occurred while changing password of user: %s" % username
         )
 
 


phillxnet:
@slowhand93 Thanks for this. And my apologies for the current inflexibility of our test build system.

## Testing
With an rpm build with this patch pre-merged we have:
```
rleap15-4:~ # grep "Low" /opt/rockstor/src/rockstor/scripts/pwreset.py
            "Low level error occurred while changing password of user: %s" % username
rleap15-4:~ # zypper info rockstor | grep Version
Version        : 4.5.5-2479
```
I'm not entirely sure how to easily invoke this error as it's a fair way down a cascade of other errors that would also not have to be the case. But never-the-less the change was incorporated in that build as intended. So I thing we are good on this one.

Merging as this has address the indicated issue as intended. Much appreciated.

phillxnet:
We also have the intended attribution in the associated rpms changelog:
```
rleap15-4:~ # rpm -q --changelog rockstor | grep slowhand93 
-Spelling typo in password reset text #2474 @slowhand93
```
which is then parsed by the Software Updates page within the Rockstor Web-UI for info and attribution purposes.

Issue #2473, opened 2023-01-19
Don't load non-json rockon config files?


kanecko:
This happened on 4.5.4-0.

I have added `omada-controller.json` to the usual place in order to install it as a rockon via WebUI. Because I make a lot of amendments during testing, my text editor creates a backup file to the same dir but appends a tilde: `omada-controller.json~`

I have found that when clicking the Update button in WebUI, an arbitrary file is chosen from that dir. Sometimes with the tilde, sometimes without.

I suggest that you search for more specific rockon config files. E.g. that they end with, and only with ".json".
Of course, it is my duty to make sure only one, valid JSON rockon config file is in that dir, but I thought I put up the issue up for discussion anyway.

I'm confident that this can be repro'd by placing any two .json and json~ files in the dir.

![1674120763716 - Copy](https://user-images.githubusercontent.com/5829467/213411079-c8098862-3494-423c-850a-fc314fdf5ad8.jpg)

phillxnet:
@kanecko Thanks for creating this issue, much appreciated. And I totally agree.
Do you fancy preparing a pull request against this issue?
We have the following extensive wiki guide on the forum written by @FroggyFlox on how our Rock-on subsystem works:
https://forum.rockstor.com/t/rock-on-framework-implementation/6288

kanecko:
Maybe. My first priority is setting up my home network before my vacation time runs out :)

phillxnet:
Closing as:
Fixed by #2482

PR #2478, opened 2023-01-19
Spelling typo in password reset text

slowhand93:
Follow-up to [2477](https://github.com/rockstor/rockstor-core/pull/2477) on testing branch

PR #2480, opened 2023-01-20
Filter out any non-JSON files from the local repo

Stitch10925:
Filter out any non-JSON files from the local Rock-on repository.

PLEASE NOTE: This has not been tested!

https://github.com/rockstor/rockstor-core/issues/2473

### DIFF
--- a/src/rockstor/storageadmin/views/rockon.py
+++ b/src/rockstor/storageadmin/views/rockon.py
@@ -20,6 +20,7 @@ import json
 import logging
 import os
 import re
+import fnmatch
 
 import requests
 from django.db import transaction
@@ -494,7 +495,7 @@ class RockOnView(rfc.GenericView):
 
         local_root = settings.ROCKONS.get("local_metastore")
         if os.path.isdir(local_root):
-            for f in os.listdir(local_root):
+            for f in fnmatch.filter(os.listdir(local_root), '*.json'):
                 fp = "{}/{}".format(local_root, f)
                 msg = "Error while processing Rock-on profile at ({}).".format(fp)
                 with self._handle_exception(self.request, msg=msg):


phillxnet:
@Stitch10925 Thanks for stepping up to this one. Much appreciated.

Any chance you can test it, so that it could be "Ready for review". It will help as the review is then a double check. Meaning there is more confidence in it's merge. We are getting close to the next stable release so I'm eager to have stuff double checked where we can.

Thanks.

Stitch10925:
@phillxnet 
I remember seeing a wiki about how to load Rockstor locally, but I can't find it back. I don't have anything set up for testing at the moment. If you point me in the right direction I can test it.

phillxnet:
@Stitch10925 Thanks.
Re:

> If you point me in the right direction I can test it.

I think the wiki you might be thinking of is this one on the forum:
https://forum.rockstor.com/t/how-to-test-a-rock-on-pull-request/5224

So you could take a look at a proposed Rock-on pull-reqeust while you were at it. Don't use an existing (already merged) one such as that wiki entries example of netdata as it will clash now that it's been merged.

It's pretty straight-forward actually, and is basically just making a specific directory and placing a Rock-on json in there, and in your case with testing your proposal, an additional legal file with a duff ending such as may be left by some editors as a 'last-copy' or 'in-transit' older version.

We also have the [README.md](https://github.com/rockstor/rockon-registry#readme) in the https://github.com/rockstor/rockon-registry

Hope that helps, as it would be great to get this into the next Stable Release Candidate :).

And thanks for going the extra mile/1.6km on this one.

Stitch10925:
@phillxnet 

I have no experience with running Python-based software (I am a C# developer myself). Don't I have to run my local branch of Rockstor somehow so I can use the UI to see if the non-json files are not being shown? It's doing this part that eludes me, hence why I have not tested it.

phillxnet:
@Stitch10925 My mistake I was thinking you were asking about the specifics of testing this pr.

You can get a quick Rockstor setup in a VM from the new installers here: https://rockstor.com/dls.html
that have 4.5.5-0 pre-installed and then just make your python based change on-the-stop. A reboot or rockstor services based systemd restart as per:
https://rockstor.com/docs/contribute/contribute.html#code-build (outdated but we still have the same 3 main systemd services
and your new python change should be live.

No need for a full-on development env, but a basic rpm install can rebuild itself in.situ via the build.sh script these days anyway. However given your proposed change here is relatively trivial and has no db migrations or other such complexities a simply reboot post python change and the associated pyc file (the equivalent of compiled c) will be auto updated. Python is interpreted so no rebuild necessary, but if you don't restart the system or the associated rockstor service the running code is still form the old version of the pyc (python byte code)  file.

And my apologies again for the miss-interpretation here. We are awaiting the update of our:
doc entry now we have changed how we go about development. I.e. much simpler now as no buildout. We are all Poetry based now, but your change makes not demands on that front either so just change the python in your install and reboot (restart relevant systemd service) to test the results. In some cases a Web-UI page refresh (via browser) is required to force an update of the invoked code.

Hope that helps.

FroggyFlox:
@Stitch10925, @phillxnet,
I just tested this PR in a VM: works as intended.
- any file in the `local_metastore` that ends in `*.json` is correctly parsed.
- renaming that file to `*.json.backup` and the file is ignored.
- switching back to current *testing* branch and the `*.json.backup` file is not ignored anymore and parsed.

**Unit tests**
```
buildvm:/opt/rockstor # cd src/rockstor/ && poetry run django-admin test
(...)
Ran 236 tests in 29.317s

OK
```

Looks all good to me!

@Stitch10925, I think we may have some complication due to the fact that your patch was made against the *master* branch and not the *testing* branch. @phillxnet recently tried to manually change branch on a PR in a similar situation but it turned out not the be possible (see #2478). See @phillxnet's comment (https://github.com/rockstor/rockstor-core/pull/2478#issuecomment-1397054484) for a much better explanation that I could write.

My apologies for this inconvenience. This shows we really need to update our docs on how to contribute as soon as possible. I'll create an issue to do just that in our [rockstor-docs](https://github.com/rockstor/rockstor-doc) repo.

Stitch10925:
@phillxnet 
Thank you for the detailed description, makes sense. Now I know for next time, since @FroggyFlox has already tested it.

Stitch10925:
@FroggyFlox 
It's a really small change. If you want you can abandon this PR and I will repeat the change on the testing branch. That was carelessness from my part.

phillxnet:
Superseded by #2482

PR #2481, opened 2023-01-20
Add form validation for share selected multiple times #2064

### LINKED ISSUE #2064: [Rock-ons] Share mapping is ignored if the same share is chosen
At rock-on install wizard, if the same share is selected for two or more volumes, only the first one is saved and sent to the rock-on API for the rock-on install.

### Illustration
For instance, if the share `emby-conf` is selected for both volume mappings of the _Emby server_ rock-on (see Picture 1 below), only the mapping for _Media Storage_ is saved (see Picture 2 below).

_**Picture 1**_
image

_**Picture 2**_
image

### Consequence
Although docker still runs the container, Rockstor detects the install as _failed_, but no error is thrown to the user.

### Cause
I suspect this is simply due to the fact that the `share:volume` mapping is stored as a dict, with the share as key. As keys must be unique, the each new mapping value overwrites the previous one. The related code is: 
https://github.com/rockstor/rockstor-core/blob/3ad7eee27a40badab7edeea6dc198e800f6b121e/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js#L603-L607

### Resolution options
Although it may be recommended to choose a dedicated share for each mapping, this is not explicitly mentioned during the install wizard, but we may choose to enforce the choice of different shares for each volume mapping. This could be done with any combination of:
1. use form validation to check for duplicate share selected
2. remove the share from the list of possible choices if selected in another drop-down select field.

Some users may still prefer to use the same share for multiple mappings, however, so we might rewrite how this form information is parsed. This would require a substantial rewrite of both the front- and back-end of things, however.

FroggyFlox:
Fixes #2064 
@phillxnet, ready for review.

During the Rock-On install wizard, selecting the same share to be bound to multiple volumes of the same container is not yet supported by Rockstor. While adding such support is planned, we currently do not prevent the user for doing so, which would thus result in a broken configuration. This has unfortunately already hit several users who expected to be able to do select the same share multiple times.

This PR proposes to add a new form validation rule (`isDuplicate`) to detect such instances of a share being selected as the option for more than one volume and return a specific error message accordingly. The error message is as follows:
**This share is already selected for another volume. This isn't supported yet.**.

Note that this form validation is occurring **only** upon clicking the "Next" button.
See https://github.com/rockstor/rockstor-core/issues/2064#issuecomment-1397674233 for a demonstration (note that the error message displayed in https://github.com/rockstor/rockstor-core/issues/2064#issuecomment-1397674233 is slightly out of date).

This has been tested on Rockstor Testing branch, on Leap 15.3.
@phillxnet , note that this PR only touches JS code, I have not run any unit test for it.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
@@ -3,7 +3,7 @@
  * @licstart  The following is the entire license notice for the
  * JavaScript code in this page.
  *
- * Copyright (c) 2012-2013 RockStor, Inc. <http://rockstor.com>
+ * Copyright (c) 2012-2023 RockStor, Inc. <http://rockstor.com>
  * This file is part of RockStor.
  *
  * RockStor is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


phillxnet (approved):
An rpm built with this patch included has the additional validation as reported:

2481-rpm-build

validate-against-multi-use-share-selection-rock-ons
But independent share selection, as per mouse-over advise, is accepted:

valid-discreet-share-entry-ok

Emby-installed-independent-shares

Issue #2349, opened 2022-01-16
[Rock-ons] Rock-ons update fails if a non-JSON file is present in the local metastore


FroggyFlox:
As reported by forum user **pepsov**, updating the list of Rock-ons fails if the local metastore contains a non-JSON file.
https://forum.rockstor.com/t/how-is-the-best-way-to-edit-rock-ons/7135/10

Indeed, we assume all files in this directory are JSON and try to load them as such. The responsible part of the process might be below:
https://github.com/rockstor/rockstor-core/blob/eeb0084a2ea6f6bdb1cdbe35428aa5021ddadc72/src/rockstor/storageadmin/views/rockon.py#L495-L503

We should try to deal with this in a more elegant way. We could try skipping that file, for instance, and not block the entire Update process.

FroggyFlox:
Closing as fixed by #2482

PR #2482, opened 2023-01-22
Update rockon.py - only load files with a .JSON extension

### LINKED ISSUE #2473: Don't load non-json rockon config files?

Stitch10925:
Filter out any non-JSON files from the local repo

https://github.com/rockstor/rockstor-core/issues/2473
Fixes #2473

This PR was re-applied from a PR in the master branch:
https://github.com/rockstor/rockstor-core/pull/2480

@FroggyFlox 
This time on the testing branch!

### DIFF
--- a/src/rockstor/storageadmin/views/rockon.py
+++ b/src/rockstor/storageadmin/views/rockon.py
@@ -20,6 +20,7 @@ import json
 import logging
 import os
 import re
+import fnmatch
 
 import requests
 from django.db import transaction
@@ -494,7 +495,7 @@ class RockOnView(rfc.GenericView):
 
         local_root = settings.ROCKONS.get("local_metastore")
         if os.path.isdir(local_root):
-            for f in os.listdir(local_root):
+            for f in fnmatch.filter(os.listdir(local_root), '*.json'):
                 fp = "{}/{}".format(local_root, f)
                 msg = "Error while processing Rock-on profile at ({}).".format(fp)
                 with self._handle_exception(self.request, msg=msg):


FroggyFlox (approved):
As in #2480: 
I just tested this PR in a VM: works as intended.

- any file in the local_metastore that ends in *.json is correctly parsed.
- renaming that file to *.json.backup and the file is ignored.
- switching back to current testing branch and the *.json.backup file is not ignored anymore and parsed.

**Unit tests**
```
buildvm:/opt/rockstor # cd src/rockstor/&& poetry run django-admin test
Ran 236 tests in 14.090s

OK
```

Looks all good to me!

FroggyFlox:
@Stitch10925 , thank you for putting up with the branch change and my apologies for such a request for this change.

@Stitch10925, @phillxnet, this PR does exactly what it is intended to do so I think it's a no brainer. It does highlight a shortcoming in our approach though as we indeed focus more on the file extension that the actual format of the file. 
We should think of also expecting incorrectly formatted json files here, especially given this directory may be used by users to develop their Rock-on json file. Currently, we fail the entire "Update available Rock-ons" process just because one file was incorrect. I think it might be good to only fail that particular file and not everything in such situation. Of course, we would need to implement this once we have a better way to surface this sort of information in the UI (probably post #1877) otherwise it may seem like nothing happened. Just a quick thought...

@Stitch10925, thanks again!

FroggyFlox:
Note this issue would also fix #2349.

phillxnet:
@Stitch10925 Thanks from me also for this work, much appreciated.
@FroggyFlox Thanks for the thorough review, also much appreciated.

I'll go ahead and merge so this ends up in the pending 4.5.6-0 testing rpm builds.

My apologies for the delay here, I've had some background dealings to deal with unfortunately.

PR #2484, opened 2023-01-25
Upload config fail #2483

### LINKED ISSUE #2483: Upload rockstor config fails
Original forum thread: https://forum.rockstor.com/t/upload-rockstor-config-fails/8638

My current version: Leap 15.4 V4.5.5-0.
My previous version: Leap 15.3 V4.(the latest)

Initially I installed the Leap 15.3 V4.1.0-0 and configured the system a little bit (raid1, scheduled scrubbing, two Rockons, NTP server).
Then I subscribed to the testing channel and updated fully.

When the new RC2 iso came out, I have made a Rockstor backup and re-installed/overwritten the system disk with the RC2.
Installation went fine. I rebooted afterwards.

The problem now is that I cannot upload my backup!
[backup-2023-01-22-135322.json.gz](https://github.com/rockstor/rockstor-core/files/10487482/backup-2023-01-22-135322.json.gz)

The errors are a slightly different from mobile to PC browsers, but the best clue I've got was trying to upload from PC Chrome: 
image

When investigating deeper we see it fails at `POST https://192.168.0.103/api/config-backup/file-upload` with status code 400.
the "filename" seems actually defined here: `Content-Disposition: form-data; name=\"file\"; filename=\"backup-2023-01-22-135322.json.gz\"`

Frontend stacktrace: 
image

Full HAR: 
[rq.har.gz](https://github.com/rockstor/rockstor-core/files/10487539/rq.har.gz)

kanecko:
Codefix provided by FroggyFlox.
Tested by @kanecko on 15.4 V4.5.5-0.

[EDIT]
Fixes #2483 
See issue text/comments for details.

### DIFF
--- a/src/rockstor/storageadmin/views/config_backup.py
+++ b/src/rockstor/storageadmin/views/config_backup.py
@@ -626,7 +626,7 @@ class ConfigBackupDetailView(ConfigBackupMixin, rfc.GenericView):
 
 
 class ConfigBackupUpload(ConfigBackupMixin, rfc.GenericView):
-    parser_classes = (FileUploadParser, MultiPartParser)
+    parser_classes = [MultiPartParser]
 
     def get_queryset(self, *args, **kwargs):
         for cbo in ConfigBackup.objects.all():


FroggyFlox:
Thanks a lot, @kanecko, for taking the time to create this PR and of course for testing the proposed fix in #2483.

A bit of explanation here, for reference.

Although I cannot reproduce *exactly* the issue detailed in #2483, I can reproduce the HTTP 400 error (Bad request) upon uploading a config backup file. In my case, however, no response content at all is returned and nothing can be found in the logs (rockstor logs, or gunicorn logs). The only error I could find (in addition to the HTTP 400 response) was the following in `/var/log/nginx/error.log`:
```
2023/01/23 16:08:33 [error] 14064#14064: *313 readv() failed (104: Connection reset by peer) while reading upstream, client: 192.168.122.1, server: ~^(?<myhost>.+)$, request: "POST /api/config-backup/file-upload HTTP/1.1", upstream: "http://127.0.0.1:8000/api/config-backup/file-upload", host: "buildvm", referrer: "https://buildvm/"
```

I thus looked further into `gunicorn` but I couldn't find any indication of error in its logs (even after increasing its log level). Increasing request size limits in nginx, gunicorn, and even Django itself didn't make a difference.

Looking back into our Django side of things, we are dealing with this upload request in the `ConfigBackupUpload()` class:
https://github.com/rockstor/rockstor-core/blob/c99f9f885b1bcc6b664f6e6351dc31d682dd712d/src/rockstor/storageadmin/views/config_backup.py#L628-L629

We can here see we are using two parsers defined by Django Rest Framework: `FileUploadParser` and `MultiPartParser`. Their [docs](https://www.django-rest-framework.org/api-guide/parsers/#multipartparser) state:

> The `FileUploadParser` is for usage with native clients that can upload the file as a raw data request. For web-based uploads, or for native clients with multipart upload support, you should use the `MultiPartParser` instead.

This thus recommends the use of `MultiPartParser` in our case, **instead** of `FileUploadParser`. We currently use both, which seems to conflict here.

The question is why is this showing up as a problem now for us, however, as this has always been the case (and working thus far). Note that #2483 appeared in 4.5.4, which could thus be related to our Django version update, or Django Rest Framework update.

@phillxnet, as you can see above, the reason why the error appeared is still uncleared to me and maybe we should look into this a bit further to make sure we're not missing the real reason for #2483 here. It may not be worth wasting too much time on that, however, as given DRF's docs quoted above, using only `MultiPartParser` seems like the appropriate way to go in our case upon a brief look.

phillxnet (approved):
And rpm built and installed with this patch applied, successfully passed the testing outlined by @FroggyFlox in the following comment https://github.com/rockstor/rockstor-core/issues/2483#issuecomment-1402297470 under the linked issue of #2483:
select-config-backup-file

config-backup-file-received

```
[25/Jan/2023 16:01:45] INFO [storageadmin.tasks:64] Task [restore_rockons], id: 70715cb9-7087-455c-8dc2-3743f2aad96f completed OK
```

PR #2490, opened 2023-01-30
Move Rock-on url default definitions to runtime #2305

### LINKED ISSUE #2305: Move Rock-on url default definitions from built time to runtime
It is proposed that we move to declaring our Rock-on related 'repo' variables at run-time to more easily facilitate future enhancement re custom Rock-on repositories and enable easier manual adjustment in the mean time for this potential future configuration element without having to rebuild from source.

Please resource the later discussion in @korg's recent related pull request: specifically @FroggyFlox 's comment onward:
https://github.com/rockstor/rockstor-core/pull/2303#issuecomment-882971576

We currently define the canonical rock-on repositories (remove and local) as Django setting variables. This is now considered to be overkill (given the re-build requirement to effect a change). We could simplify this process and reduce our currently extensive reliance on the Django settings for such elements and keep things simple by just establishing constants (or as near as we can with Python) for these. It is further proposed that this move add an easily accessible edit to allow users to define their own rock-on repos to address such things as crippling internet latencies, or to avoid potentially unwanted out-of-lan calls by using a local lan mirror for example.

FroggyFlox:
Fixes #2305 
@phillxnet, ready for review.

We currently define the location of the remote and local Rock-Ons metastores in a constant placed in settings.py. As a result, any need to modify any of these settings would require the user to rebuild Rockstor.

This pull request (PR) moves the definition of this constant to rockon.py where it is used, enabling the user to only need a restart of the rockstor service after customization.

## Testing
1. The Rock-Ons service was configured and turned ON
2. Rock-Ons list was updated by pressing the Update button on the Rock-Ons page.
3. The `/opt/rockstor/rockons-metastore` dir (default local metastore path) was created and a custom Rock-On was placed inside it.
4. Rock-Ons list was updated by pressing the Update button on the Rock-Ons page. **The custom Rock-On appears.**
5. The custom Rock-On was removed from this dir.
6. Rock-Ons list was updated by pressing the Update button on the Rock-Ons page. The custom Rock-On is no longer listed.
7. A new local metastore was defined in the `ROCKONS` constant at `/opt/my_custom_metastore`.
8. The dir `/opt/my_custom_metastore` was created and a custom Rock-On was placed inside it.
9. Restart the Rockstor service: `systemctl restart rockstor`.
10. Rock-Ons list was updated by pressing the Update button on the Rock-Ons page. **The custom Rock-On appears**

## Unit testing
```
buildvm:/opt/rockstor # cd src/rockstor/ && poetry run django-admin test
Ran 236 tests in 12.831s

OK
```

### DIFF
--- a/src/rockstor/settings.py
+++ b/src/rockstor/settings.py
@@ -421,12 +421,6 @@ UPDATE_CHANNELS = {
 		},
 }
 
-ROCKONS = {
-	'remote_metastore': 'https://rockstor.com/rockons',
-	'remote_root': 'root.json',
-	'local_metastore': '{}/rockons-metastore'.format(BASE_DIR),
-}
-
 HUEY = SqliteHuey(filename='{}/rockstor-tasks-huey.db'.format(BASE_DIR))
 
 TASK_SCHEDULER = {
--- a/src/rockstor/storageadmin/views/rockon.py
+++ b/src/rockstor/storageadmin/views/rockon.py
@@ -49,6 +49,12 @@ from django.conf import settings
 
 logger = logging.getLogger(__name__)
 
+ROCKONS = {
+    "remote_metastore": "https://rockstor.com/rockons",
+    "remote_root": "root.json",
+    "local_metastore": "{}/rockons-metastore".format(settings.BASE_DIR),
+}
+
 
 class RockOnView(rfc.GenericView):
     serializer_class = RockOnSerializer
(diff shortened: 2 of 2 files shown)


phillxnet:
@FroggyFlox Thanks for seeing to this improvement. Much appreciated.

My apologies for the delay, as you know already, I was using this pr as a test case for some back-end improvements that are now sorted so I'll go ahead and merge this. Thanks for the test run-down. I've also tested redefining the "remote_metastore" to source from a local network nginx hosted Rock-ons repo such as we do prior to publishing new Rock-ons. A feature others may also appreciate now that you've made it easier to enable via this patch.

Issue #2394, opened 2022-07-11
Improve 'Maintenance required' advice for btrfs-single/btrfs-raid0


phillxnet:
In a btrfs-single degraded scenario (missing member) we still offer the generic advice of:

> **Maintenance required**
> Missing disk removal requires degraded,rw mount options. Pool is currently (unmounted).
> Consider degraded,ro to refresh backups first.
> Reload page to refresh active mount options. 

in our Pool details page.

However btrfs-raid single (like btrfs-raid0) has no redundancy capability. If a disk is missing (detached in Rockstor speak) the only recourse is to either reattach the missing device and scrub the pool, or wipe, recreate, and restore from backup. As per our doc section:
"Data Recovery: btrfs-raid0 Pools": https://rockstor.com/docs/data_loss.html#data-recovery-btrfs-raid0-pools
(N.B we are missing a sister btrfs-single doc section - issue to follow).

It is proposed that we contextually change the above **Maintenance required** advice to be wise to this brtrfs-single/btrfs-raid0 lack of redundancy. We might also link to the recently updated Data Recovery doc section.

phillxnet:
Referencing the issue and now merged (long ago) pull request that petained to:

> (N.B we are missing a sister btrfs-single doc section - issue to follow).

Issue: "add missing 'Data Recovery: btrfs-single Pools": https://github.com/rockstor/rockstor-doc/issues/393
Pull Request: "add missing 'Data Recovery: btrfs-single Pools'": https://github.com/rockstor/rockstor-doc/pull/394

phillxnet:
Update to "Maintenance required" section when raid0 (or any known raid level) is found to be in the newly recognised 'redundancy exceeded' state internally recognised by the code changes to be associated with this issue:


![raid0-missing-dev](https://user-images.githubusercontent.com/2521585/221376284-927a3496-4316-4a68-a5e0-f4fbdd76b6d5.png)

## Text contents in above:
Still in development:

> Disks  (1 Missing)
> Maintenance required
> See: [Data Loss-prevention and Recovery in Rockstor](https://rockstor.com/docs/data_loss.html)
> Btrfs-raid level (raid0) redundancy exceeded: too many missing devices (1).
> If possible reattach a sufficient number of the missing devices and refresh this page.
> This sections advice should change according to findings. A scrub must be performed as soon as rw access is established.
> If reattachment is NOT possible, this pool is beyond repair: delete, re-create, and restore from backup.
[...]

PR #2492, opened 2023-02-06
Pending 'missing' case issue re btrfs fi show #2311

### LINKED ISSUE #2311: Pending 'missing' case issue re btrfs fi show
There is a pending proposal to enhance the output of 'btrfs fi show' to include the devid and an advisory, or missing, device path. I.e. currently there is the following expected output of btrfs fi show:

```
...
        devid    1 size 5.00GiB used 1.26GiB path /dev/loop0
        *** Some devices missing
```
proposed to be replaced with:
```
...
        devid    1 size 5.00GiB used 1.26GiB path /dev/loop0
        devid    2 size 0 used 0 path /dev/loop1 MISSING
```

Where the device path "/dev/loop1" in this case, may be omitted if unavailable.

This additional missing dev info could be of great use to our 'missing' logic which is currently a little convoluted (for another issue) but more pressing is the capitalisation of the "missing" to "MISSING" as we currently have  an overarching flaging mechanism that simply searches for any line ending with "missing" in 'btrfs fi show':

See in fs/btrfs:
- is_pool_missing_dev(label)
- degraded_pools_found()

See also our test data for examples of captured 'btrfs fi show' output with pools where a device has one or more missing members:
- fs/tests/test_btrfs.py test_degraded_pools_found(self)

btrfs mailing list references:
https://lore.kernel.org/linux-btrfs/f31fb924-847a-8b54-3da6-707914135f05@suse.com/T/#md987208db4b27afe89e903d6873e2c5c36f3c167
https://lore.kernel.org/linux-btrfs/b983789e-3671-a5f3-1803-37b8992261b4@suse.com/T/#t

phillxnet:
Address by adjusting our existing mechanisms to first lower case the line prior to comparison. N.B. our prior pool (model) specific is_pool_missing_dev() required more modifications as the prior (older kernel) observed ordering/placement of the 'missing' line was no longer correct. So we adopt the same brute force approach used in our system-wide (header) tell degraded_pools_found().

An additional previously absent test for
is_pool_missing_dev() was also added with test data from older and newer kernels included. Latest kernel observed was Stable backport 6.1.10-lp154.3.g2a0570b-default on a Leap 15.4 base, with accompanying filesystem repo updates.

Light reformatting included.

Fixes #2311

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -203,8 +203,7 @@ def is_pool_missing_dev(label):
     Simple and fast wrapper around 'btrfs fi show --raw label' to return True /
     False depending on if a device is reported missing from the given pool by
     label. Works by matching the end of output lines for the string 'missing',
-    which is either, post remount reboot (first and last populated line) or
-    the last line (dev removed and no remount).
+    after lower the case of the line.
     Label is used as this is preserved in our Pool db so will work if the pool
     fails to mount, and there by allows surfacing this as a potential reason
     for the mount failure.
@@ -216,8 +215,11 @@ def is_pool_missing_dev(label):
     # --raw used to minimise pre-processing of irrelevant 'used' info (units).
     cmd = [BTRFS, "fi", "show", "--raw", label]
     o, e, rc = run_command(cmd)
-    if o[-3].endswith("missing") or o[0].endswith("missing"):
-        return True
+    for line in o:
+        if not line:
+            continue
+        if line.lower().endswith("missing"):
+            return True
     return False
 
 
@@ -240,7 +242,8 @@ def degraded_pools_found():
         if not in_pool and line[0:3] == "Lab":
             in_pool = True
             continue
-        if in_pool and line.endswith("missing"):
+        # Account for older and newer kernels respectively:
+        if in_pool and line.lower().endswith("missing"):
             # we are in pool details and have found a missing device
             degraded_pool_count += 1
             # use in_pool switch to avoid counting this pool twice if it has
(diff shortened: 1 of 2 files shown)


phillxnet:
## Testing
The newly updated Howto "Installing the Stable Kernel Backport": https://rockstor.com/docs/howtos/stable_kernel_backport.html
was followed to update the running kernel of a "Built on openSUSE" Leap 15.4 installer created rockstor 4.5.6-0 instance.

Degraded pools were then created and used to generate the required test data. Tests were extended accordingly (failing on the newer kernel data sets) and the tested code extended to fit. The Web-UI was then observed to be working as expected re the degraded pools, unmounted and mounted degraded via the custom mount option set within the Web-UI, with both the default and stable kernel backport.

```
uname -a
Linux buildvm 6.1.10-lp154.3.g2a0570b-default #1 SMP PREEMPT_DYNAMIC Mon Feb  6 07:27:26 UTC 2023 (2a0570b) x86_64 x86_64 x86_64 GNU/Linux
```

```
buildvm:/opt/rockstor/src/rockstor # poetry run django-admin test
.............................................................................................................................................................................................................................................
Ran 237 tests in 31.194s

OK
```

phillxnet:
@FroggyFlox I'm about to start working on another issue involving more extension to our testing in the fs/btrfs area. So I'll go ahead and merge this into testing channel to help avoid merging difficulties in the similarly located scrub issue which does look like it will need some similar text coverage extension.

FroggyFlox:
Thanks, @phillxnet!

phillxnet:
The test additions in this pr are buggy, however the data used (and skipped due to the missing .appends) was from test systems proven to work as expected.
This testing error is now being addressed in a pending pr under issue #2394 where the same test, and tested code, is being developed further for that issue.

PR #2157, opened 2020-04-29
BTRFS SCRUB status enhancements for rockstor

ubenmackin:
Updated the scrub status code to check the version of btrfs-progs that is installed. Starting with 5.1.2, the scrub status output changed. This code will handle both cases, either using the newer output format or the older.

Additionaly, if running with the newer btrfs progs, modified the Scrub Status Summary and Detail pages to show the ETA as well as the current scrub speed.

Fixes #2162
Fixes #1922

PR #2493, opened 2023-02-10
Refactor scrub status parsing to enhance/enable testing #2342

### LINKED ISSUE #2342: Support future kernel scrub output
Thanks to forum member dunkelfalke for highlighting this issue. If one runs a non default kernel e.g. a Stable backports kernel such as is referenced in the following pull request:
"Remarked Kernel_Head & Kernel_stable_Backport + filesystems repos ..."
See: https://github.com/rockstor/rockstor-installer/pull/88
The btrfs scrub status output is radically changed. This in turn throws our scrub reporting and ends up in a database error akin to:
```
DataError: integer out of range
```
As detailed in the following forum thread:
https://forum.rockstor.com/t/integer-out-of-range/8183

[EDIT]
From a mostly functional point of view this issue has been addressed: see the comment copied from below:

> Linking to potentially related work-done in the interim:
> 
> Scrub UI - Integer Out of Range https://github.com/rockstor/rockstor-core/issues/2397 https://github.com/rockstor/rockstor-core/pull/2398 @Hooverdan96
> 
> And to a prior adaptation on scrub format changes:
> 
> BTRFS SCRUB status enhancements for rockstor https://github.com/rockstor/rockstor-core/pull/2157 @ubenmackin @FroggyFlox

But is now awaiting the initially missed test coverage: See the following linked pull request below for this connection.
Refactor scrub status parsing to enhance/enable testing #2342 #2493

phillxnet:
Previous scrub status enhancements lacked test coverage. By refactoring our scrub status parsing mechanisms we can return almost full test coverage to this capability.

Replaces all-in-one scrub_status() with a wrapper that invokes:
1. scrub_status_raw() 'btrfs scrub status -R' parser.
2. scrub_status_extra() non '-R' variant of 1. Combining the results there-after.

## Includes
- Replacing pkg_resources (setuptools) 'parse_version' with custom btrfsprogs_legacy() function.
- Tests for btrfsprogs_legacy() and previously missing scrub status parsing extensions now refactored into (1.) and (2.) above.

Fixes #2342 
N.B. the stated issue is primarily functionally addressed by the referenced pull requests in that issue:
"BTRFS SCRUB status enhancements for rockstor #2157
However there remained, there-after, important test deficits indicated by @FroggyFlox 's review comment here:
https://github.com/rockstor/rockstor-core/pull/2157#pullrequestreview-404667948

This pull request addresses the following code review comments intended for follow-up work; such as this pr:

> ## Caveats
> 
> ... I did notice, however, a few points that could improved in further PRs. These do not break the current system, however, so I believe they shouldn't necessarily be blocking this current PR.
> 
> 1. in scrub_status(), we now have a total of three calls to run_command(). While these work well as it, having several calls to the same command within a given function will prevent us (I believe) from properly covering these with our units tests. I would thus suggest to keep in mind the possibility for separating them from the scrub_status() function. The first one, for instance, simply fetches the btrfs-progs version, which could thus be separated and thus easily re-used as necessary.
> 2. Following-up on the previous point, we should think of updating our relevant unit tests (test_btrfs_*) to cover the newly-parsed metrics.

And my own forum comments here: https://forum.rockstor.com/t/integer-out-of-range/8183/12

> ...that we don’t have test coverage of the most recent additions to that rather critical function. So I’ll leave the associated issue open where we can also add test data from the latest kernel.

regarding the same initial issue.

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -37,7 +37,6 @@ from huey.contrib.djhuey import task
 from django.conf import settings
 import logging
 from datetime import datetime
-from pkg_resources import parse_version
 
 """
 system level helper methods to interact with the btrfs filesystem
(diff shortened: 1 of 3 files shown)


phillxnet:
## Tests post proposed patch

```
 poetry run django-admin test
...
.....................................................................................................................................................................................................................................................
Ran 245 tests in 30.416s

OK
```

phillxnet:
A fresh rpm, built with this patch pre-applied, was installed on an ex Rockstor Installer instance with Stable Kernel Backport of:
```
uname -a
Linux buildvm 6.1.11-lp154.2.g16a4964-default #1 SMP PREEMPT_DYNAMIC Thu Feb  9 10:56:42 UTC 2023 (16a4964) x86_64 x86_64 x86_64 GNU/Linux
```
All Web-UI reporting of scrub looked to be as expected.

PR #2500, opened 2023-02-23
Remove raid0 dev restriction & tidy pool dev fencing code #2385

### LINKED ISSUE #2385: Revise draconian raid0 disk removal restriction
While working on #2384, which has inevitably involved overhauling some of our outdated tests, it came to light that we block all disk removal from a raid0 pool. This is likely down to an early development-based fencing that has inadvertently remained. We have had the same for btrfs-single but this has since been removed (although it too had some legacy test content).

It is proposed that we move our absolute block on disk removal for btrfs-raid0 pools to have the same minimum disk restriction already imposed on btrfs-raid1 pools: i.e. don't go below 2 disks. Size/usage constrains for disk removal are already in place for all raid levels.

As this measure affects critical path functions it is proposed for testing branch (update channel) only.

phillxnet:
We previously blocked raid0 disk removal outright. Move to allowing device removal, as per all other profiles, within profile limits.

Abstract btrfs-raid profile minimum device count & maximum device missing info and use the former through-out the fencing code. Adapt all related tests to the newly uniform error messages that result from hitting the remaining revised limitations.

These code changes are also intended to ease our adoption of raid1c3 & raid1c4 profiles. And form the initial awareness of these newer btrfs-raid profiles.

Fixes #2385 

## Testing
An existing subtest was modified from the older behaviour to what is proposed, and a previously missing sub-test was added for below minimum dev count on the single profile.

With the proposed changes we have the following:
```
cd /opt/rockstor/src/rockstor
poetry run django-admin test
...
.....................................................................................................................................................................................................................................................
Ran 245 tests in 32.836s

OK
```

### DIFF
--- a/src/rockstor/storageadmin/tests/test_pools.py
+++ b/src/rockstor/storageadmin/tests/test_pools.py
@@ -43,6 +43,9 @@ bin/django dumpdata storageadmin.pool storageadmin.disk storageadmin.share \
 --natural-foreign --indent 4 > \
 src/rockstor/storageadmin/fixtures/test_pools.json
 
+cd /opt/rockstor/src/rockstor
+export DJANGO_SETTINGS_MODULE=settings
+poetry run django-admin test -p test_pools.py -v 2
 ./bin/test -v 2 -p test_pools.py
 """
 
@@ -107,7 +110,6 @@ class PoolTests(APITestMixin):
         cls.patch_balance_status = patch("fs.btrfs.balance_status")
         cls.mock_balance_status = cls.patch_balance_status.start()
         cls.mock_balance_status.return_value = cls.default_balance_status
-
 
     @classmethod
     def tearDownClass(cls):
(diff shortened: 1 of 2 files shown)


phillxnet:
While working on #2394 I have need for a structure created in this pull request. As we are in testing still (although mid  RC phase) I think we are OK to merge this as-is. I do see a comment typo but I may end up moving the associated new structure in my current issue anyway, so can sort the typo in the process.

I'll leave merging for a little longer in case anyone is mid review.

FroggyFlox:
@phillxnet ,
Note that I'm unfortunately too short on time for the moment to review this PR so please do not wait for me if that was the case.

phillxnet:
@FroggyFlox Thanks, and no worries. Cheers for the heads up. I just didn't want to merge it if anyone was mid review.
I'll likely merge tomorrow now and take it from there as I need to build on this for the sighted issues fix.

PR #2512, opened 2023-03-06
alt arch causing false positive update available #2468

### LINKED ISSUE #2468: (t) alt arch causing false positive update available
Akin to the following commit:
https://github.com/rockstor/rockstor-core/pull/2462/commits/6d553a879de25c16c50538a9bdbc7290a84c8fee
associated with merged pr:
"move to multi-arch repos from leap 15.4 onwards #2458" #2462
we have a cosmetic bug arising from our move to mult-arch repositories:

Our existing dnf-yum changelog parsing magic ends up flagging an alternative architecture (say aarch64 on an x86_64 install) as a potential update.

I.e.
**Rockstor 4.5.4-0.aarch64, update is available!**
is displayed as an available update but as the changelogs are the same there is not changelog displayed.

This issue is primarily cosmetic as the "zypper up rockstor" command invoked as a result of the "Start Update" will not end up installing the alternative architecture rpm due to it's incompatibility.

It is proposed that we further enhance our dnf-yum changelog 'magic' to filter out this false positive rockstor package update indicator.

phillxnet:
As we are now multi arch we need to specify the arch of the package we are interested in as the dnf-yum changelog facility we use to retrieve this info seems to return changelogs for all architectures.
Fix by adding "*arch" where arch = host architecture to our "rockstor" package search term.

Add TODO to associated tests as we don't yet have this procedure covered.

Fixes #2468

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -321,7 +321,7 @@ def rockstor_pkg_update_check(subscription=None):
     machine_arch = platform.machine()
     if subscription is not None:
         switch_repo(subscription)
-    pkg = "rockstor"
+    pkg = "rockstor*{}".format(machine_arch)
     version, date = rpm_build_info(pkg)
     if date is None:
         # None date signifies no rpm installed so list all changelog entries.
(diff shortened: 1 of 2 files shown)


phillxnet:
Tested as detailed in the associated issue comments from here: https://github.com/rockstor/rockstor-core/issues/2468#issuecomment-1456436888

## Before proposed change:
before-proposed-change

## After 
after-proposed-change

PR #2519, opened 2023-03-21
improve behaviour on missing /dev/disk/by-id dir #2401

### LINKED ISSUE #2401: improve behaviour on missing /dev/disk/by-id directory
Thanks to forum member Omx for highlighting this failing. In some incompatible hardware configurations, there can be no by-id sub-directory created by udev: such as when no device has a serial number. Currently we fail with an inelegant:

> Houston, we've had a problem:
> "... "/usr/bin/ls: cannot access ‘/dev/disk/by-id’: No such file or directory" ..."

It would be a far better experience to catch this specific error and advise Web-UI users as to there likely being an incompatibility regarding a complete lack of device serial, with a reference to the original no by-id directory error just in case. We could also link to our existing Quick start: Minimum system requirements doc section:
https://rockstor.com/docs/installation/quickstart.html#minimum-system-requirements

LInking for context to the associated forum post:
https://forum.rockstor.com/t/error-running-a-command-cmd-usr-bin-ls-lr-dev-disk-by-id-rc-2-stdout-stderr-usr-bin-ls-cannot-access-dev-disk-by-id-no-such-file-or-directory/8510/2

phillxnet:
By catching the instance of no by-id directory we avoid an inadvertent 'general' error and enable our existing Web-UI info on missing device serial numbers with its embedded link to our 'Minimum system requirements' docs section. An info log is also added when this directory is nonexistent, again with a pointer to the same doc section.

Fixes #2401 

## Testing
See issue comments from https://github.com/rockstor/rockstor-core/issues/2401#issuecomment-1478321138

In brief pre patch we produce the issue stated error. Post patch we have our usual no-serial indicator but against all disks. We also have an uninterrupted entry into the main Web-UI after the initial setup screen. Where-as before we were blocked from this access by the rouge 'general' error message. But by-hand url manipulation could be used to by-pass this error.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -1871,6 +1871,11 @@ def get_byid_name_map():
     was encountered by run_command or no by-id type names were encountered.
     """
     byid_name_map = {}
+    if not os.path.isdir("/dev/disk/by-id"):
+        logger.info(
+            "-- /dev/disk/by-id missing. See 'Minimum system requirements' in docs. --"
+        )
+        return byid_name_map
     out, err, rc = run_command([LS, "-lr", "/dev/disk/by-id"], throw=True)
     if rc == 0:
         for each_line in out:
--- a/src/rockstor/system/tests/test_osi.py
+++ b/src/rockstor/system/tests/test_osi.py
@@ -43,6 +43,11 @@ class OSITests(unittest.TestCase):
         self.patch_os_path_isfile = patch('os.path.isfile')
         self.mock_os_path_isfile = self.patch_os_path_isfile.start()
 
+        # some procedures use os.path.isdir so setup mock
+        self.patch_os_path_isdir = patch('os.path.isdir')
+        self.mock_os_path_isdir = self.patch_os_path_isdir.start()
+        self.mock_os_path_isdir.return_value = True
+
         # root_disk() default mock - return /dev/sda for /dev/sda3 '/'
         self.patch_root_disk = patch('system.osi.root_disk')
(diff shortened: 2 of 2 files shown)


phillxnet:
## Unit tests
Unit test modifications to follow: addressing our outstanding host system sensitivity re the existence of the /dev/disk/by-id directory.

I.e. when running our existing test suite on a host system without the /dev/disk/by-id directory we have the following failures:
FAIL: test_get_byid_name_map (rockstor.system.tests.test_osi.OSITests)
FAIL: test_get_byid_name_map_prior_command_mock (rockstor.system.tests.test_osi.OSITests)

Pre-patch via exception of linked issue, post patch via intended return of "{}".

phillxnet:
# rpm-build test run

```
poetry run django-admin test
...........................................................................................................................................................................................................................................................
Ran 251 tests in 26.765s
OK
```
This 4.5.8-2519 was installed on a 4.5.8-0 Rockstor installer created instance and all dive serials were removed (via virtio divices) and a hard reset was performed(https://rockstor.com/docs/contribute/contribute.html#code-test). The resulting initial Web-UI setup behaved as expected, as did the Disks overview page: where all disks (system + 4 data disks) displayed our existing:

> Warning! Disk unusable as pool member - serial number is not legitimate or unique.
...

message.

A local test run was also executed:

> Ran 251 tests in 25.845s
> 
> OK

and this system was confirmed to not have a by-id directory:

> ls -la /dev/disk/by-id
> ls: cannot access '/dev/disk/by-id': No such file or directory

PR #2530, opened 2023-04-16
adds scrollbar on large env vars forms #2533

### LINKED ISSUE #2533: RockOn Installation: the environment form has no scrollbar when there are many vars defined
If a RockOn has a lot of environment variables defined, they will be added to the form, but the form itself will be simply cut off, when there is no more height available on the browser.
This means height becomes a rare commodity on smaller devices, and even on desktop monitors it becomes an issue after 10-14 variables.

kanecko:
I hope I did it right...

- feat(UI): adds scrollbar on large env vars forms:

Fixes #2533

In case of a RockOn with many environment vars, the form will be very long. To the point that it will continue outside the viewport. I have set its max-height to 80vh, which will keep the form nice and short. But in the case of larger forms, it will keep the form within the viewport. In the latter case, a vertical scrollbar is added to allow for a nicer user experience.

Animation

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/css/style.css
+++ b/src/rockstor/storageadmin/static/storageadmin/css/style.css
@@ -522,6 +522,11 @@ table.tablesorter thead tr .headerSortDown {
   padding: 0 20px 20px 20px;
 }
 
+#install-rockon-overlay > .overlay-content {
+  overflow-y: auto;
+  max-height: 80vh;
+}
+
 #pool-quick-details {
   background: #fff;
 }


phillxnet:
@kanecko Hello again.
Could you create an issue against which this pull request can be associated.
It's required as we use links within our Web-UI Changelog that point to each contributors issue.
This is also how we 'surface' attribution to our users which is important. Hence asking if you can do this so it's in your name. I can do if need be but nicer if created under you own GitHub user for consistency and clearer attribution.

Having a quick look at this now by the way.

phillxnet (approved):
Hello @kanecko and thanks for the fix. Nice.

I build an rpm with this fix included and did a before and after using one of your prior contributions: the Omada Controller Rock-on - as it had a shed load of ports to provoke this enhancements place as it were.

## Before (small monitor)
We have no scrollbar and so can't see the entire list:
before-pr2530-with-omada

## After (same small monitor)
We have the new fancy scroll bar:
after-pr2530-with-omada

I also tried playing with the 80vh setting but it looks to be just so actually.

So upon a partner issue being created we can merge this one ready for our next release.

Thanks again for a nice and effective UI enhancement.

FroggyFlox:
@kanecko , @phillxnet,
A quick question in case you already have that setup for testing:
Does this autoscroll also appear in the post-installation settings summary (clicking on the wrench icon)? I can't remember how the table there displays in case of many many rows...

phillxnet:
@kanecko Hello again.
Re:

> Does this autoscroll also appear in the post-installation settings summary (clicking on the wrench icon)?

Good point.

I've just checked and we have the summary screen with scroll-bar now:

after-pr2530-with-omada-summary

And as you mention, the spanner screen with scroll-bar now:
after-pr2530-with-omada-spanner

However I did notice that post what seemed like a successful install, that Omada Controller Rock-on now seems not to have a functioning Web-UI via the button: not related to this pr but just mentioning in case you fancy re-visiting. Likely due to either error on my part, or upstream image changes.

I'm pretty sure I put in the same as I did before when I reviewed that Rock-on here:
https://github.com/rockstor/rockon-registry/pull/341#pullrequestreview-1278976324
So if you do get the chance it would be good to know if I've just fumbled the install this time.

But for this pr: yes the scroll bars are there and my pics don't show it but they show-up better once in mouse-over (Firefox).

kanecko:
All comments handled.

phillxnet:
@kanecko Thanks for creating the issue as requested. Attribution is super important and this is our mechanism that helps with that.

Incidentally I think that the "Fixes issue-num-here" entry has to be on a line of it's own. I've edited the original text so that this is the case anyway.

phillxnet:
@kanecko Thanks again for this contribution. It will be included in what is planned to be our release candidate 6 (RC6) rpm.

Incidentally, I see that you used a branch named "testing" in your own repository. This is fine but can lead to inadvertent overlap when working with more than one pull request at a time. Which give your activity currently may end up happening :). If you setup issue specific branches, i.e. named after the issue your pending/proposed pull request is to address; this can help with separation of concerns. My apologies if this is already something you are aware of.

The following section in our Developer docs section has some more words on this approach:
https://rockstor.com/docs/contribute/contribute.html#making-changes
As you have indicated your are relatively new to git/GitHub it would be great if you could indicate if those words (and their surrounding counterparts) make sense to you :). Always room for improvement and more eyes always helps.

Cheers, and I look forward to merging yet more of your contributions as and when as it were :). Keep an eye on the current testing forum thread, where you have already featured of course (multiple times), for the 4.5.9-0 (RC6) release notes.

kanecko:
> My apologies if this is already something you are aware of.
Not really :)

> The following section in our Developer docs section has some more words on this approach:
I think I've read them when I did my very 1st PR, but not this time...

> As you have indicated your are relatively new to git/GitHub it would be great if you could indicate if those words (and their surrounding counterparts) make sense to you :)
Yes they do! :)

> Keep an eye on the current testing forum thread
I do

PR #2279, opened 2021-02-14
Retrieve remote groups information using SSSD InfoPipe (#2240)

FroggyFlox:
Fixes #2240
Fixes #1580 

@phillxnet, ready for review

As detailed in #2240, we currently fail to fetch and list groups information after joining and AD or LDAP server, until a machine or rockstor service restart. This has been tracked down to some issue with `grp` not being able to fetch the groups information of such remote users until after a `gunicorn` or rockstor service restart. This pull request (PR) thus proposes to implement a fail-over mechanism to fetch this groups information in such cases based on the D-Bus API of SSSD: InfoPipe.

### Overall description
The error described in #2240 occurs when trying to fetch the *groups* information of a remote user. This ultimately boils down the fact that we are using two different approaches to fetch *users* and *groups*. While users are fetched from the system database using `getent passwd`, the *groups* information is based on the python `grp` module. As the former sees remote users (in these conditions) but the latter does not, we end up in a situation where we ask `grp` to retrieve information on a group it doesn't know, thereby causing the error.
As a result, this PR proposes to implement a new way to fetch the *groups* information based on the D-Bus API built in SSSD: InfoPipe. Instead of completely replacing the existing way we fetch *groups*, this PR proposes to use the InfoPipe-based mechanisms as a fail-over if `grp` fails to fetch this information.

#### New packages needed
This PR requires **2** additional packages:
- `sssd-dbus`: the D-Bus responder of SSSD
- `python-dbus-python`: dbus python module

```shell
# zypper in --no-recommends python-dbus-python sssd-dbus
Loading repository data...
Reading installed packages...
'python-dbus-python' not found in package names. Trying capabilities.
Resolving package dependencies...

The following 3 NEW packages are going to be installed:
  python2-dbus-python python2-gobject sssd-dbus

3 new packages to install.
Overall download size: 729.6 KiB. Already cached: 0 B. After the operation, additional 2.4 MiB will be used.
Continue? [y/n/v/...? shows all options] (y):
Retrieving package python2-gobject-3.34.0-lp152.2.3.x86_64                                      (1/3), 396.2 KiB (  1.3 MiB unpacked)
Retrieving: python2-gobject-3.34.0-lp152.2.3.x86_64.rpm ..........................................................[done (80.8 KiB/s)]
Retrieving package sssd-dbus-1.16.1-lp152.15.2.x86_64                                           (2/3), 138.5 KiB (295.3 KiB unpacked)
[...]

PR #2532, opened 2023-04-17
Use InfoPipe as fallback mechanism to fetch user/group information #2526

### LINKED ISSUE #2526: [Active Directory] Failure to set an AD user as Samba admin_user until Rockstor service restart
Thanks to forum user **ocelot11** for reporting this issue: [https://forum.rockstor.com/t/active-directory-issues/8728/4](https://forum.rockstor.com/t/active-directory-issues/8728/4?u=flox)

# Description
As described in that forum post, when clicking on Submit during the creation of a Samba export using an AD user as the *admin_user*, Rockstor throws the following Exception:
```
File “/opt/rockstor/src/rockstor/rest_framework_custom/generic_view.py”, line 41, in _handle_exception
yield
File “/opt/rockstor/src/rockstor/storageadmin/views/samba.py”, line 239, in put
self._set_admin_users(admin_users, smbo)
File “/opt/rockstor/src/rockstor/storageadmin/views/samba.py”, line 141, in _set_admin_users
auo.smb_shares.add(smb_share)
UnboundLocalError: local variable ‘auo’ referenced before assignment
```
As detailed by @Hooverdan96 in that forum thread, the section throwing that exception is the following:
https://github.com/rockstor/rockstor-core/blob/e490c32c3308d0d18b468b902750aa9663784a5f/src/rockstor/storageadmin/views/samba.py#L118-L141

# Reproducer
1. Setup, configure, and turn ON the Active Directory service.
2. Go to Storage > Samba, and click on "Add Samba export".
3. Use an AD user as admin_user.
4. Click submit, and the error above is returned.

**Note** that if the Rockstor service is restarted or the machine rebooted, everything works as intended without error.
To reproduce this error again after a Rockstor service restart, one has to turn OFF the AD service, manually remove the AD user from the storageadmin.user table, manually remove the group from the storageadmin.group table, and then restart the Rockstor service. Without this "cleanup/reset" steps, the AD user remains "in cache" and thus no error will be returned upon creation of the Samba export as detailed in the reproducer steps above.

# Explanation
The `_set_admin_users()` method aims at adding the User <-> smb_share relationship in the database. Its logic can be summarized as follows:
1. Try to get the `storageadmin.user` object corresponding to the user set as admin_user in the Samba export form.
2. If no entry is found in `storageadmin.user`, fetch that info from system users using the `pwd` library.
3. Create corresponding `storageadmin.user` entry, save it, and then add the User <-> smb_share relationship.

In our case, the user is an AD user and thus does not have an entry in the `storageadmin.user` table. We thus try step 2 above. This one fails for a similar reason as investigated and solved in #2240. This is also why a Rockstor service restart works (as it did in #2240).

# Resolution
We may follow the same approach to fix this as we did to fix #2240: use InfoPipe as a failover mechanism if `pwd` cannot retrieve the required information. This will need further investigation to verify it is the best option, though.

FroggyFlox:
Fixes #2526 
@phillxnet, ready for review

We currently use the built-ins `pwd` or `grp` to fetch user/group information throughout the project. Unfortunately, this can sometimes fail in systems that have freshly been joined to an Active Directory domain. See #2526 for details on the conditions required for such failures.

# Proposed fix
To fix this issue, this pull request (PR) proposes to increase the robustness of users/groups fetching by adding InfoPipe as a fallback mechanism during the creation of a Samba export; we already use such fallback to retrieve the groupname information for a domain user (see #2279).
Briefly, we add this InfoPipe-based fallback mechanism in two places and to retrieve:

  - user information when setting Samba admin users
  - group name for a given user when listing system users

To simplify and harmonize our user retrieval mechanisms, switch to using the already established `combined_users()` method during `_set_admin_users()` to retrieve all users seen by `getent passwd`, complemented by InfoPipe as fallback. See https://github.com/rockstor/rockstor-core/issues/2526#issuecomment-1510424000 for details.

This PR also fixes a related bug in retrieving groupname from an AD user: see https://github.com/rockstor/rockstor-core/issues/2526#issuecomment-1510492500 for details.

It also includes the following minor improvements:

  - a refactoring of the now-replaced `ifp_get_groupname()` function to make it more flexible: able to retrieve any given property or several properties of either a user or a group
  - simplify our current check for the active status of SSSD with a simple boolean wrapper using init_service_op to return True when the given systemd service is active.
  - small docstring formatting according to PEP257 and Sphinx's Python signatures (https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-signatures)
  - return ifp-fetched groupname in User model property only if SSSD is active
  - Black formatting and update copyright

# Demonstration
1. Configure and turn ON the AD service
2. Go to Storage > Samba and create a new Samba export using an AD user as Samba admin user
3. The Samba share is successfully created

# Testing
No new unit test was added.
```
(...)
Ran 253 tests in 13.295s

OK
```

# Final note
I noticed the following behavior. When we create a Samba export, the user selected as Samba admin user (if any) is added to the Django User table along with its link to the SambaShare table. We thus now have an AD user in the User table. When this Samba export is deleted, however, we do NOT clean this up. This means that if the machine leaves the AD domain, we will still have that AD user listed as user. This is beyond the scope of this PR, but we may need to think about that as cleaning this up may prove tricky.

### DIFF
--- a/src/rockstor/storageadmin/models/user.py
+++ b/src/rockstor/storageadmin/models/user.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2021 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <http://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
@@ -25,7 +25,7 @@ from django.core.validators import validate_email
 from django.db import models
 
 from storageadmin.models import Group
-from system.users import ifp_get_groupname
+from system.users import ifp_get_properties_from_name_or_id
 
 
 class User(models.Model):
@@ -46,6 +46,13 @@ class User(models.Model):
 
     @property
     def groupname(self, *args, **kwargs):
+        """Get user's groupname
+
+        Get the user's groupname from the following sources, in order:
+          - linked Group model entry
+          - from the system, using `grp`
+          - from InfoPipe: useful for domain users
+        """
         if self.group is not None:
             return self.group.groupname
         if self.gid is not None:
@@ -54,9 +61,13 @@ class User(models.Model):
                 charset = chardet.detect(groupname)
                 groupname = groupname.decode(charset["encoding"])
                 return groupname
-            except Exception:
-                # Failed to fetch user using grp, so let's try with infofipe
-                return ifp_get_groupname(self.gid)
+            except KeyError:
+                # Failed to fetch user using grp, so let's try with InfoPipe
+                ifp_res = ifp_get_properties_from_name_or_id(
+                    "ifp_groups", int(self.gid), "name"
+                )
+                if ifp_res is not None:
+                    return str(ifp_res["name"])
         return None
 
     @property
--- a/src/rockstor/storageadmin/views/samba.py
+++ b/src/rockstor/storageadmin/views/samba.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2020 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <http://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
@@ -17,7 +17,6 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
 import logging
-import pwd
 
 from django.conf import settings
 from django.db import transaction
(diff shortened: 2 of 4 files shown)


Hooverdan96:
@FroggyFlox, excellent!
On your Final Note, would it make sense to open an issue for this to keep track of it?
Also, I assume, there could be two options:
 - manual cleanup, i.e. the SambaShare needs to be edited and the user has to be replaced with an existing one (if that is the solution) - this could be addressed via paragraph in the documentation as a `N.B.` or something like that.
 - systemic cleanup - that, like you said, could be much trickier and thus also much further down the road.

phillxnet (approved):
@FroggyFlox This is a most welcome improvement, refactor.
I don't currently have the setup to properly test this which makes this a non-review I'm afraid, but we do have a currently active forum member that you have been engaging with so I'm merging this as-is so that they can test in their setup. Plus we have your own testing. You explanations on the what and why are much appreciated and will serve us well as we revisit this code down the line.

Thanks again for the code improvements and much needed revamp. And for the improved comments. Hopefully in time we can further rationalise these fail-through mechanisms and just use what works all of the time: but this will likely be for when we have more of our libraries etc updated.

PR #2547, opened 2023-05-01
use cryptsetup to retrieve container uuid #2545

### LINKED ISSUE #2545: LUKS auto unlock via keyfile sometimes missing uuid
Thanks to forum member Jorma_Tuomainen for highlighting this bug. Our Auto unlock option within Web-UI initiated LUKS full disk encryption currently fails to assign an associated UUID, resulting in the following file name:

> /root/keyfile-

Linking to the original pull request that added LUKS capability for reference and details of the prebug behaviour: #1716

See the following forum post for context: https://forum.rockstor.com/t/auto-unlock-with-keyfile/8779

phillxnet:
Previously we used lsblk for this purpose, but sporadically lsblk would return an empty uuid for our just-created LUKS container. Thought to be due to it not having yet been updated re system state. Hence the use of the more specific LUKS command.

Fixes #2545 

## Include
- New get_luks_container_uuid(disk_byid) in luks.py.
- Fresh black formatting.
- Minor comment improvements.
- Leaving in-place the prior general procedure lsblk based mechanism as this may server elsewhere.

### DIFF
--- a/src/rockstor/storageadmin/views/disk.py
+++ b/src/rockstor/storageadmin/views/disk.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2020 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
@@ -13,7 +13,7 @@ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 General Public License for more details.
 
 You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
+along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
 import os
 import re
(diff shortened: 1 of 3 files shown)


phillxnet:
# Testing

Prior to this pr the issue detailed failure (no uuid in keyfile name) would appear sporadically but by far not the majority of the time. One attempt to reproduce took 4 consecutive tries before there was a failure. On the exact same test system (updated "Uses openSUSE Tumbleweed" Rockstor instance installed from our latest installer) post proposed patch no occurrence of the bug was observed over the creation and deletion of 4 LUKS containers in the following pattern:

1. reboot
2. make all 4 disk into LUKS containers.
3. delete all 4 LUKS containers (locked still).
4. repeat 2 & 3
5. repeat steps 1. to 4. - 3 more times.

The above was done on both the reproducer Tumbleweed base as well as a Lea 15.4 base. No instance of the missing uuid was observed.

Resulting in the creation and deletion of 24 LUKS containers (interleaved reboots) on each of the base OS platforms. All within the Rockstor Web-UI to provide an end-to-end testing scenario.

PR #2552, opened 2023-05-09
Make rockon-delete compatible with duplicate RockOn model entries #2549

### LINKED ISSUE #2549: Custom Rockon management deletion script enhancement.
This issue is triggered by the experience of forum user [`aymenso`](https://forum.rockstor.com/u/aymenSo):
https://forum.rockstor.com/t/get-returned-more-than-one-rockon-it-returned-2/8790

the Rockon delete script should consider an inconsistency scenario, where the same rockon is represented twice in the database: this should not occur, however we should probably prepare for that case, while working towards avoiding this scenario (possibly occurring during the json ingestion step).

https://github.com/rockstor/rockstor-core/blob/0249f5283203461d71093447611c762901f6f378/src/rockstor/scripts/rockon_delete.py#LL40C1-L57C1

May be the above scenario can be addressed by adding an outer loop over the `ro` object, so it would treat possible duplicates as individual rockons and delete accordingly?

Additional observation
While trying to reproduce the inconsistency (which I was not successful in), this is what I saw (all on 4.5.9-1):

- taking example of scrutiny rockon. Create two custom json files, differentiated by file name (scrutiny1.json, scrutiny2.json) and `containers` name (scrutiny1 and scrutiny2), but having the same Rockon name: `Scrutiny.local`. I made a non-relevant (I think) change to the description to be able to differentiate them, everything else remained the same.

- After a refresh, only one Rockon appeared, namely the scrutiny2.json's content.

image

- Without installing it, but directly going for the script for rockon deletion, it removed the custom scrutiny entry without error.

- Redoing the setup. Now installing the visible custom Rockon - no issue
image

- Refreshing the Rockon list, while the Scrutiny.local is installed. scrutiny1.json contents are not picked up, i.e. there is still only 1 Scrutiny.local (installed).

- Stopping and uninstalling Scrutiny.local using the UI. No issue. Selecting refresh, still only 1 (scrutiny2.json) is visible in the list.

- Reinstalling Scrutiny.local and while it's running use the delete script. Reload the web page, Scrutiny.local rockon has disappeared completely (as expected). Now Refresh Rockons again. scrutiny2.json is picked up as before.

So, not sure really how this inconsistency can happen

FroggyFlox:
Fixes #2549 
@phillxnet, @Hooverdan96, ready for review.

The rockon-delete currently only supports the "normal" case of a single RockOn model entry. In case one of these entries is duplicated, however, it currently fails.

This Pull Request (PR) thus adapts the current logic to get all entries (0, 1, or more) with a given name.

# Demonstration
The example laid out by @Hooverdan96 in #2549 was followed: create a local Rock-On named *Scrutiny.local* and update the list of available Rock-Ons in the webUI. Then, manually duplicate the `RockOn` model entry. We end up with the following:
```
rockdev:/opt/rockstor # psql -U rocky -d storageadmin -c "SELECT id,name,state,status,taskid FROM storageadmin_rockon WHERE name = 'Scrutiny.local';"
Password for user rocky: 
 id |      name      |   state   | status  | taskid 
----+----------------+-----------+---------+--------
 87 | Scrutiny.local | available | stopped | 
 85 | Scrutiny.local | available | stopped | 
(2 rows)
```

We then run the updated script:
```
rockdev:/opt/rockstor # poetry run delete-rockon Scrutiny.local
The metadata for the Rock-On named Scrutiny.local (id: 82) has been deleted from the database
The metadata for the Rock-On named Scrutiny.local (id: 84) has been deleted from the database
```

Verify it is gone:
```
rockdev:/opt/rockstor # poetry run delete-rockon Scrutiny.local
No Rock-On named Scrutiny.local was found in the database

rockdev:/opt/rockstor # psql -U rocky -d storageadmin -c "SELECT id,name,state,status,taskid FROM storageadmin_rockon WHERE name = 'Scrutiny.local';"
Password for user rocky: 
 id | name | state | status | taskid 
----+------+-------+--------+--------
(0 rows)
```

The script still works for a Rock-On with a single entry (normal case):
```
rockdev:/opt/rockstor # poetry run delete-rockon Jellyfin
The metadata for the Rock-On named Jellyfin (id: 83) has been deleted from the database
```

Note that we also have verification by the two forum users who reported this issue that the updated script resolved their duplicate `RockOn` entries:
- [https://forum.rockstor.com/t/get-returned-more-than-one-rockon-it-returned-2/8790/19](https://forum.rockstor.com/t/get-returned-more-than-one-rockon-it-returned-2/8790/19?u=flox)
- [https://forum.rockstor.com/t/get-returned-more-than-one-rockon-it-returned-2/8790/22](https://forum.rockstor.com/t/get-returned-more-than-one-rockon-it-returned-2/8790/22?u=flox)

# Testing
```
Ran 253 tests in 12.550s

OK
```

### DIFF
--- a/src/rockstor/scripts/rockon_delete.py
+++ b/src/rockstor/scripts/rockon_delete.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 
 """
-Copyright (c) 2012-2020 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <http://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


Hooverdan96:
Most excellent!

phillxnet:
@FroggyFlox My apologies for taking so long to see to this pull request. Merging directly.

PR #2559, opened 2023-05-25
Improve permissions on crontabs created by rockstor

### LINKED ISSUE #2556: (t) Suboptimal permissions on crontabs created by Rockstor
This issue results from a discussion with @phillxnet related to #2540: https://github.com/FroggyFlox/rockstor-core/commit/5fb66a66e1e0d0a0baf68c13c556a70d7181c04a#r114644668

As detailed and demonstrated by @phillxnet in that conversation thread, our current creation of crontabs keeps the default mode of 644 whereas security hardening would recommend keeping these files accessible by `root` only.

It is thus proposed that we ensure a mode of 600 for the creation of our crontabs:
- scheduled tasks: https://github.com/rockstor/rockstor-core/blob/8e65ada122112ac45cebd1dd5dd03c0ce6796069/src/rockstor/smart_manager/views/task_scheduler.py#L88
- replication service: https://github.com/rockstor/rockstor-core/blob/8e65ada122112ac45cebd1dd5dd03c0ce6796069/src/rockstor/smart_manager/views/replication.py#L40

FroggyFlox:
Fixes #2556 
@phillxnet, @Hooverdan96, ready for review.

We currently do not enforce any specific permissions on the crontabs generated by Rockstor. This can lead to some instances where these are set with a mask of `644`, for instance.
This pull request (PR) proposes to enforce a more restrictive mask of `600` at creation of these crontabs.

Note that this PR also proposes to slightly change the logic used for the creation of these files. Indeed, we currently edit/write the file "in-place", at its final destination (for instance, `/etc/cron.d/rockstortab`). In this PR, we use of potentially safer yet slightly slower (just a couple more calls) process akin to what is used more widely throughout the project:
1. Create a tempfile
2. Write the contents as needed
3. Set permission on this file
4. Copy to its final destination (`/etc/cron.d/rockstortab` for instance)

# Demonstration
1. Manually set one of the 2 crontabs in question with different permissions:
```
rockdev:/opt/rockstor # chmod 644 /etc/cron.d/rockstortab 
rockdev:/opt/rockstor # ls -lah /etc/cron.d/rockstortab
-rw-r--r-- 1 root root 193 May 25 14:56 /etc/cron.d/rockstortab
```
2. Make a change to this crontab from the webUI; in this case, I deleted an existing scheduled task
3. The crontab is re-written with proper persmissions:
```
rockdev:/opt/rockstor # ls -lah /etc/cron.d/rockstortab
-rw------- 1 root root 124 May 25 15:08 /etc/cron.d/rockstortab
```

# Tests
On a Leap 15.4 VM:
```
rockdev:/opt/rockstor # cd src/rockstor/ ; poetry run django-admin test ; cd -
Ran 253 tests in 14.082s

OK
```


phillxnet (approved):
@FroggyFlox Thanks for creating and seeing to this issue so quickly. Much appreciated.

Simple test using an rpm build with suggested changes included:

Creating a new scheduled task results in:
```
ls -la /etc/cron.d/rockstortab
-rw------- 1 root root 192 May 27 16:19 /etc/cron.d/rockstortab
```
As intended.
Scheduled task concerned executed:

tasks-executed-successfully
```
May 27 16:20:01 rleap15-3 CRON[1987]: (root) CMD (/opt/rockstor/.venv/bin/st-snapshot 1 \*-*-*-*-*-*)
...
May 27 16:25:01 rleap15-3 CRON[3007]: (root) CMD (/opt/rockstor/.venv/bin/st-snapshot 1 \*-*-*-*-*-*)
```

Thanks for easing the review process by splitting out the black/string.format changes into a separate commit.

PR #2567, opened 2023-06-06
Preliminary python 3.6 port - development #2564

### LINKED ISSUE #2564: Update pyproject.toml and poetry.lock for python 3.6
As part of our move to Python 3 we must first establish our new python 3 dependencies via Poetry. And resolve all issues re dependencies form there on.

phillxnet:
As part of our move to Python 3 we must first establish our
initial python 3 dependencies via Poetry.

Fixes #2564 

## Includes:
- Update pyproject.toml and poetry.lock for python 3.6.
- Remove int long designator from settings.py: Python 3
no longer uses the "L" suffix.
- Use full path in initial import conversions. Python 3
has more stringent import path requirements.
- Update imports for 3.6 email module. These differ in Python 3.
- Encode unicode line for system.osi.md5sum.
TypeError: Unicode-objects must be encoded before hashing
- Tidy/update init_sftp_config() re Python 3.
There was a subtle file pointer behaviour change
between 2.7 and 3.6 which broke this mechanism.
- Address TODO re custom opener re Python 3.
- Services page & Users page - Python 3 style sorted():
"cmp" sorted parameter and buildin function dropped
in Python 3.
- init fix of groups page by adaptation to str type.
- Replace chardet dependency with charset-normalizer:
also used by requests.
- Removes redundant encoding on already encoded groupname.
- Init dashboard fix via gevent/websocket update.
Thanks to @FroggyFlox for investigations here.
- Init fix of get_users() - non admin users did not show.
Move to more modern - time limited subprocess.run().
- fix samba, disk partition, rockon, re iteritems()
In Python 3 we use iter(d.items()) instead:
https://peps.python.org/pep-0469/
- Address TODO re ascii to utf-8 on SMART custom options.
- Untested bytecast in smart.py:
custom options autodev related byte assertion.

# Independent commits for attribution from @FroggyFlox 
- Move UDEVADM and SHUTDOWN definitions to runtime constants.
Removes now redundant and incompatible mechanism.
- Move passwd to str and use crypt.mksalt() to generate salt.
Updates this mechanism re Python 3 capabilities.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -7,12 +7,26 @@ optional = false
 python-versions = "*"
 
 [[package]]
-name = "chardet"
-version = "4.0.0"
-description = "Universal encoding detector for Python 2 and 3"
+name = "cffi"
+version = "1.15.1"
+description = "Foreign Function Interface for Python calling C code."
 category = "main"
 optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+python-versions = "*"
+
+[package.dependencies]
+pycparser = "*"
+
+[[package]]
+name = "charset-normalizer"
+version = "2.0.12"
+description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
+category = "main"
+optional = false
+python-versions = ">=3.5.0"
+
+[package.extras]
+unicode_backport = ["unicodedata2"]
 
 [[package]]
 name = "dbus-python"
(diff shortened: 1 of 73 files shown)


phillxnet:
N.B. Although these changes enable most functionality within a far newer Python 3.6 local .venv, they break rpmbuild as we require all tests to pass there. However the key purpose of this pull request is to enable spin-off issues to focus developer efforts on ease parallel tasks on fixing the following:

```
Ran 252 tests in 20.288s

FAILED (failures=12, errors=6)
```
Via clearly defined tasks to restore our testing channel to a build-able and thus release capable status.

phillxnet:
@FroggyFlox & @Hooverdan96 

Given this PR gets us to a source build state under python 3.6 (but not rpm buildable due to unit test failure) I propose that we merge it into testing and address the unit test failures as a matter of priority. As many of the tests fail due to Python 3.6 differences re unicode / string / byte  i.e. u"some text" etc we likely need a dedicated test issue/s. However we also need to avoid stepping on each others toes. Some tests will of course reflect tested code changes regarding the underlying Python changes we have introduced here.

FroggyFlox:
Agreed on merging this; that's already in an excellent state, relatively speaking of course. I was going to say "starting point" but that's actually much better than that.

Hooverdan96:
I know, this is a very minor detail and I don't know where to put this, but since you were moving some of the paths from the `settings.py` to `constants.py`, I noticed the `shutdown` path. In the spirit of simplification, would it make sense to use `systemctl poweroff` instead of the `shutdown` command/path to get rid of another setting/path?

PR #2573, opened 2023-06-06
Py3.6 update test_btrfs.py and use built-in mock #2571

### LINKED ISSUE #2571: Py3.6 Errors in (rockstor.fs.tests.test_btrfs.BTRFSTests)
Post #2567 in testing branch we have the following test failures in rockstor.fs.tests.test_btrfs.BTRFSTests

```
cd /opt/rockstor/src/rockstor/fs
poetry run django-admin test -p test_btrfs.py -v 2
```
Results in:
```
ERROR: test_balance_status_internal_unknown_mnt (rockstor.fs.tests.test_btrfs.BTRFSTests)
  File "/opt/rockstor/src/rockstor/fs/tests/test_btrfs.py", line 789, in test_balance_status_internal_unknown_mnt
    result = balance_status_internal(pool)
  File "/opt/rockstor/src/rockstor/fs/btrfs.py", line 2238, in balance_status_internal
    if unallocated >= 0:
TypeError: '>=' not supported between instances of 'NoneType' and 'int'

ERROR: test_balance_status_internal_unknown_parsing (rockstor.fs.tests.test_btrfs.BTRFSTests)
  File "/opt/rockstor/src/rockstor/fs/tests/test_btrfs.py", line 754, in test_balance_status_internal_unknown_parsing
    result = balance_status_internal(pool)
  File "/opt/rockstor/src/rockstor/fs/btrfs.py", line 2238, in balance_status_internal
    if unallocated >= 0:
TypeError: '>=' not supported between instances of 'NoneType' and 'int'

ERROR: test_get_pool_raid_levels_identification (rockstor.fs.tests.test_btrfs.BTRFSTests)
  File "/opt/rockstor/src/rockstor/fs/tests/test_btrfs.py", line 297, in test_get_pool_raid_levels_identification
    None, raid_levels_tested, btrfs_fi_di, return_dict
TypeError: 'NoneType' object is not callable
```
They are suspected as also having a knock-on effect with some Pool related tests but this will be more apparent once these 3 tests, or tested code has been fixed re our Py3.6 move.

phillxnet:
As of Python 3.3 mock is available via Python standard library as unittest.mock:
https://docs.python.org/3.6/library/unittest.mock.html https://pypi.org/project/mock/
Includes
- "btrfs.py" fix re Python 3.6 and None type comparison.
- Drop mock dependency.

Fixes #2571 

This patch break a large number of other tests that we also need to transition to the built-in mock/patch in unittest.
A spin-off issue will be created and linked.

## Testing
```
cd src/rockstor/fs/
poetry run django-admin test --settings=settings -v 3 -p test_btrfs*
...
Ran 58 tests in 0.078s

OK
...
```

### DIFF
--- a/src/rockstor/fs/tests/test_btrfs.py
+++ b/src/rockstor/fs/tests/test_btrfs.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2017 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2017 RockStor, Inc. <https://rockstor.com>
 This file is part of RockStor.
 RockStor is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published
@@ -10,10 +10,12 @@ WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 General Public License for more details.
 You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
+along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
 import json
 import unittest
+from unittest import mock
+from unittest.mock import patch
 from datetime import datetime
 from fs.btrfs import (
     get_pool_raid_levels,
@@ -41,7 +43,14 @@ from fs.btrfs import (
     scrub_status_extra,
     get_pool_raid_profile,
 )
-from mock import patch
+
+
+"""
+The tests in this suite can be run via the following commands:
+
+cd /opt/rockstor/src/rockstor/fs
+poetry run django-admin test --settings=settings -v 3 -p test_btrfs*
+"""
 
 
 class Pool(object):
@@ -52,13 +61,6 @@ class Pool(object):
 
 
 class BTRFSTests(unittest.TestCase):
-    """
-    The tests in this suite can be run via the following commands:
-    N.B. 'root' dir of rockstor is normally /opt/rockstor
-    cd /opt/rockstor/src/rockstor/fs
-    poetry run django-admin test --settings=settings -v 3 -p test_btrfs*
-    """
-
     def setUp(self):
         self.patch_run_command = patch("fs.btrfs.run_command")
         self.mock_run_command = self.patch_run_command.start()
@@ -73,7 +75,7 @@ class BTRFSTests(unittest.TestCase):
         self.mock_os_path_exists = self.patch_os_path_exists.start()
 
     def tearDown(self):
-        patch.stopall()
+        mock.patch.stopall()
 
     # # sample test
     # def test_add_pool_mkfs_fail(self):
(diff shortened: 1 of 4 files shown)


phillxnet:
@FroggyFlox & @Hooverdan96
As detailed in this first of our testing fixes under Py3.6 we have nearly 30 tests that now fail due to the dropping of the now outdated mock library. I'll create and start on an issue to address this separately.

src/rockstor/fs/tests/test_btrfs.py
```diff
@@ -10,10 +10,12 @@
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 General Public License for more details.
 You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
+along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
 import json
 import unittest
+from unittest import mock
+from unittest.mock import patch
```
FroggyFlox, line 18: @phillxnet ,
I'll show my lack of understanding here but: is this import necessary if we do use `mock.patch()` and `mock.patch.stopall()` for instance?

src/rockstor/fs/tests/test_btrfs.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2017 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2017 RockStor, Inc. <https://rockstor.com>
```
Hooverdan96, line 2: no functional value add, but maybe also adjust the copyright years: `2012-2023`

src/rockstor/fs/tests/test_btrfs.py
phillxnet: I had thought the same actually but it seems PyCharm was a little confused. Until I did this explicitly there were a number of flags. I'm keen to merge this so that I can approach the linked issue. I'll retrofit if all looks OK with the other 30 or so related changes.

src/rockstor/fs/tests/test_btrfs.py
phillxnet: Do you want to pop in the changes then - I can then merge this working pr and work on the other 30 unit test mock transitions. :)

phillxnet: i.e. modify your suggestion here so I can commit it.

Hooverdan96:
```suggestion
Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
```

phillxnet: Thanks @Hooverdan96 These are important - as well as the https as I'm trying to remove as many warnings as possible form the editor so it become useful again and it always flags all http as an issue !!

FroggyFlox:
Makes sense!
I too tried in PyCharm first and it seemed ok with the single import... It was seeing `stopall()`as coming from `unittest.mock`... that's all I looked at though, so I agree with you: we'll look at that after this.

PR #2583, opened 2023-06-07
Save network usage processed data as List object #2577

### LINKED ISSUE #2577: Py3.6 Dashboard Network widget not functional
This problem was first reported in: https://github.com/rockstor/rockstor-core/pull/2565#issuecomment-1576937019

Post #2567, we have an empty Network widget:
image

In the browser's developer console, we see the following message when resizing the Network widget.
```
Uncaught TypeError: this.NetworkChart is null
    resize https://buildvm/static/storageadmin/js/storageadmin.js:22189
    jQuery 2
        dispatch
        handle
```

It does seem that the data for the network widget do not come through.

FroggyFlox:
Fixes #2577 
@phillxnet, ready for review

The network widget was not displaying any data due to a change in what the `map()` function returns:
- Python2: a list
- Python3: an iterator

We then pick many elements of that object to construct our "payload" to the widget. Because is is now an iterator, this worked only the first element, but then returned empty for all others, leading to an empty payload to the widget.

Saving this as a list allows us to get back to previous behavior, resulting in a functional widget:
image

### DIFF
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2021 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


phillxnet (approved):
@FroggyFlox Thanks for sorting this out, and for the detailed explanation.
Adding the list cast locally and what do you know:
pr-review-pic-working-net-widget

Nice. Merging now directly.

PR #2589, opened 2023-06-08
Py3.6 Update certifi, urllib3, idna #2588

### LINKED ISSUE #2588: Py3.6 Update certifi, urllib3, idna
Post #2567 in testing branch we are no longer limited by Py2.7. It is proposed that we update these secondary dependencies of our direct **requests** dependency by unpinning their versions.

In this case, now we use Poetry, it is proposed that we remove the version constraint in our pyproject.toml
https://pypi.org/project/certifi/ Latest still 3.6 compatible.
https://pypi.org/project/idna/ Latest still 3.6 compatible.
https://pypi.org/project/urllib3/ Latest need 3.7.

phillxnet:
Remove version pinning as these are secondary dependencies on our unchanged direct dependency of requests.
"poetry update" resulted in:

  • Updating certifi (2021.10.8 -> 2023.5.7)
  • Updating idna (2.10 -> 3.4)
  • Updating urllib3 (1.26.12 -> 1.26.16)

Fixes #2588

## Testing
We have no change from our current testing branch unittest status of:
```
Ran 253 tests in 25.671s

FAILED (failures=2, errors=1)

```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,10 +1,10 @@
 [[package]]
 name = "certifi"
-version = "2021.10.8"
+version = "2023.5.7"
 description = "Python package for providing Mozilla's CA Bundle."
 category = "main"
 optional = false
-python-versions = "*"
+python-versions = ">=3.6"
 
 [[package]]
 name = "cffi"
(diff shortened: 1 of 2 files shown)


phillxnet:
A system was build via source method (/opt/rockstor removed prior) from this patch applied to current testing and all tests run as reported. Log function subscription, scheduled tasks all appear to be working as before.

As we are early testing here still lets get these in  - it also helps clear some of the way for our next Python update.

PR #2597, opened 2023-07-03
Add GitHub Action to trigger post-release updates across repositories #2596

### LINKED ISSUE #2596: Release process  changes in accompanying repositories
Once a release in rockstor-core is created, we currently need to also proceed with a few modifications in two of our accompanying repositories.:
- rockstor/rockstor-jslibs
- rockstor/rockstor-rpmbuild

This process is rather simple but requires quite a bit of manual work that is very time-consuming. We should instead strive to automate the process.

FroggyFlox:
Fixes #2596 

After each release, we are currently manually performing the required changes and udpates across our repositories.
This pull request proposes to add a GitHub Action workflow to trigger these changes automatically.

This workflow needs to be manually triggered against either the `master` or `testing` branch, according to the release that was created. This workflow will gather the release-version information, and trigger the workflows triggering the required edits in the two other repositories:
  - in rockstor-jslibs: the workflow simply creates a new release according to version, targeting the base branch (`master` or `testing` according to what was chosen when manually triggering the workflow here).
  - in rockstor-rpmbuild: the workflow edits `rockstor.spec` according to version-release and adds the changelog according to the release notes present in rockstor-core. It then pushes the changes to rockstor-rpmbuild (as a new branch) and creates a pull request against the base branch (`master` or `testing`).

@phillxnet: in order to have this workflow run on either the `master` or `testing` branch, it needs to be present in both branches. I'm submitting this PR against `master` as it needs to be present in `master` even if we want to run it against `testing`. This means we'll have to add it to `testing` as well.

### DIFF
--- a/.github/workflows/trigger_updates.yml
+++ b/.github/workflows/trigger_updates.yml
@@ -1,0 +1,54 @@
+name: Trigger updates
+on: workflow_dispatch
+jobs:
+  set-vars:
+    name: Get version, release, and notes
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.set-ver.outputs.version }}
+      release: ${{ steps.set-rel.outputs.release }}
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Set Version
+        id: set-ver
+        run: |
+          echo "version=$(gh release list --limit 1 -R rockstor/rockstor-core | cut -f1 | cut -d "-" -f1)" >> "$GITHUB_OUTPUT"
+      - name: Set Release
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox Thanks for seeing to this much needed bit of automation - As stated in another of the 3 pull requests across as many repos I've no real experience with this so have only done a read through and implemented what I understand from our side channel chats is required in the repos. So I'll pop this in and look to cheery picking it into testing as well. That way we can begin our 'field trials' of this within the Org as I know you have already tested the mechanism extensively in a regular personal GitHub context. Likely any failing is going to be something I've missed in setup so I'll apologies in advance :).

PR #2621, opened 2023-07-19
Update specifics of Python version in pyproject.toml #2620

### LINKED ISSUE #2620: Update specifics of Python version in pyproject.toml
All our Leap target versions have our testing base Py3.6 requirement satisfied by OS default of Python 3.6.15. However to accommodate our current Tumbleweed target, post our current Py3.6 Milestone: https://github.com/rockstor/rockstor-core/milestone/20 we could resource Py3.8 Tumbleweed packages (default repos). But with the proviso that we have more work to do on the full compatibility front (next Milestone). However this would server as a test bed for such things which is our main use of Tumbleweed anyway.

phillxnet:
Move from Python 3.6.*, to above 3.6 and below 3.9. Primarily to accommodate both Leap default Python, and Tumbleweed optional Python.

Fixes #2620

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,10 +3,13 @@
 set -o errexit
 
 # Install Poetry, a dependency management, packaging, and build system.
-# We currently require Python 2.7 compatibility which was last in v1.1.15.
+# We currently require Python 3.6 compatibility which was last in v1.1.15.
 # We use the official installer which installs to: ~/.local/share/pypoetry.
 # The installer is python 3 only: https://python-poetry.org/docs/#installation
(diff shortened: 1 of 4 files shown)


phillxnet:
On a Tumbleweed client we have the following behaviour:
```
rtumbleweed:/opt/rockstor # poetry install
The currently activated Python version 3.10.12 is not supported by the project (>= 3.6, < 3.9).
Trying to find and use a compatible version. 
Using python3.8 (3.8.17)
Creating virtualenv rockstor in /opt/rockstor/.venv
Installing dependencies from lock file
...
```

Issue #2633, opened 2023-07-31
Improve OS independence re unit tests


phillxnet:
All of the following likely represent unit test improvements to be had regarding unwanted / unrequited OS sensitivity during unit test runs. I.e. insufficient mocking of OS components:

- [x] FileNotFoundError: [Errno 2] No such file or directory: '/usr/sbin/dmsetup': '/usr/sbin/dmsetup'
- [x] FileNotFoundError: [Errno 2] No such file or directory: '/etc/postfix/master.cf'
- [x] 'Failed to restart avahi-daemon.service: Unit avahi-daemon.service not found.'
- [x] Failed to enable unit: Unit file snmpd.service does not exist.
- [x] No such file or directory: \'/etc/postfix/master.cf
- [x] No such file or directory: \'/usr/bin/nmcli\': \'/usr/bin/nmcli\'\n'
- [x] No such file or directory: \'/usr/sbin/exportfs\': \'/usr/sbin/exportfs\'\n'
- [x]  No such file or directory: \'/opt/rockstor/static/config-backups\'\n']
- [x] File "/opt/rockstor/src/rockstor/storageadmin/tests/test_disks.py", line 84, in test_disk_scan
- [x] User (lp) does not exist.', 'NoneType: None\n

See the following rpmbuild issue development / comments for the reproducer:
https://github.com/rockstor/rockstor-rpmbuild/issues/28

The aim here is to minimise the system required in order to successfully rpmbuild - with %check (which runs all our unit tests) enabled. On a rockstor installer base non of the above are seen - however all are thought to be unwanted OS sensitivities: i.e. calls to real files etc.

phillxnet:
Our last test that fails during an rpmbuild on a Minimal_VM that has not had rockstor installed previously and prepared only as detailed in ongoing issue https://github.com/rockstor/rockstor-rpmbuild/issues/28 is test_user.py.

An effort will be made to move this test to using fixtures within this issue, but only if it is not too time-consuming. There are a number of remarked out test in this file already so I'm reluctant to add another. And also reluctant to modify the existing test to rely, as it does currently, on the existence of a real user. Which in the failure observed here does not actually exist on a Minimal_VM as prepared for/by rpmbuild (BuildRequries).

phillxnet:
As we now have the minor tested-code change merged into testing branch via the linked issue/pr - I will squash and rebase the work-done against this issue, which can then be presented in its own pull request.

phillxnet:
[...]

PR #2636, opened 2023-08-01
user admin enable without password change not internal error #2635

### LINKED ISSUE #2635: user admin enable without password change not internal error
When adding Web-UI admin to an existing user, without also changing the password, results in us responding as per HTTP 500, i.e. back trace and internal server error indication. What is more appropriate here is HTTP 400: client error. The client has not fulfilled the requirement to also submit a password while enabling Web-UI admin.

phillxnet:
Return HTTP_400_BAD_REQUEST not HTTP_500_INTERNAL_SERVER_ERROR.

Fixes #2635 

## Before proposed change

We return an internal server error code of 500 to our client side which results in our Web-UI providing a back-trace/debugging type feedback on an internal error that has not happened.

before-with-500

## After patch

The change to a 400 bad client request error code results in the following more appropriate Web-UI dialog:

after-with-400

## Counterpart unit test changes

This pull request and its associated issue address a small tested code modification surfaced while improving our tests in this area in work done against issue:
"Improve OS independence re unit tests #2633"
The appropriate testing change will be included in a referenced pull reqeust against that issue.

### DIFF
--- a/src/rockstor/storageadmin/views/user.py
+++ b/src/rockstor/storageadmin/views/user.py
@@ -257,7 +257,7 @@ class UserDetailView(UserMixin, rfc.GenericView):
                                 "enable admin access. Please provide "
                                 "a new password."
                             )
-                            handle_exception(Exception(e_msg), request)
+                            handle_exception(Exception(e_msg), request, status_code=400)
                         auser = DjangoUser.objects.create_user(username, None, new_pw)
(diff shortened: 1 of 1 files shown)


phillxnet:
Merging this into testing to enable a rebase of the work in progress associated with the test improvements that brought this issue to light.

PR #2641, opened 2023-08-04
Address compression cosmetics #2640

### LINKED ISSUE #2640: Compression cosmetics
<blockquote cite="https://github.com/rockstor/rockstor-core/pull/2619#pullrequestreview-1558934778_">
@StephenBrown2 Hello again.
Functional/cosmetic review only here as we can re-visit in follow-up issue-pr sets as and when.

## Cosmetics 
It would be nice to normalise our ordering across and within tooltips/selectors.

### Share compression options
Take a look at for example our share detail where the tooltip order is not that of the selector. Not a show-stopper but a nice to have in some usability follow-up as I'm keen to get this functional extension of a PR in as-is given your work to date to get it working and roughly in shape.

- tooltip and selector
share-detail-compression-tooltip_and_selector

Tooltip ordering is consistent and I agree with putting the new zstd option at the end: sort of chronological ordering, at least for us. So we likely just need to re-order our share selector. But in another pr I think as only a cosmetic. On the same note we have existing issues re formatting etc. Note our missing space after "lzo:" in share and pool tooltips. Oh and thanks for throwing in some other little fixes re "it's" correction, that one was probably my mistake actually.

### Pool compression options
**Overview**
- tooltip
pool-overview-compression-tooltip
- selector
pool-overview-compression-selector

**Detail**
- tooltip
pool-detail-compression-tooltip
- selector
pool-overview-compression-selector

As you noted in your PR we really need to abstract these out to a single source of truth, easier said than done but it would be another great addition - especially as this is not going to be our last addition here. That way future changes / corrections would be far more trivial and have consistency by design. But bit by bit and all good.

As you also noted, we have a reflection of the setting at the pool level in our active mount options:
zstd-pool-active-mount-options
Nice.

Thanks again for your effort here: much appreciated. Likely we need some infrastructure improvements in this area to support further enhancements but again: bit by bit.
</blockquote>

_Originally posted by @phillxnet in https://github.com/rockstor/rockstor-core/pull/2619#pullrequestreview-1558934778_

I've done some work to pretty up the tooltip text, and re-ordered to be consistent, PR coming shortly. Wanted to reference this in a new issue for tracking.

StephenBrown2:
This PR pretties up the Compression Info Tooltip, and also ensures that the ordering is consistent throughout the repo as to which algorithms are supported.

Pool overview tooltip:
image
Pool overview dropdown:
Screenshot from 2023-08-04 16-04-11

Pool detail tooltip:
image

Pool detail dropdown:
Screenshot from 2023-08-04 16-08-57

Share tooltip and dropdown:
Screenshot from 2023-08-04 16-06-35

Fixes #2640

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -2333,7 +2333,7 @@ def get_property(mnt_pt, prop_name=None):
     properties. But if called with a single property then the value and type
     appropriate for that property ie:
     string for label (in presented in properties),
-    string for compression ie: lzo, zlib (if presented in properties), zstd,
+    string for compression ie: zlib (if presented in properties), lzo, zstd,
     and Boolean for prop_name='ro'
     If prop_name specified but not found then None is returned.
     N.B. compression property for subvol only, vol/pool uses mount option.
--- a/src/rockstor/settings.py
+++ b/src/rockstor/settings.py
@@ -363,7 +363,7 @@ REST_FRAMEWORK = {
 
 CONFROOT = '{}/conf'.format(BASE_DIR)
 CERTDIR = '{}/certs'.format(BASE_DIR)
-COMPRESSION_TYPES = ('lzo', 'zlib', 'zstd', 'no',)
+COMPRESSION_TYPES = ('zlib', 'lzo', 'zstd', 'no',)
 
 SNAP_TS_FORMAT = '%Y%m%d%H%M'
 
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/pool/compression_info.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/pool/compression_info.jst
@@ -6,12 +6,13 @@
                 {{pool.compression}}
             {{else}}
             <strong>
-            <a href="#" id="comprOptn" data-type="select" data-title="Pool compression algorithm.<br>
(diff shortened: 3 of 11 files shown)


.gitignore
```diff
@@ -42,3 +42,13 @@ node_modules/
 
 # VS Code
 .vscode
+
+# Rockstor build
```
StephenBrown2, line 46: I added these because I ended up working directly in the VM with VSCode remote dev, but this change can be separated out to it's own PR if you would prefer.

phillxnet: @StephenBrown2 These changes to .gitignore are a nice addition actually - and yes I think it would be good to have them split out into their own issue/pull request set. We will then have them to point to for folks that want to add similar dev environment additions/enhancements such as these. It will also help with cherry picking these changes, in isolation, if they end up being required in another branch.

Thanks for thinking of this and highlighting it. Plus it will help to split out 'things' from one another. Always nice.
It can also help with reviewing to have stuff split down into more manageable chunks.

phillxnet (approved):
@StephenBrown2 This is a nice improvement, and well presented. Much appreciated.

An rpm was build successfully with these changes and the resulting modified pages were as presented in the PR, i.e.:

share-compression-tooltip-and-selector-pr2641

phillxnet:
@StephenBrown2 if all goes to plan this should be included in 5.0.3-0.

Thanks again for the follow-up improvement here.

PR #2643, opened 2023-08-07
Update .gitignore with Rockstor build artifacts #2644

### LINKED ISSUE #2644: Rockstor build changes to .gitignore
> @StephenBrown2 These changes to .gitignore are a nice addition actually - and yes I think it would be good to have them split out into their own issue/pull request set. We will then have them to point to for folks that want to add similar dev environment additions/enhancements such as these. It will also help with cherry picking these changes, in isolation, if they end up being required in another branch.
> 
> Thanks for thinking of this and highlighting it. Plus it will help to split out 'things' from one another. Always nice.
> It can also help with reviewing to have stuff split down into more manageable chunks.

_Originally posted by @phillxnet in https://github.com/rockstor/rockstor-core/pull/2641#discussion_r1285107554_

StephenBrown2:
Per comment in https://github.com/rockstor/rockstor-core/pull/2641#discussion_r1285107554 , splitting these changes out as their own changeset.

These allow for development in a "live" Rockstor testing instance, without worrying about accidentally committing a database or the jslibs, etc.

Closes #2644

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,13 @@ node_modules/
 
 # VS Code
 .vscode
+
+# Rockstor build
+.initrock
+PKG-INFO
+certs/
+jslibs/
+poetry-install.txt
+rockstor-jslibs.tar.gz*
+rockstor-tasks-huey.db*
+static/


phillxnet (approved):
@StephenBrown2 Thanks, these are long overdue as it goes. Thanks for stepping up to this one. Much appreciated.

PR #2654, opened 2023-08-14
Web-UI update fails to recreate venv #2652

### LINKED ISSUE #2652: Web-UI update fails to recreate venv
Thanks to @FroggyFlox for doing much of the investigation on this one. During our experimental support for openSUSE Leap 15.5  (which in Minimal-VM form has only Py3.6 and no "python" in-path) we observed the following failure to re-create our dedicated .venv via Poetry:

```
rleap15-5:/opt/rockstor # cat /opt/rockstor/poetry-install.txt

  FileNotFoundError

  [Errno 2] No such file or directory: 'python': 'python'

  at /usr/lib64/python3.6/subprocess.py:1364 in _execute_child
      1360│                     if errno_num != 0:
      1361│                         err_msg = os.strerror(errno_num)
      1362│                         if errno_num == errno.ENOENT:
      1363│                             err_msg += ': ' + repr(err_filename)
    → 1364│                     raise child_exception_type(errno_num, err_msg, err_filename)
      1365│                 raise child_exception_type(err_msg)
      1366│
      1367│
      1368│         def _handle_exitstatus(self, sts, _WIFSIGNALED=os.WIFSIGNALED,
```
The same failure was not observed if our programmatically generated AT script that we hand off to for our updates was run via cli invokation by the root users (still resourcing AT). @FroggyFlox also found, among many other things, that this same failure could be reproduced on an openSUSE Leap 15.4 system that also did not have a "python" interpreter in the path, only a python3 one.

After quite the journey/exploration it tuns out that we have negative interference from an environmental variable in play that serves no need within our AT/rpm/zypper managed update procedure. This issues failing is also likely related to our recent wiping and re-creating of our .venv virtual environment during rpm upgrade (see https://github.com/rockstor/rockstor-rpmbuild/pull/31) that was required to negotiate our past and future python version jumps.

The proposed fix is to unset the VIRTUAL_ENV environmental variable via the addition of the following line in our AT intended update script:
```
unset VIRTUAL_ENV
```
As this setting end up confusing/failing our subsequent rpm/zypper managed rebuild of our venv according to the installing package version via it's %posttrans scriptlet. We have more tidy-up to do in this area as a whole with the following spin-off issues resulting:

https://github.com/rockstor/rockstor-core/issues/2651
https://github.com/rockstor/rockstor-core/issues/2650

A pull request with this proposal is to follow shortly.

phillxnet:
Unset no longer required, and counterproductive to Poetry venv rebuild, VIRTUAL_ENV environmental variable within the programmatically created AT update script.

Fixes #2652 

See issue text for development context.

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -463,6 +463,8 @@ def update_run(subscription=None, update_all_other=False):
                 )
             )
             atfo.write(pkg_refresh_cmd)
+            # Unset inherited VIRTUAL_ENV environmental variable before invoking rpm/zypper
+            atfo.write('unset VIRTUAL_ENV\n')
             # account for moving from dev/source to package type install:
             atfo.write(pkg_in_up_rockstor)
             # rockstor-bootstrap Requires rockstor which Requires rockstor-pre (initrock)


FroggyFlox:
Thanks a lot for taking the time to find the fix create the issue, and prepare the PR!

If I'm correct, this fix will apply only to future updates after the one containing this PR, but I don't see any other way so far. We always have the functioning update through cli so far.

phillxnet:
@FroggyFlox Re:
> If I'm correct, this fix will apply only to future updates after the one containing this PR

Yes, but we have only seen this failure (without by-hand recreation) on our as-yet unsupported 15.5 base OS. So getting this out before we begin that support was the hope.  Plus we only cover 15.5 in testing (officially) as of yet.

Thanks for the clarification on this one. Quite important point as it goes.

I've now successfully built and installed a 15.5 hosted rpm with this patch pre-applied, and given we have tested this fix in development re it's modified update script fixing the observed issue, I'll go ahead and get this merged.

PR #2655, opened 2023-08-15
Enhance development ease #2653

### LINKED ISSUE #2653: Enhance development ease
During remove development it can be handy to pass environmental variables to define our Django running mode: specifically the DEBUG setting. It would also aid in on-boarding to have a defined configuration for invoking, for example, runserver. We also have a static files failure re incomplete setting definition.

## Proposal re settings change from hardwired `DEBUG = False` is as follows:
- [x] Add the following more flexible definition:
```
# By default DEBUG = False, honour this by True only if env var == "True"
DEBUG = os.environ.get("DJANGO_DEBUG", "") == "True"
```

## Proposal re runserver in-project setting:

- [x] Incorporate PyCharms .run top directory containing 'runner' configurations that can then be easily imported along with the code.

## Proposal re currently non functional runserver static file serving.

We are missing the following setting from our Django settings STATICFILES_FINDERS:
(see: https://django-pipeline.readthedocs.io/en/latest/installation.html#installation)

- [x]  'pipeline.finders.PipelineFinder',
Thanks to @FroggyFlox for this find.

Also note that our settings file is as-yet un black formatted: unlike the vast majority of our Python code base (thanks to @FroggyFlox again).

phillxnet:
- Set Django debug via environmental variable.
- Add PyCharm /.run in-project runserver config.
- Add missing STATICFILES_FINDER config re django-pipeline.
- Black format of settings.py.
- Europe/Lisbon timezone - same as prior London setting; but reflects our Open Collective funding.

Fixes #2653

### DIFF
--- a/.run/rock-runserver.run.xml
+++ b/.run/rock-runserver.run.xml
@@ -1,0 +1,28 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="rock-runserver" type="Python.DjangoServer" factoryName="Django server">
+    <module name="rockstor" />
+    <option name="INTERPRETER_OPTIONS" value="-Wd" />
+    <option name="PARENT_ENVS" value="true" />
+    <envs>
+      <env name="PYTHONUNBUFFERED" value="1" />
+      <env name="DJANGO_SETTINGS_MODULE" value="src.rockstor.settings" />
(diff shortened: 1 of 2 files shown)


phillxnet:
## Testing
An rpm was build (15.5 host) and installed successfully with the product of this patch.

PR #2665, opened 2023-08-21
Use user.is_authenticated as an attribute #2664

### LINKED ISSUE #2664: (t) `models.User.is_authenticated()` will be deprecated in Django 2.0
When run using the Python warning flag (thanks @phillxnet!, #2655), we see the following warning:
```
/opt/rockstor/src/rockstor/storageadmin/views/home.py:68: RemovedInDjango20Warning: Using user.is_authenticated() and user.is_anonymous() as a method is deprecated. Remove the parentheses to use it as an attribute.
  if request.user.is_authenticated():
```

# Steps to reproduce
As this is used in our `home`, the easiest way to trigger this warning is to "fully refresh" any webpage. For instance, go to system>users and press Ctrl + Shift + R.

# Information
The warning itself is very clear and provides the solution. For background, we can have a look at the docs:
https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated

> is_authenticated[¶](https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated)
>
>    Read-only **attribute** which is always True (as opposed to AnonymousUser.is_authenticated which is always False). This is a way to tell if the user has been authenticated. This does not imply any permissions and doesn’t check if the user is active or has a valid session. Even though normally you will check this attribute on request.user to find out whether it has been populated by the [AuthenticationMiddleware](https://docs.djangoproject.com/en/1.11/ref/middleware/#django.contrib.auth.middleware.AuthenticationMiddleware) (representing the currently logged-in user), you should know this attribute is True for any [User](https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User) instance.

>  **Changed in Django 1.10:**
>
> In older versions, this was a method. Backwards-compatibility support for using it as a method will be removed in Django 2.0.

# Our use
As per the log file, our only use of this now-attribute (was method) is in `home.py`:
```
buildvm155:/opt/rockstor # grep -r --include "*.py" "is_authenticated" src/rockstor/
src/rockstor/storageadmin/views/home.py:    if request.user.is_authenticated:
```

FroggyFlox:
Fix #2664 

Prior to Django 1.10, `user.is_authenticated()` was a method. Since then, compatibility was ensured but this compatibility will be removed in Django 2.0.  See #2664 for more details and background.

This Pull Request (PR) simply proposes to move to using the `user.is_authenticated` attribute instead.

# Functional testing
Prior to this PR, fully refreshing a webpage (Ctrl + Shift + R) would show the warning described in #2664.
After this PR, this no longer appears.

# Unit testing
All tests still pass:
```
buildvm155:/opt/rockstor # cd src/rockstor/ && poetry run django-admin test ; cd -
Ran 262 tests in 14.807s

OK
```

### DIFF
--- a/src/rockstor/storageadmin/views/home.py
+++ b/src/rockstor/storageadmin/views/home.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2020 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
@@ -13,7 +13,7 @@ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 General Public License for more details.
 
 You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
+along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
 
 from django.shortcuts import render
@@ -27,6 +27,7 @@ from django.conf import settings
 import logging
 
 logger = logging.getLogger(__name__)
+
 
 def login_page(request):
     return render(request, "login.html")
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox Merging on the basis of the testing provided. As you say this does look like a simple drop-in replacement.
Thanks again for yet more preparation re the current Django milestone.

PR #2670, opened 2023-09-04
surface Distro Version to breadcrumb bar (2668)

### LINKED ISSUE #2668: Make OS base version indicator more obvious
We have had a number of reports indicating our OS version 'mouse-over' is not accessible / lacks affordance. It is proposed that we remove this mouse-over and instead indicate our base OS version directly in the header. Hopefully then removing the potential confusion with the Linux kernel version.

Current state example:
uses-openSUSE-mouse_over_example

Proposed text would then  read:

**Uses openSUSE Leap: 15.5 Linux: ...**

Hooverdan96:
Fixes #2668.  Instead of showing the distro version via the mouse over, this change will show it directly as part of the breadcrumb bar.

After the changes and `static` collection, the version now shows up alongside the distro name and kernel version:

image

A mouseover doesn't show up anymore.
Changes also survive reboots.

Not sure whether that was necessary, but in any case ran the tests:

```
Ran 253 tests in 40.992s
```

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/router.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/router.js
@@ -979,8 +979,7 @@ $(document).ready(function() {
     };
 
     var distroInfo = function(data) {
-        $('#distro-info').text("Uses " + data.distro);
-        $('#distro-info').attr('title', data.version);
+        $('#distro-info').text("Uses " + data.distro + ": " + data.version);
     };
 
     var displayLocaleTime = function(data) {


phillxnet:
@Hooverdan96 Thanks for seeing to this. Much appreciated.

phillxnet (approved):
@Hooverdan96 Thanks for stepping up to this one. Much appreciated.

Leap15 5-uses-header

Tumbleweed-uses-header

This looks to be exactly as intended. And should help with folks reporting their base OS more easily.

Issue #2680, opened 2023-09-23
Avoid system calls to interact with systemd services


FroggyFlox:
We currently use a lot of system calls via `run_command()` (a wrapper around `subprocess.Popen()`) to interact with various systemd units. See `system.services.py` for a very good list of examples. This is very cumbersome to operate, heavy to parse the outputs, and most likely rather wasteful.

A better approach would most likely be to directly interact with systemd DBus. While we already have the `dbus` library as a dependency for interacting with the SSSD DBus responder InfoPipe, there is a python library to simplify interacting with systemd units and services: `pystemd`

GitHub: https://github.com/systemd/pystemd
Their description seems to fit our needs exactly: 
> This library allows you to talk to systemd over dbus from python, without actually thinking that you are talking to systemd over dbus. This allows you to programmatically start/stop/restart/kill and verify services status from systemd point of view, avoiding executing `subprocess.Popen(['systemctl', ...` and then parsing the output to know the result.

The project is hosted by the `systemd` organization on Github, and seems actively maintained (https://github.com/systemd/pystemd/releases). It is also easily available on PyPI.org, making it easy to be added to our `pyproject.toml`.

It is licensed **LGPL-2.1**: https://github.com/systemd/pystemd/blob/main/LICENSE
It is currently supporting Python3.6 and later.
Note that to build it, we would need to following new system package: `systemd-devel`.

An example use on how to restart a service is described at https://github.com/systemd/pystemd/issues/60#issuecomment-895982402

It would allow us to access all properties and methods described at https://www.freedesktop.org/software/systemd/man/org.freedesktop.systemd1.html#Properties1

Hooverdan96:
This looks interesting and seems to allow for simplification from constructing the commands for systemd to a simpler way!
The example you posted up there, and the question posed therein about `sudo` would not apply to Rockstor, since we are running these things with root in general anyway?

Would this come before or after below design decision?

https://github.com/rockstor/rockstor-core/issues/2650

In my mind this would probably come before further moving items into the systemd realm, so the transition would be simpler, too. Or am I connecting two things that are really independent?

FroggyFlox:
> The example you posted up there, and the question posed therein about `sudo` would not apply to Rockstor, since we are running these things with root in general anyway?

That's something we'll have to verify, indeed. I'm thinking the same as you so far. I have only briefly tried with restarting the `rockstor` service that way and it was fine. It would need testing with a service such as sshd, I suppose.



[...]

Issue #2650, opened 2023-08-14
Replace supervisord with further systemd intergration


phillxnet:
At one time, before systemd, we did all our service orchestration via supervirord: we have progressively moved to using systemd services for our various 'parts' but have yet to migrate the following:

- gunicorn (/opt/rockstor/.venv/bin/gunicorn) - runs our main wsgi:application.
- replication daemon (/opt/rockstor/.venv/bin/replicad)
- data-collector (/opt/rockstor/.venv/bin/data-collector) - live-info
- Huey (/opt/rockstor/.venv/bin/django-admin run_huey ...) longer running tasks manager.

It is proposed that these be incorporated into rockstor.service via multiple ExecStart entries. Or possibly host only gunicorn in rockstor.service and create dedicated services for each of the other services. This would normalise all our service management on a single orchestration mechanism and one native to the base OK. It would also remove a central Python dependency.

There may however be concurrency/threading issues that are otherwise handled better by supervisord given it is itself, a python app.

phillxnet:
Observed pending issue regarding supervisord's use of a deprecated packaging technology:
```
May 30 10:34:37 machine-name systemd[1]: Starting Rockstor bootstrapping tasks...
May 30 10:34:38 machine-name poetry[22248]: /opt/rockstor/.venv/lib/python3.11/site-packages/supervisor/options.py:13: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
May 30 10:34:38 machine-name poetry[22248]:   import pkg_resources
```

phillxnet:
Closing as:
Fixed by #3040 

Thanks to @Hooverdan96 for the majority of the work on this fix.

PR #2682, opened 2023-09-30
Don't throw exception when getting supervisord service status #2681

### LINKED ISSUE #2681: [Services] Websocket failure to update service status
In the "System" > "Services" page, we have websocket that fetches the status of each service and sends it to our front-end to update the toggle buttons. This unfortunately is currently failing, resulting in a toggled button remaining "misaligned" (see screenshot below) until a manual page refresh:
image

By using Firefox's dev tools, we can see that the websocket is correctly created and removed upon visiting/leaving the "Services" page, but it does not seem to send/return any data on the services' statuses, hence the failure to refresh the toggle button state.

It turns out this failure results from an Exception thrown when checking the status of the "replication" service. This exception is normal (when the replication service is OFF), but we shouldn't fail the checking of service status as a result.

FroggyFlox:
Fixes #2681 
@phillxnet, @Hooverdan96: ready for review.

We currently throw an Exception when checking the status of a surpervisor-controlled service such as replication. This leads to an exception being thrown even when expected (service is off), which can itself interrupt other parts such as our `get_services` websocket.

This pull-request adds a new arg to `superctl()` allowing for the possibility to not throw an exception. Note that `run_command()`'s default for throw is `True`, so we respect this here.

# Functional testing
The Samba service was toggled ON.
Within 15 sec (websocket "refresh" interval), the status of all services was refreshed and the toggle button was re-aligned with the others, and stayed ON.
The Samba service was then toggled back OFF and the opposite was observed: button switched to OFF and was re-aligned to the other buttons within 15 secs.

# Unit testing
All tests still pass. Note that this was tested on Leap 15.4 only:
```
buildvm:/opt/rockstor # cd src/rockstor/ && poetry run django-admin test ; cd -
Ran 262 tests in 15.827s

OK
```

### DIFF
--- a/src/rockstor/system/services.py
+++ b/src/rockstor/system/services.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2023 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
@@ -13,7 +13,7 @@ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 General Public License for more details.
 
 You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
+along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
 
 import os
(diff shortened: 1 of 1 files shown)


phillxnet (approved):
@FroggyFlox This is a nice find. Thanks.
As discussed we need to move all our stuff to systemd only, rather than our current service  manger inside another arrangement.
Re: "Replace supervisord with further systemd intergration" #2650
and your follow-up of:
"Avoid system calls to interact with systemd services" #2680
But all in good time. And my apologies for missing this earlier in our more recent refactoring re our move to Poetry.
I't been bugging me for a while now that we kept seeing that exception; I just hadn't realised the ramifications.

And this is a nice fix for what we have currently.

Cheers.

PR #2701, opened 2023-10-14
Update python-socketio & python-engineio to latest #2591

### LINKED ISSUE #2591: Update python-socketio & python-engineio to latest
Post #2567 in testing branch we are no longer limited by Py2.7. It is proposed that we update this primary dependency accodingly. From https://pypi.org/project/python-socketio/ we have the latest version still supporting our now prior Py3.6 base. [EDIT: we now use Py3.9 as of #2692],

N.B. we use a js library as our client so we must co-ordinate an update accordingly in our https://github.com/rockstor/rockstor-jslibs repository.

Linking to the last update of this client js library for context:
https://github.com/rockstor/rockstor-core/issues/1503

Note also the interdependency within our Python dependency on 
```
python-engineio 2.3.2 Engine.IO server
└── six >=1.9.0
python-socketio 1.6.0 Socket.IO server
├── python-engineio >=1.0.0
│   └── six >=1.9.0 
└── six >=1.9.0
```

Also noting the larger context of replacement of this mechanism with the Django build-in channels 4.0:
https://github.com/django/channels/
https://channels.readthedocs.io/en/latest/releases/4.0.0.html
Which in turn requires at least Py3.7 [EDIT: we use Py3.9 now] and our next planned Django LTS version of 3.2, see Milestone: https://github.com/rockstor/rockstor-core/milestone/24

As the original Channels (Channels 1) underwent a major breaking change re-write in Channels 2:
https://channels.readthedocs.io/en/latest/releases/2.0.0.html
we want to avoid Channels 1 (which is compatible with our current Django) entirely.

phillxnet:
Raise pinned versions to latest available:
## Includes:
- Revised nginx reverse proxy for data_collector app.
- Replace gevent-websocket with simple-websocket.
- Replace gevent for regular Python threads in socketio.Server.

Fixes #2591 

Requires resolution of / depends upon: https://github.com/rockstor/rockstor-jslibs/issues/36

### DIFF
--- a/etc/nginx/nginx.conf
+++ b/etc/nginx/nginx.conf
@@ -73,11 +73,15 @@ http {
 			proxy_pass http://127.0.0.1:8000/;
 		}
 		location /socket.io {
-			proxy_pass http://127.0.0.1:8001;
+			proxy_pass http://127.0.0.1:8001/socket.io;
+			proxy_set_header Host $http_host;
+			proxy_set_header X-Forwarded-Proto $scheme;
+			proxy_set_header X-Forwarded-Host $http_host;
 			proxy_redirect off;
 			proxy_http_version 1.1;
+			proxy_buffering off;
 			proxy_set_header Upgrade $http_upgrade;
-			proxy_set_header Connection "upgrade";
+			proxy_set_header Connection "Upgrade";
 		}
 		location /shell/ {
 			valid_referers server_names;
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,4 +1,17 @@
 [[package]]
+name = "bidict"
+version = "0.22.1"
+description = "The bidirectional mapping library for Python."
+category = "main"
+optional = false
+python-versions = ">=3.7"
+
+[package.extras]
+docs = ["furo", "sphinx", "sphinx-copybutton"]
+lint = ["pre-commit"]
+test = ["hypothesis", "pytest", "pytest-benchmark", "pytest-cov", "pytest-xdist", "sortedcollections", "sortedcontainers", "sphinx"]
+
+[[package]]
 name = "certifi"
 version = "2023.7.22"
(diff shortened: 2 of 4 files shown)


phillxnet:
# Testing
An rpm built with this proposed change pre-merged but without the associated (linked) update in the js client in rockstor-jslibs we see, via browser dev tools network (Disable Cache selected), the expected:

**"The client is using an unsupported version of the Socket.IO or Engine.IO protocols"**

With example 404 URLs listed as follows:

> https://rleap15-5.lan/socket.io/?EIO=3&transport=polling&t=Oip1gWz

Which indicate that our prior client is attempting to use version 3 of the protocol (?EIO=3). We expect to see ?EIO=4 once our client js is updated in testing branch of rockstor-jslibs.

Consequently we have much depleted header info on all pages and the Dashboard page has the following non-functional widgets:
- Disk Activity
- CPU
- Memory
- Network

All of which depend on websockets to have their info streamed to them.

phillxnet:
# Testing post jslibs fix

Post having the newest jslib socket.io client in place in rockstor-jslibs testing branch we now have full function of our prior detailed widgets and websocket streamed header info (kernel version uptime etc).

A further tests was to ensure our desktop maintain activity during a Web-UI initiated system update (upstream updates). In some development incarnations we had blocking behaviour where the dashboard would freeze during this event.

phillxnet:
@FroggyFlox & @Hooverdan96 I'll go ahead and pop this one in to ease our other ongoing concerns re updating dependencies. All looks to be behaving as before and we have simpler dependencies. With a key reduction here being the removal of our dependency on gevent-websocket. We have some refinement to do as-yet re allowed clients etc but we can address those as we morph our underlying technologies to using more django native ones. But that, in turn, depends on the following Milestone: https://github.com/rockstor/rockstor-core/milestone/24

PR #2705, opened 2023-10-15
Update Requests library to latest #2704

### LINKED ISSUE #2704: Update Requests library
Our prior pined version of 'requests' can now be re-addressed re testing branch development concerning our base Python version, i.e. we now use Py3.9:

> requests = "==2.27.1"  # Last Python 2.7/3.6 version

Re-examine our options on updating this pinning or remove our pinning all-together. Note that there may well be additional parameters expected such as timeout etc. These should be addressed as part of this issue also.

A potential enhancement of this update could be to enable server side compression compatibility that it is thought was introduced in an interim version of 'requests' post what were have pinned currently. This could help with the performance of such things as our Rock-on update mechanism.

phillxnet:
Remove our prior pinning for 'requests' as no longer required.
## Includes:
- Adding 10 second timeout on update repo status check, to avoid indefinite hang potential.
- Re Black formatting on changed pkg_mgmt.py file.

Fixes #2704

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -32,14 +32,11 @@ pycparser = "*"
 
 [[package]]
 name = "charset-normalizer"
-version = "2.0.12"
+version = "3.3.0"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
 optional = false
-python-versions = ">=3.5.0"
-
-[package.extras]
-unicode_backport = ["unicodedata2"]
+python-versions = ">=3.7.0"
 
 [[package]]
 name = "dbus-python"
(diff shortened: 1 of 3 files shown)


phillxnet:
## Testing
An RPM based install (Leap 15.5 amd64 build host) could successfully:
- Subscribe to stable, move to testing, then move back to stable and back to testing without issue.
- Pull in the newest list of Rock-ons, with the User-Agent of 'python-requests/2.31.0' confirmed in the rock-ons server logs.

phillxnet:
Popping this one in as per our current trend re addressing our updates backlog (technical dept). And to assist with current milestone goals.

PR #2706, opened 2023-10-16
Update gunicorn to latest - use gthread - discrete config file #2702

### LINKED ISSUE #2702: Update gunicorn to latest - use gthread - discrete config file
We have a legacy gunicorn version held back by a prior Python version. Review our options re updating this python wsgi server.

An additional aim, in line with our improved Python base, is to move from sync threading, to the more appropriate gthread gunicorn worker type.

See: https://docs.gunicorn.org/en/latest/settings.html#worker-class
And: https://docs.gunicorn.org/en/stable/design.html#choosing-a-worker-type

> The default synchronous workers assume that your application is resource-bound in terms of CPU and network bandwidth. Generally this means that your application shouldn’t do anything that takes an undefined amount of time. An example of something that takes an undefined amount of time is a request to the internet.

https://docs.gunicorn.org/en/stable/design.html#gthread-workers

> The worker gthread is a threaded worker. It accepts connections in the main loop. Accepted connections are added to the thread pool as a connection job. On keepalive connections are put back in the loop waiting for an event. If no event happens after the keepalive timeout, the connection is closed.

phillxnet:
## Includes
- Unpin gunicorn version to use latest.
- Establish discrete gunicorn.conf.py config file.
- Move from sync to gthread gunicorn worker for main app.
- Remove defunct init.d file for gunicorn
- Add unrelated .gitignore entry re docker-daemon.json.

Fixes #2702


phillxnet:
@Hooverdan96 I know what we are putting in here has much reduced timeouts to what we had: but I'm pretty sure, with the very different nature of gthread workers in comparison to our prior inappropriate use of sync workers, we are moving in the right direction. So I'l like to return these timeouts to their default of 30 seconds to re-assess what we actually need here. Plus the prior sync worker type was no sensitive to on-going activity; where-as gthread workers are, so they will not be timed out if they are still dealing with an ongoing communication.

Note also, with the custom log additions here, we should now have far better timing info with which to debug such things as our available rock-on update mechanism. I intend also to use this new log entry to see if server side compression can help us out on that one.

I have also now removed the distracting gunicorn file that should have been deleted long ago. My apologies for leaving that in place all this time.

@FroggyFlox I'll pop this one in also to enable your ongoing development to be re-based prior to submission.

Hooverdan96:
@phillxnet I think that's good to return the lower numbers. With the changes you mentioned I would hope that it becomes less of a potential bottleneck, and once @FroggyFlox is able to take another look at how we do the refresh/downloads, there might be possibly a more effective way of doing delta updates ...

phillxnet:
## Testing
An RPM was successfully built and installed with this match pre-merged. The resulting instance was able to:
- Initiate via the Web-UI setup.
- Import a pool.
- Configure Rock-ons system.
- Install and run the netdata Rock-on.
- Peform a re-raid (balance).
- Perform a scrub.

So we look to have persisted a working thread arrangement having moved to the very different gthread  - at least for the main app. And we now have a Py3 only gunicorn.

PR #2711, opened 2023-10-18
remove django-braces dependency #2709

### LINKED ISSUE #2709: re-asses django-braces requirement
Django-braces "django-braces 1.13.0 Reusable, generic mixins for Django" may be an inclusion from a prior experiment. This issue is to assess if we actually use this library any longer, and if so for what. The project itself is far from regularly maintained but does seem to be just a set of convinience classes. And as such we likely want to drop this dependency and define just the helper classes that we actually/use need.

https://django-braces.readthedocs.io/en/latest/index.html

https://github.com/brack3t/django-braces/#notes:
django-braces is stable and time-tested. It does not receive a lot of updates and is not in active development.

Also note that the linked changelog does not include PyPi releases:
https://django-braces.readthedocs.io/en/latest/changelog.html

https://pypi.org/project/django-braces/

It is proposed that we drop this dependency as it may well hinder our ability to update freely in the near future.

phillxnet:
Remove now defunct django-braces dependency. Likely left over from a prior requirement.

## Includes
- Incidental/unrelated update to urllib3 2.0.6 to 2.0.7.

Fixes #2709

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -71,14 +71,6 @@ argon2 = ["argon2-cffi (>=16.1.0)"]
 bcrypt = ["bcrypt"]
 
 [[package]]
-name = "django-braces"
-version = "1.13.0"
-description = "Reusable, generic mixins for Django"
-category = "main"
-optional = false
-python-versions = "*"
-
-[[package]]
 name = "django-oauth-toolkit"
 version = "1.1.2"
 description = "OAuth2 Provider for Django"
@@ -360,7 +352,7 @@ testing = ["pytest", "pytest-cov"]
 
 [[package]]
 name = "urllib3"
-version = "2.0.6"
+version = "2.0.7"
 description = "HTTP library with thread-safe connection pooling, file post, and more."
 category = "main"
 optional = false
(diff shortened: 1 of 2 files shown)


phillxnet:
## Testing
Actually unrelated to PR as there is no known use of this library. Noting just as a quick current-state of capability.

CLI test run:
```
lbuildvm:~ # cd /opt/rockstor/src/rockstor/
lbuildvm:/opt/rockstor/src/rockstor # export DJANGO_SETTINGS_MODULE=settings
lbuildvm:/opt/rockstor/src/rockstor # poetry run django-admin test -v 2
...
Ran 262 tests in 26.129s

OK

```
An RPM was successfully built and installed with this match pre-merged. The resulting instance was able to:

- Initiate via the Web-UI setup.
- Install all pending upstream base OS update.
> 127.0.0.1 - - [18/Oct/2023:18:18:46 +0100] "POST /api/commands/update HTTP/1.1" 200 6 "-" "python-requests/2.31.0" 120145ms
- Import a pool.
- Configure Rock-ons system.
- Install and run the netdata Rock-on.
- Stop and remove Rock-on.
- Stop Rock-ons service.
- Collect and present in Web-UI a drives SMART data.
- Create share.
- Export share via SMB.
- Remove export.
- Remove share.
- Turn off SMB servcie.
- Create scheduled task to snap.
- Remove scheduled task.
- Remove snapshots.
- add new user.
- remove new user.

phillxnet:
@FroggyFlox & @Hooverdan96 I'll pop this in as I'm pretty sure, bar the incidental urllib3 minor update, we have no actual change here. It seems this was a ghost import that I failed to appreciate until our current run on dependency re-examination/update-run!

Issue #2332, opened 2021-12-13
Ability to reset services to original status/settings


Hooverdan96:
In reference to this conversation: https://forum.rockstor.com/t/config-restore-tests/8059/5?u=hooverdan

Ability to reset a given service to its default values. There are use cases, where a user configures a service with required parameters and then realizes that the additional parameters might not be needed, or the service overall is not needed. Just removing required field entries will not remedy the situation, so it would be good to possibly have a "reset to default" function to ensure that a given service is restored to its original state (and then could be turned off if so desired, but would still have a consistent underlying configuration).

FroggyFlox:
Linking to a forum thread for feedback on how to implement that:
[https://forum.rockstor.com/t/reset-service-configuration-to-default/9063](https://forum.rockstor.com/t/reset-service-configuration-to-default/9063?u=flox)

FroggyFlox:
I've started thinking about that issue and tested a few things. I mainly see 2 options to do that so far.

## Option A: the `PUT` route
Given we are modifying an existing model instance, my first thought was to use a `PUT` call to `api/sm/services/<service_name>` and modify `BaseServiceDetailView` to stop the service and set its `config` field to its default.
This works well for almost all configurable services, except for the following:
- shellinaboxd: has a set default: we can still easily change the new `config` value to that default instead of `None`. No problem really
- tailscaled: we need to also run `systemctl stop tailscaled && systemctl disable tailscaled`. No problem either.
- rockstor: that one is tricky for 2 reasons:
    - stopping the service: here we would need to add an exception as this service cannot be stopped. `RockstorServiceView` doesn't even include a `stop/start` command as it cannot be toggled ON/OFF from the webUI.
    - changing its config requires a `POST` to `api/sm/services/rocsktor/config` with the value of the default config (not `None`).
If we were to use `PUT`, we would thus need to add an exception to send the `POST` to the `config` command with a payload mimicking the default service config instead of sending a `stop` command and resetting the model instance config field to `None`.

The advantage of this option:
- more appropriate for the modification of an existing model instance
- a single `def put()` in `BaseServiceDetailView` will be able to deal with resetting all services
- no need to code a `if command == "reset"` for each `<service_name>ServiceView`

Disadvantages:
- need some customization, and rather an extensive one for the Rockstor service. This can be ugly and rather hacky?

## Option B: the `POST` route
[...]

PR #2712, opened 2023-10-18
Implement Tailscale service #2679

### LINKED ISSUE #2679: Add Tailscale integration
@phillxnet, @Hooverdan96, this is a proposal transfer from our previous discussion at https://github.com/rockstor/rockon-registry/pull/338.

From that previous conversation, it seems the agreement was to integrate Tailscale into Rockstor at the "system" level. I'm thus opening this issue accordingly to discuss **how** best to do that.

# Requirements
Going with this route, we would need to install the `tailscale` package, which provides a (single?) binary `tailscale` that represents the interface to manage everything. This, in turn, creates a systemd service (`tailscaled`).
To install this package, we have a few options, but it seems the choice will be very limited. See the official instructions for install on Linux at https://tailscale.com/kb/1031/install-linux/.

They have a curl-piped-to-bash option, but that essentially does the "manual" option below.

The *manual* option is thus to:
- add the tailscale repos
- install the `tailscale` package
- enable the `tailscaled` systemd service
- "activate" tailscale (`tailscale up`)

The important point here is the **repositories**. We can go with:
- official repos hosted by Tailscale (`https://pkgs.tailscale.com/stable/opensuse/`), available for Leap 15.{1,2,3,4,5} and TW.
- OBS: https://build.opensuse.org/package/show/network:vpn/tailscale

The official repos always have the latest release available.
The OBS repo unfortunately is a bit out-of-date (a couple minor releases behind) and fails to build for Leap and TW at the moment (issue with GO build dependency).
Given we need add a repository either way, it seems using the official repos is the best choice.

# Results
## File system level
Using the official install instructions lead to the following new repository...
```
buildvm155:~ # zypper repos --uri tailscale-stable
Alias          : tailscale-stable
Name           : Tailscale stable
URI            : https://pkgs.tailscale.com/stable/opensuse/leap/15.5/x86_64
Enabled        : Yes
GPG Check      : (r ) Yes
Priority       : 99 (default priority)
Autorefresh    : Off
Keep Packages  : Off
Type           : rpm-md
GPG Key URI    : https://pkgs.tailscale.com/stable/opensuse/leap/15.5/repo.gpg
Path Prefix    : 
Parent Service : 
Keywords       : ---
Repo Info Path : /etc/zypp/repos.d/tailscale-stable.repo
MD Cache Path  : /var/cache/zypp/raw/tailscale-stable
```
... defined with:
```
buildvm155:~ # cat /etc/zypp/repos.d/tailscale-stable.repo
[tailscale-stable]
name=Tailscale stable
enabled=1
autorefresh=0
baseurl=https://pkgs.tailscale.com/stable/opensuse/leap/15.5/$basearch
gpgcheck=1
gpgkey=https://pkgs.tailscale.com/stable/opensuse/leap/15.5/repo.gpg
```
Note the `$basearch` bit above... **whether an aarch64 package exist remains to be verified!**

The RPM installs the following files:
```
buildvm155:~ # rpm -ql tailscale
/etc/default/tailscaled
/lib/systemd/system/tailscaled.service
/usr/bin/tailscale
/usr/sbin/tailscaled
/var/cache/tailscale
```

The resulting systemd unit is as follows:
```
buildvm155:~ # systemctl cat tailscaled.service 
# /lib/systemd/system/tailscaled.service
[Unit]
Description=Tailscale node agent
Documentation=https://tailscale.com/kb/
Wants=network-pre.target
After=network-pre.target NetworkManager.service systemd-resolved.service

[Service]
EnvironmentFile=/etc/default/tailscaled
ExecStartPre=/usr/sbin/tailscaled --cleanup
ExecStart=/usr/sbin/tailscaled --state=/var/lib/tailscale/tailscaled.state --socket=/run/tailscale/tailscaled.sock --port=${PORT} $FLAGS
ExecStopPost=/usr/sbin/tailscaled --cleanup

Restart=on-failure

RuntimeDirectory=tailscale
RuntimeDirectoryMode=0755
StateDirectory=tailscale
StateDirectoryMode=0700
CacheDirectory=tailscale
CacheDirectoryMode=0750
Type=notify

[Install]
WantedBy=multi-user.target
```

## Activating Tailscale
Starting the systemd service leads to the following status:
```
buildvm155:~ # systemctl status tailscaled
● tailscaled.service - Tailscale node agent
     Loaded: loaded (/lib/systemd/system/tailscaled.service; enabled; vendor preset: disabled)
     Active: active (running) since Thu 2023-09-21 11:41:23 EDT; 1s ago
(...)
     Status: "Stopped; run 'tailscale up' to log in"
(...)
```

We thus need to login.

## Login/authentication
I see 3 main options here:
- cli
- web portal
- authkey

### CLI
Manually run `tailscale up` prints a URL to visit and authenticate the device to the user. Simple and robust, but not very practical from Rockstor's perpsective.

### Web portal
The `tailscale` command has a `web` subcommand that creates a simple web page to login, made for solutions like Rockstor actually. It was recently made more flexible (https://github.com/tailscale/tailscale/commit/9a3bc9049cfda551774f62cc7b63415d5a2e71a8) but that last addition has not made it to the RPM release yet, it seems. Note it also remains to be verified whether Rockstor actually needs that last addition.

This could be done as follows:
```
buildvm155:~ # tailscale web --listen $(hostname):8088
2023/09/21 11:49:25 web server running on: http://buildvm155:8088
```
Visiting `http://buildvm155:8088` brings the following:
image

After clicking on that "Login" button, the machine is added to the user's tailnet and we can see that the activation is complete:
```
buildvm155:~ # tailscale web --listen $(hostname):8088
2023/09/21 11:49:25 web server running on: http://buildvm155:8088
2023/09/21 11:59:34 Doing edit: MaskedPrefs{WantRunning=true AdvertiseRoutes=[]}
2023/09/21 11:59:34 tailscaleUp(reauth=true, logout=false) ...
2023/09/21 11:59:35 tailscaleUp = (URL true, <nil>)
```
Note that this `tailscale web` process is still live! We probably would need to find a way to kill it once we no longer need it.

### Authkey
Documentation: https://tailscale.com/kb/1085/auth-keys/
This would require us to ask the user for that key. It would be best not to store this in the database so useful only if we can pass it directly from the UI... I'm unsure whether this would be less secure, though. It also require another configuration step by the user (from Tailscale's website) so it's probably not the best option. It could be worth implementing for users who want that option, however... maybe in a later stage?

## State persistence
What happens after that initial activation?

### Reboot
Given the `tailscaled` service is enabled by the install instructions, it is back up and running after a reboot, and the machine is joined to the tailnet:
```
buildvm155:~ # tailscale status
xxx.xxx.xxx.xxx buildvm155           xxxxx@      linux   -
```

To turn it OFF, we can run `tailscale down`:
```
buildvm155:~ # tailscale down

buildvm155:~ # tailscale status
Tailscale is stopped.
```
Notably, this is disconnected from the `tailscaled` service as if we reboot then, the `tailscaled.service` is back active (as it was enabled), but tailscale is still down:
```
buildvm155:~ # systemctl is-active tailscaled
active

buildvm155:~ # tailscale status
Tailscale is stopped.
```

### Tailscaled and network connections
The `tailscaled` service actually "manages" the underlying connection.
Indeed, when turning it ON, we now see a new connection:
```
buildvm155:~ # systemctl start tailscaled
buildvm155:~ # nmcli
eth0: connected to eth0
(...)

lo: unmanaged
(...)

tailscale0: unmanaged
        "tailscale0"
        tun, sw, mtu 1280

DNS configuration:
(...)
```

Stopping the `tailscaled` service and it is now gone:
```
buildvm155:~ # systemctl stop tailscaled
buildvm155:~ # nmcli
eth0: connected to eth0
(...)

lo: unmanaged
(...)

DNS configuration:
(...)
```

# Overall implementation
My first thought would be to add Tailscale as one of our supported services and would thus be listed as its own on the "System" > "Services" page. We still need to decide what happens when we **configure** that service, and when we **toggle it ON/OFF**.

Its configuration window could (at first) simply include a "login" button that would simply run the `tailscale web ...` command and open a new browser page at the URL that was created by that command. Here, note that if the `tailscale web ...` command is run again on a machine already joined to the user's tailnet, it doesn't hurt and simply shows a "confirmation" window:
image
We can thus keep showing that "login" button unconditionally (whether the machine is joined or not).

Later, we can think of adding configuration to other Tailscale features such as Taildrop or Tailscale SSH.

Toggling the service ON/OFF could simply run `tailscale up`/`tailscale down`.

Note that by default, each machine "expires" on a user's tailnet, which means that each user will need to re-authenticate after some time: https://tailscale.com/kb/1028/key-expiry/. All of this is configurable on Tailscale's website, but this means that we would need to catch this "expired" state to show a nice info windows prompting the user to press again the "Login" button in the Service's configuration window.

## What about the tailscale package?
We also need to figure out how/when to add the tailscale RPM repo and install the package.
Should we check for this need and do that when the user clicks on the "login" button in the Service's "configuration " window? I'm still a bit unsure about that. It does seem like adding the repo and the package in the kiwi recipe would be a bit much, however, so we may just do this repo + package install "post Rockstor installation", no?
Also, should we undo this when the user turns the service OFF? This would also require to do the repo and package install when the user turns it back ON, though. It would be nice to offer a way for the user to avoid carrying an extra repository and package if they either don't use the Tailscale service, or no-longer use it.

Hope this helps begin the planning phase for this... I'm personally eager to see that implemented once we figure out how.

FroggyFlox:
Fixes #2679

This pull-request (PR) proposes to add a new Service (in System > Services) for Tailscale, allowing the users to connect their Rockstor machine to their existing Tailscale account entirely from Rockstor's UI, without command line interaction. See related issue for rationale.

# Overall implementation
## Introductory notes
**This PR makes the assumption that the `tailscale` package is already installed on the system**, providing the `tailscaled` daemon systemd service as well as the `tailscale` cli interface (on which this PR entirely relies). Ensuring the tailscale package is installed should thus be the subject of a dedicated issue in our rockstor-installer repository.

To better understand the implementation in this PR, it is important to note that there are two main components to Tailscale on the system:
- the `tailscaled` daemon (systemd service): creates the `tun` device and "routes" traffic. This service being active is not sufficient for having an active Tailscale connection. This is done by the `tailscale` binary (see below).
- the `tailscale` cli binary: represents the interface between the user and its Tailscale account and connection configuration. This binary **requires** the `tailscaled` daemon to up and running to do anything.

@phillxnet, Tailscale licensing is detailed at: https://github.com/tailscale/tailscale/blob/main/LICENSE
Their website (https://tailscale.com/opensource/) details:
> The client, which runs on each of a user’s devices, is mostly open source. The core client code for the Tailscale daemon used across all platforms is open source, and the full client code is open source for platforms that are also open source.
> We enthusiastically support open source operating systems such as Linux and Android, and we believe that our time and energy is best directed toward the communities for these platforms. As a result, where the operating system is open source, the daemon and GUI are open source; and where the operating system is closed, the daemon is open source and the GUI is closed source.
> This means that you can build the Linux and Android clients yourself, and you can build the Windows and macOS clients without the GUI.

## Rockstor's implementation
1. The Tailscale service is configurable. The user can thus configure a various set of parameters using Rockstor's UI. *By default, it is NOT configured*, and the `tailscaled` service is NOT running.
2. When configured, we assume the user intends on using the Tailscale service and thus start the `tailscaled` daemon. This creates the `tun` device but does not activate the Tailscale connection. This step is a requirement for the initiation of the login/authentication mechanism.
3. In this state (configured, but Service OFF), we detect if the user needs to authenticate this machine to their Tailscale account. If it is the case, we display a "Login" button *in lieu* of the usual ON/OFF toggle.
4. When clicking the "Login" button, we trigger and shortly thereafter timeout a `tailscale up` call; this generates an authURL that we then retrieve and use to open a new browser window that the user can thus use to complete the process of authentication to their Tailscale account. When completed, the Tailscale service is ON and the machine is fully connected to their Tailscale network.
5. Toggling the Tailscale service OFF disonnects the machine from their Tailscale network. It does *not* turn OFF the `tailscaled` daemon as this PR assumes the user might still want to turn it back ON (shortly) thereafter. I'm thinking of using https://github.com/rockstor/rockstor-core/issues/2332 here: when resetting the service to its default, we consider that the user does not want Tailscale anymore and we would thus stop the `tailscaled` daemon then.

# Implementation details
## Tailscale service configuration
As detailed in https://github.com/rockstor/rockstor-core/issues/2679#issuecomment-1741838236, `tailscale` supports many options, and this PR proposes to provide a simple form to set what might be the most commonly used ones.
For all other options, a *custom configuration* box is offered so that the user can manually enter what they want. **Note** that if the user enters in this custom configuration box a parameter that is already supported in the form above it, this PR considers what was defined in the form as taking precedence over the custom configuration and thus ignores it.
The configuration form includes a variety of Javascript-based validation to prevent the user to enter incompatible or invalid settings. Note that this covers the incompatibilities that I have personally witnessed... there may be more that I have not seen yet.

Just like all other services, submitting the configuration form sends that data to the new `TailscaledView` that parses it and validates it as follows:
- if `hostname` param is present: verify it is alphanumeric and does not include special characters. Underscores are also invalid so we change them to hyphens (same as what the Tailscale webUI does).
- parses the custom configuration and remove parameter if it is a duplicate of what's already in the webUI form.
After this validation, we save the config in the database.

This also *enables* and *starts* the `tailscaled` systemd service, required for the next steps.

See below for a screenshot of the configuration modal:
image

## Tailscale service status
Unlike other Services, Tailscale needs to be authenticated to function. We thus consider the Tailscale service as **ON** if the machine is connected to the user's Tailscale network. This, however, requires authentication, explained in the next section.

For Tailscale, however, we need to have a more granular knowledge of Tailscale status than for other services (due to the reasons above). Fortunately, `tailscale` includes a subcommand (`status`) that can report a very detailed report in JSON. This includes one important key: `BackendState`. This key indeed is enough to give us the information we need:
- NeedsLogin
- Stopped
- Running

We thus have 3 states. On top of that, we also have the state of "tailscaled systemd service not active" -> 4 states. In reality, however, we just need to know the following:
- is the `tailscaled` daemon running?
- if the `tailscaled` daemon is running:
  - does `tailscale` need to be authenticated (BackendState: "NeedsLogin")? If yes: show "Login" button (see below).
  - is tailscale running?
  - is tailscale stopped?

In `services.js`, we currently use return codes as returned by systemd/supervisorctl to represent services status to display the ON/OFF toggle accordingly:
- for return code 0: running
- for return code > 0: not running

Note that systemd already returns: 0 = unit is active, 3 = unit is not active, 4 = no such unit (@phillxnet's notes).
Here, we need a new state: needs login. I thus chose to attribute a return code of 5 to represent that. This is detailed in `system.services.tailscale_service_status()`. As a result, if service status is 5, we need to display the "Login" button.

## Tailscale authentication/login
Running `tailscale up` at this point would hang, waiting for the user to visit an automatically generated URL to login/authenticate to their Tailscale account. As a result, this would make the front-end hang as well. I have thus ended up relying on the following "feature": when `tailscale up` is triggered, it checks if authentication is needed and generates an authURL, waiting for the user to copy this URL, paste it in the browser and then complete the authentication process. When completed, `tailscale up` detects that and finishes activating the tailscale connection. Now, the "detail" that this PR uses is that when that initial `tailscale up` call that generates the needed authURL finishes (successfully or not), the generated authURL can be retrieved from `tailscale status`, and is still valid for the user to complete the authentication process. Rockstor thus uses that as follows:
- when authentication is required, we display a "Login" button *in lieu* of the ON/OFF toggle (design choice to be discussed if needed).
- clicking this button calls a new endpoint with the `login` **action** (new definition in urls) that calls `tailscale up` with a timeout of 2 sec. That is enough to generate the authURL, and then completes. We can then call `tailscale status` to retrieve this authURL and save it in the config model entry. We then refresh the Backbone collection for the services to have that authURL available to our front-end and use that to open a new window with that URL. That process is contained in the `tailscaleAuthAction` function in `services.js`.
- once the authentication process is completed by the user, the tailscale service is automatically turned on (done by tailscale itself, not Rockstor).
- the latter is picked up by our automatic websocket-based service status update mechanism (polling interval: 15 sec) that will then show the Tailscale service as ON. This explains why https://github.com/rockstor/rockstor-core/pull/2682 was needed here.

# Machine expiration
By default, each machine newly connected to a Tailscale network expires in 30 days.
I've tried to mimick this as follows:
1. "remove" the machine from Tailscale's dashboard. Unfortunately, this does not stop the tailscale connection on the machine as it still shows a status of "Running".
2. in Rockstor, manually toggle the Tailscale service OFF. `tailscale status` shows it as Stopped, but not needing login, unfortunately, so we have the ON/OFF toggle still in Rockstor's UI.
3. toggling the service ON, though, triggers `tailscale up`, which then detects it needs login. In Rockstor UI, this is detected and we show the "Login" button. One can then complete the login process.

It is not as smooth as I would like, but I'm unsure how to deal with this "weird" state given the Needs login state is not detected by Tailscale before we need to know it. Maybe a "normal" machine expiration is reflected differently than manually removing it from the dashboard?

The `tailscale up` command does have a `--force-reauth` flag that will force a re-authentication. Maybe this should be included in the configuration modal as a button in case of need.

# Testing
All development and testing was conducted on Leap 15.5; all tests are passing (note that we now have 17 more tests included in this PR):
```
.......................................................................................................................................................................................................................................................................................
Ran 279 tests in 15.877s

OK
```

# Miscellaneous
## Services table
I had to make a few re-arrangements in the table listing all services:
- increase default display of services to 30.
- remove the `text-align: left` css that was forcing a left alignment for the buttons.
- reduce width of toggle/login buttons
The resulting table does have a lot of empty space in the center, though... This is another design choice to be discussed as/if desired.

## Change to `run_command()`
**This PR does include a change to `system.osi.run_command()**.
This change was required to easily load the JSON output from `tailscale status --json` as a dict. I've slightly edited `run_command()` to allow returning the "raw" output from `p.communicate()`, while leaving the default behavior the same as before.

## Help button link
The link included in the help button at the top of the configuration modal points to the docs' root. This should be changed to point to a dedicated description of the Tailscale service once it exists.

## Change to .gitignore
We previously had `static/` in `.gitignore` to ignore the static dir created by `collectstatic`. This, however, was not selective enough and was causing `static/` subfolders to be ignored. That included `configure_tailscaled.jst`, for instance. The change in this PR should concern only the `static/` folder located in the project's root: `/static/`.

## How the new tailscale0 tun device shows in the Network page
`tailscaled` systemd service is inactive:
image

`tailscaled` systemd service is active, but tailscale is OFF:
image

Tailscale service ON:
image

### DIFF
--- a/src/rockstor/system/tailscale.py
+++ b/src/rockstor/system/tailscale.py
@@ -1,0 +1,204 @@
+"""
+Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
+This file is part of RockStor.
+
+RockStor is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published
+by the Free Software Foundation; either version 2 of the License,
+or (at your option) any later version.
+
+RockStor is distributed in the hope that it will be useful, but
+WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <https://www.gnu.org/licenses/>.
+"""
+import logging
+import re
+import time
+from typing import List
+
+from storageadmin.exceptions import RockStorAPIException
+from system.constants import TAILSCALE
+from system.network import enable_ip_forwarding, disable_ip_forwarding
+
+from system.osi import run_command
+from system.services import get_tailscale_status
+
+# Config options supported in webUI
+# (used to filter out eventual duplicates in ts_custom_config)
+TS_UI_SETTINGS = [
+    "accept_routes",
+    "advertise_exit_node",
+    "advertise_routes",
+    "exit_node",
+    "exit_node_allow_lan_access",
+    "hostname",
+    "reset",
+    "ssh",
+]
+logger = logging.getLogger(__name__)
+
+
+def tailscale_up(config: dict = None, timeout: int = None):
+    """Start tailscale
+    Wrapper around `tailscale up`. It first gets args from config (if any)
+    and constructs the tailscale up command accordingly.
+    Adds the timeout flag is specified (used to generate authURL during the
+    LOGIN action from the webUI).
+    """
+    # Construct tailscale up command
+    cmd = [
+        TAILSCALE,
+        "up",
+    ]
+    if config is not None:
+        cmd = construct_tailscale_up_command(config)
+        # Enable IP forwarding if needed (subnet router or exit node enabled)
+        if any(key in config for key in ("advertise_exit_node", "advertise_routes")):
+            enable_ip_forwarding(name="tailscale", priority=99)
+    # Add short timeout to just trigger the creation of authURL
+    if timeout:
+        cmd.append(f"--timeout={timeout}s")
+    return run_command(cmd, log=True, throw=False)
+
+
+def tailscale_down(config: dict = None):
+    """Stop tailscale
+    Wrapper around `tailscale down`.
+    """
+    cmd = [
+        TAILSCALE,
+        "down",
+    ]
+    o, e, rc = run_command(cmd, log=True, throw=False)
+    if config is not None:
+        if any(key in config for key in ("advertise_exit_node", "advertise_routes")):
+            disable_ip_forwarding(name="tailscale")
+    return o, e, rc
+
+
+def extract_param(param: str) -> str:
+    """Get and format the name of a custom parameter
+    Isolates the name of the parameter from its cli format and
+    returns it after brief formatting.
+
+    There are 2 types of params:
+      - in the form --<key>=<value>
+      - in the form --<key>
+
+    These would return, for instance:
+      - "--accept-risk=all" -> "accept_risk"
+      - "--shields-up" -> "shields_up"
+    """
+    if re.search("=", param) is not None:
+        param = param.split("=")[0]
+    param = param.split("--")[1]
+
+    # Tailscale cli params use "-" whereas Rockstor uses "_" for storage
+    param = re.sub("-", "_", param)
+
+    return param
+
+
+def validate_ts_hostname(config: dict) -> dict:
+    """Ensure hostname is alphanumeric with hyphens only
+    No special character (including hyphens) is allowed as a custom hostname.
+
+    "hostname": "rock-dev_@#~!$%^&*()+123"
+    should return
+    "hostname": "rock-dev-123"
+    """
+    if "hostname" in config:
+        config["hostname"] = re.sub("_", "-", config["hostname"])
+        to_exclude = [
+            "@",
+            "#",
+            "~",
+            "!",
+            "$",
+            "%",
+            "^",
+            "&",
+            "*",
+            "(",
+            ")",
+            "+",
+        ]
+        config["hostname"] = "".join(
+            c for c in config["hostname"] if not c in to_exclude
+        )
+    return config
+
+
+def validate_ts_custom_config(custom_config: List) -> List:
+    """Validate Tailscale service custom config
+    Check for the existence of duplicates in the custom config section
+    and filter them out if any.
+
+    :param custom_config: list of parsed custom config
+    :return: list of filtered custom config
+    """
+    # Keep only param in the form '--<param>'
+    custom_config = [x for x in custom_config if re.match("--", x)]
+    # Keep only settings not already in UI modal
+    filtered_conf = [
+        setting
+        for setting in custom_config
+        if extract_param(setting) not in TS_UI_SETTINGS
+    ]
+    return filtered_conf
+
+
+def construct_tailscale_up_command(config: dict) -> List:
+    """Construct the tailscale up command from service config"""
+    yes_flags = ["yes", "true"]
+    cmd = [
+        TAILSCALE,
+        "up",
+    ]
+    for param in config:
+        if param == "auth_url":
+            # do not add auth_url to cmd as it is not valid
+            continue
+        elif param == "custom_config":
+            for flag in config[param]:
+                cmd.append(flag)
+        elif config[param] in yes_flags:
+            cmd.append("--" + re.sub("_", "-", param))
+        else:
+            formatted_key = re.sub("_", "-", param)
+            cmd.append(f"--{formatted_key}={config[param]}")
+    return cmd
+
+
+def get_ts_auth_url() -> str:
+    """Get AuthURL from tailscale status output
+    Query the json output from `tailscale status --json`
+    to get the generated AuthURL. Repeat every sec for 10 sec or raise
+    a RockstorAPIException.
+    """
+    cur_wait = 0
+    while True:
+        logger.debug(f"Attempt number {cur_wait} to get AUTH_URL")
+        ts_status = get_tailscale_status()
+        auth_url = ts_status["AuthURL"]
+        if len(auth_url) > 0:
+            break
+        if cur_wait > 10:
+            logger.error(
+                f"Waited too long ({cur_wait} seconds) for tailscale "
+                "to generate login URL... giving up."
+            )
+            # break
+            e_msg = (
+                f"Waited too long ({cur_wait} seconds) for tailscale to generate the login URL. "
+                "Please try again. If the problem persists, email support@rockstor.com "
+                "with this message, or inquire on our forum (https://forum.rockstor.com)."
+            )
+            raise RockStorAPIException(status_code=400, detail=e_msg)
+        time.sleep(1)
+        cur_wait += 1
+    return auth_url
--- a/.gitignore
+++ b/.gitignore
@@ -51,7 +51,7 @@ jslibs/
 poetry-install.txt
 rockstor-jslibs.tar.gz*
 rockstor-tasks-huey.db*
-static/
+/static/
 
 # Rockstor error tgz
 /src/rockstor/logs/error.tgz
--- a/src/rockstor/scripts/prep_db.py
+++ b/src/rockstor/scripts/prep_db.py
@@ -20,7 +20,7 @@ from smart_manager.models import Service
 from storageadmin.models import Setup
 
 
-def register_services():
+def register_services() -> None:
     services = {
         "NFS": "nfs",
         "Samba": "smb",
@@ -39,6 +39,7 @@ def register_services():
         "Bootstrap": "rockstor-bootstrap",
         "Shell In A Box": "shellinaboxd",
         "Rockstor": "rockstor",
+        "Tailscale": "tailscaled",
     }
 
     # N.B. all other services have null as their default config with service.
--- a/src/rockstor/smart_manager/data_collector.py
+++ b/src/rockstor/smart_manager/data_collector.py
@@ -886,18 +886,15 @@ class ServicesNamespace(RockstorIO):
     start = False
 
     def on_connect(self, sid, environ):
-
         self.emit("connected", {"key": "services:connected", "data": "connected"})
         self.start = True
         self.spawn(self.send_service_statuses, sid)
 
     def on_disconnect(self, sid):
-
         self.cleanup(sid)
         self.start = False
 
     def send_service_statuses(self):
-
         while self.start:
 
             data = {}
@@ -914,7 +911,6 @@ class ServicesNamespace(RockstorIO):
                 data[service.name] = {}
                 output, error, return_code = service_status(service.name, config=config)
                 data[service.name]["running"] = return_code
-
             self.emit("get_services", {"data": data, "key": "services:get_services"})
             gevent.sleep(15)
 
@@ -1122,7 +1118,6 @@ class SysinfoNamespace(RockstorIO):
                 data["status"] = "errors"
                 data["message"] = "Pools found with device errors: "
                 data["message"] += "({})".format(", ".join(labels))
-
             self.emit("pool_dev_stats", {"key": "sysinfo:pool_dev_stats", "data": data})
 
             gevent.sleep(30)
--- a/src/rockstor/smart_manager/fixtures/services.json
+++ b/src/rockstor/smart_manager/fixtures/services.json
@@ -1,128 +1,155 @@
 [
 {
-  "pk": 1, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "Replication", 
-    "name": "replication"
-  }
+    "model": "smart_manager.service",
+    "pk": 1,
+    "fields": {
+        "name": "replication",
+        "display_name": "Replication",
+        "config": null
+    }
 },
 {
-  "pk": 2, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "Samba", 
-    "name": "smb"
-  }
+    "model": "smart_manager.service",
+    "pk": 2,
+    "fields": {
+        "name": "smb",
+        "display_name": "Samba",
+        "config": "{\"workgroup\": \"WORKGROUP\"}"
+    }
 },
 {
-  "pk": 3, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "NFS", 
-    "name": "nfs"
-  }
+    "model": "smart_manager.service",
+    "pk": 3,
+    "fields": {
+        "name": "nfs",
+        "display_name": "NFS",
+        "config": null
+    }
 },
 {
-  "pk": 4, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "AD", 
-    "name": "winbind"
-  }
+    "model": "smart_manager.service",
+    "pk": 5,
+    "fields": {
+        "name": "ntpd",
+        "display_name": "NTP",
+        "config": null
+    }
 },
 {
-  "pk": 5, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "NTP", 
-    "name": "ntpd"
-  }
+    "model": "smart_manager.service",
+    "pk": 6,
+    "fields": {
+        "name": "nis",
+        "display_name": "NIS",
+        "config": null
+    }
 },
 {
-  "pk": 6, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "NIS", 
-    "name": "nis"
-  }
+    "model": "smart_manager.service",
+    "pk": 7,
+    "fields": {
+        "name": "ldap",
+        "display_name": "LDAP",
+        "config": null
+    }
 },
 {
-  "pk": 7, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "LDAP", 
-    "name": "ldap"
-  }
+    "model": "smart_manager.service",
+    "pk": 8,
+    "fields": {
+        "name": "sftp",
+        "display_name": "SFTP",
+        "config": null
+    }
 },
 {
-  "pk": 8, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "SFTP", 
-    "name": "sftp"
-  }
+    "model": "smart_manager.service",
+    "pk": 9,
+    "fields": {
+        "name": "rockstor",
+        "display_name": "Rockstor",
+        "config": null
+    }
 },
 {
-  "pk": 9, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "Data Collector", 
-    "name": "data-collector"
-  }
+    "model": "smart_manager.service",
+    "pk": 10,
+    "fields": {
+        "name": "smartd",
+        "display_name": "S.M.A.R.T",
+        "config": null
+    }
 },
 {
-  "pk": 10, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "Service Monitor", 
-    "name": "service-monitor"
-  }
+    "model": "smart_manager.service",
+    "pk": 11,
+    "fields": {
+        "name": "active-directory",
+        "display_name": "Active Directory",
+        "config": null
+    }
 },
 {
-  "pk": 11, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
-    "display_name": "AFP", 
-    "name": "netatalk"
-  }
+    "model": "smart_manager.service",
+    "pk": 12,
+    "fields": {
+        "name": "nut",
+        "display_name": "NUT-UPS",
+        "config": null
+    }
 },
 {
-  "pk": 12, 
-  "model": "smart_manager.service", 
-  "fields": {
-    "config": null, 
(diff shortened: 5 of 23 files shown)


src/rockstor/system/tailscale.py
```diff
@@ -0,0 +1,204 @@
...
+
+def validate_ts_hostname(config: dict) -> dict:
+    """Ensure hostname is alphanumeric with hyphens only
+    No special character (including hyphens) is allowed as a custom hostname.
+
+    "hostname": "rock-dev_@#~!$%^&*()+123"
+    should return
+    "hostname": "rock-dev-123"
+    """
+    if "hostname" in config:
```
Hooverdan96, line 114: quick question, can a hostname not have unicode characters in it?  If that's the case, wouldn't some "inclusive" regex be better than excluding what we currently know as "special" characters"?

Hooverdan96:
@FroggyFlox  - this is amazing!!! Looks simple enough :).  I had a question on the hostname validation that I placed as a comment in the pull request.

We will definitely need the documentation for that in the Rockstor docs.

src/rockstor/system/tailscale.py
FroggyFlox: Oh that is one very good question here. It's true that this validation is rather ugly at the moment...
To answer your question on unicode characters: I've tested using Tailscale's UI, which allows you to rename a hostname. This is a good way to test what is valid or not. I've pasted the following into the box: `München`, and it autocorrected to a hyphen:
image

It thus seems we can't use anything besides alphanumeric and `-`. I'll change to using a regex filtering/correction as you recommended.

phillxnet: @FroggyFlox @Hooverdan96 Re:

> I'll change to using a regex filtering/correction as you recommended.

Given this is a first draft, and a large one at that, I'm keen to get this out as-is. So this could be an enhancement by way of a follow-up. It likely won't be the last with such a significant addition.

Building now from what we have currently with a view to getting this out to the testing channel. We are getting along in this testing phase but not at the end yet so still time for some additions/refinements before we start heading for the next stable.

phillxnet (commented):
@FroggyFlox This is a superb effort/addition, I'm keen to get this out into testing for proper field tests as this is not a significant knowledge area for me.
@Hooverdan96's point was keen and we can follow that up as we progress in the current testing channel if need be.

Your additional test coverage here is most welcome. And thanks for the excellent tests on the new code presented in this PR.

Noting the following single test failure when Tailscale is not installed on the rpm build host:
```
FAIL: test_disable_ip_forwarding (rockstor.system.tests.test_system_network.SystemNetworkTests)
test proper call of os.remove() and final sysctl command
  File "/home/buildbot/worker/Poetry-Build-on-Leap15-5/rpmbuild/rockstor-core-5.0.4-2712/src/rockstor/system/tests/test_system_network.py", line 795, in test_disable_ip_forwarding
    self.mock_os_remove.assert_called_once_with("/etc/sysctl.d/99-tailscale.conf")
  File "/usr/lib64/python3.9/unittest/mock.py", line 918, in assert_called_once_with
    raise AssertionError(msg)
AssertionError: Expected 'remove' to be called once. Called 0 times.
Ran 279 tests in 26.910s
FAILED (failures=1)
```
We may want to assess if this is acceptable or even avoidable?

After installing the repo and package (Leap 15.5 only currently) via the instructions in our rockstor-installer issue comment: https://github.com/rockstor/rockstor-installer/issues/142#issuecomment-1770819322

We have all tests pass as per PR text.

Thanks for thinking to extend our services page: it would have been a shame for our ongoing page refresh bug to have returned us to page one and lost this entry.

If one submits the default config (accessed via spanner or auto shown via enable switch):
Tailscale-config-rpm-build

There is a:
>  Unknown internal error doing a POST to /api/sm/services/tailscaled/config 

```
/opt/rockstor/var/log/rockstor.log

system.exceptions.CommandException: Error running a command. cmd = /usr/bin/systemctl stop tailscaled. rc = 5. stdout = ['']. stderr = ['Failed to stop tailscaled.service: Unit tailscaled.service not loaded.', '']
```

Am I doing it wrong currently ?
I have installed the package via my indicated comment above. But that is all. Trying to simulate folks doing the same to see how we cope with that situation.

phillxnet:
So the full rockstor.log entry for what is likely an incomplete install of Tailscale on my part, is:
```
[19/Oct/2023 13:46:46] ERROR [system.osi:261] non-zero code(5) returned by command: ['/usr/bin/systemctl', 'stop', 'tailscaled']. output: [''] error: ['Failed to stop tailscaled.service: Unit tailscaled.service not loaded.', '']
[19/Oct/2023 13:46:46] ERROR [storageadmin.middleware:34] Exception occurred while processing a request. Path: /api/sm/services/tailscaled/config method: POST
[19/Oct/2023 13:46:46] ERROR [storageadmin.middleware:35] Error running a command. cmd = /usr/bin/systemctl stop tailscaled. rc = 5. stdout = ['']. stderr = ['Failed to stop tailscaled.service: Unit tailscaled.service not loaded.', '']
  File "/opt/rockstor/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/rockstor/.venv/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/opt/rockstor/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/rockstor/.venv/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/opt/rockstor/.venv/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/opt/rockstor/.venv/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/opt/rockstor/.venv/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/lib64/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/opt/rockstor/src/rockstor/smart_manager/views/tailscaled_service.py", line 98, in post
    systemctl(self.name, "stop")
  File "/opt/rockstor/src/rockstor/system/services.py", line 107, in systemctl
    return run_command(arg_list, log=True)
  File "/opt/rockstor/src/rockstor/system/osi.py", line 263, in run_command
system.exceptions.CommandException: Error running a command. cmd = /usr/bin/systemctl stop tailscaled. rc = 5. stdout = ['']. stderr = ['Failed to stop tailscaled.service: Unit tailscaled.service not loaded.', '']
```
I'll review my referenced Tailscale install comment.

phillxnet:
OK, some progress: In my test setup I have the following:
```
rleap15-5:~ # systemctl status tailscaled.service
Unit tailscaled.service could not be found.
```
So a partial install where the service has not been registered but installed !!

phillxnet:
OK, likely my fault here, I had assumed the failed test, which passed later on, was due to the package actually having been installed: but alas I had only added the repo !!! Strange how the indicated test failed initially (two builds, but later passed). Something likely interesting there.
I'll continue as currently build system has:
```
rleap15-5:~ # zypper info tailscale
...
Information for package tailscale:
----------------------------------
Repository     : Tailscale stable
Name           : tailscale
Version        : 1.50.1-1
Arch           : x86_64
Vendor         : 
Installed Size : 43.9 MiB
Installed      : No
Status         : not installed
Source package : tailscale-1.50.1-1.src
Upstream URL   : https://www.tailscale.com
Summary        : The easiest, most secure, cross platform way to use WireGuard + oauth2 + 2FA/SSO
Description    : 
    The easiest, most secure, cross platform way to use WireGuard + oauth2 + 2FA/SSO

```
So thus far: user error on my part.

phillxnet (approved):
OK, so bar my "doing it wrong" and discovering an anomaly re one of the tests; once tailscale is actually installed, visiting the config and submitting (all defaults) I get the following "Login" appear:
Tailscale-service-login
```
rleap15-5:~ # systemctl status tailscaled.service
● tailscaled.service - Tailscale node agent
     Loaded: loaded (/lib/systemd/system/tailscaled.service; enabled; vendor preset: disabled)
     Active: active (running) since Thu 2023-10-19 14:10:29 WEST; 6min ago
       Docs: https://tailscale.com/kb/
    Process: 16372 ExecStartPre=/usr/sbin/tailscaled --cleanup (code=exited, status=0/SUCCESS)
   Main PID: 16415 (tailscaled)
     Status: "Stopped; run 'tailscale up' to log in"
      Tasks: 10 (limit: 2334)
     CGroup: /system.slice/tailscaled.service
             └─ 16415 /usr/sbin/tailscaled --state=/var/lib/tailscale/tailscaled.state --socket=/run/tailscale/tailscaled.sock --port=41641

Oct 19 14:10:30 rleap15-5 tailscaled[16415]: Backend: logs: be:3c910da634826d267eded683e93b9dbe247ea758eb2c0207468b5298fd7a0244 fe:
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: Switching ipn state NoState -> NeedsLogin (WantRunning=false, nm=false)
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: blockEngineUpdates(true)
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: health("overall"): error: state=NeedsLogin, wantRunning=false
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: wgengine: Reconfig: configuring userspace WireGuard config (with 0/0 peers)
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: wgengine: Reconfig: configuring router
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: wgengine: Reconfig: configuring DNS
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: dns: Set: {DefaultResolvers:[] Routes:{} SearchDomains:[] Hosts:0}
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: dns: Resolvercfg: {Routes:{} Hosts:0 LocalDomains:[]}
Oct 19 14:10:30 rleap15-5 tailscaled[16415]: dns: OScfg: {}
rleap15-5:~ # echo $?
0
```

After using the "Login" which opened a new tab to initialte a Tailscale sign-up/login with machine registration attached. Once logged in we get:
Login-to-Tailscale-auth-rockstor-instance


I was informed/provided-with a link in the same session for admin to accept this new machine. This was done successfully.  The Tailscale 'Login' in Rockstor services page had then changed to the on/off switch in On position.

Network is as per PR text:
network-connections-tailscale0-connected

And sharing this machine via an invite link with @FroggyFlox provided the capability to ping from their tailnet. This is with the default configuration.

As my working knowledge of Tailscale is somewhat limited I'd like to get this out into our testing channel so folks more knowledgable there an test what we have to-date.

And issue can be created for @Hooverdan96 key point re inclusive regex improvement on hostname validation and likely some enhancements to cope more elegantly with no Tailscale installed situation.

phillxnet:
In addition @FroggyFlox has now tested Rockstor Web-UI access to the build machine used in this end-to-end RPM build/install test. So that's all looking quite promising. I'll go ahead and merge this so we can get to the next testing release. Some follow-ups likely but we can issue/pr them in turn.
To date:
- @Hooverdan96 note re hostname validation improvements.
- @FroggyFlox re blocking our native network edit for the tailscale interface. We likely want to replace this with a link to the servcie 'spanner' to centralise these settings.

N.B. only the default configuration has been tested by myself.

PR #2718, opened 2023-10-19
Fix mocking insufficiencies in system.network.py #2717

### LINKED ISSUE #2717: Initial rpmbuld failure - test_disable_ip_forwarding
During the initial review of #2712 (comment https://github.com/rockstor/rockstor-core/pull/2712#pullrequestreview-1687632231) there was an as yet unexplained fail in a new test regarding ip forwarding:
```
FAIL: test_disable_ip_forwarding (rockstor.system.tests.test_system_network.SystemNetworkTests)
test proper call of os.remove() and final sysctl command
  File "/home/buildbot/worker/Poetry-Build-on-Leap15-4/rpmbuild/rockstor-core-5.0.5-0/src/rockstor/system/tests/test_system_network.py", line 795, in test_disable_ip_forwarding
    self.mock_os_remove.assert_called_once_with("/etc/sysctl.d/99-tailscale.conf")
  File "/usr/lib64/python3.9/unittest/mock.py", line 918, in assert_called_once_with
    raise AssertionError(msg)
AssertionError: Expected 'remove' to be called once. Called 0 times.
Ran 279 tests in 32.169s
FAILED (failures=1)
```
On further fresh rpmbuild attempts (other OS/arch targets) this has been seen again on all systems that have not previously done an rpmbuild with this test included. We likely have some inadvertent system sensitivity/interaction here.

FroggyFlox:
Fixes #2717

As detailed in #2717, we have two mocking deficiencies:
 - one that resulted in one test actually writing a file to disk
 - one failing to properly mock `os.scandir()` (we were thus depending on the real system state)

This pull request fixes the former by combining 2 previously split tests, thereby sharing the same mock (we're not writing to disk anymore). To fix the latter, we instantiate a new class used to mock `os.scandir` return values that we can now properly control.

To test, Tailscale was entirely removed from the system and the repo removed as well. All tests now pass and the absence of `/etc/sysctl.d/99-tailscale.conf` (which was erroneously making `test_disable_ip_forwarding()` pass before) was verified.
```
buildvm155:/opt/rockstor/src/rockstor # export DJANGO_SETTINGS_MODULE=settings
......................................................................................................................................................................................................................................................................................
Ran 278 tests in 15.243s

OK
```

Hopefully this really corrects the situation this time and not just on my machine.

### DIFF
--- a/src/rockstor/system/tests/test_system_network.py
+++ b/src/rockstor/system/tests/test_system_network.py
@@ -13,7 +13,7 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
 import unittest
-from unittest.mock import patch, mock_open, call, MagicMock
+from unittest.mock import patch, mock_open, call
 
 from system.exceptions import CommandException
 from system.network import (
@@ -24,6 +24,25 @@ from system.network import (
     SYSCTL,
     disable_ip_forwarding,
 )
+
+
+class MockDirEntry:
+    """Create a fake DirEntry object
+    A DirEntry object is what's os.scandir returns. To properly mock os.scandir,
+    we thus need to re-create what it would return. We currently only need some
+    of the attributes that a real DirEntry has, so let's include only these:
+    - name
+    - path
+    - is_file: bool
+    """
+
+    def __init__(self, name, path, is_file):
+        self.name = name
+        self.path = path
+        self._is_file = is_file
+
+    def is_file(self):
+        return self._is_file
 
 
 class SystemNetworkTests(unittest.TestCase):
@@ -741,7 +760,7 @@ class SystemNetworkTests(unittest.TestCase):
         with self.assertRaises(CommandException):
             get_con_config(con_name)
 
-    def test_enable_ip_forwarding_write(self):
+    def test_enable_ip_forwarding(self):
         """enable_ip_forwarding() calls write with the correct contents
         Enable_ip_forwarding should write to /etc/sysctl.d/99-tailscale.conf
         with the following lines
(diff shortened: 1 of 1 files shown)


phillxnet (approved):
@FroggyFlox Thanks for getting to this fix so quickly.
By way of confirmation, I have, with git release 5.0.5 code confirmed that we have the failure when the file "/etc/sysctl.d/99-tailscale.conf" does not exist, but there after it does exist. And further test runs all pass.

Removing the file again and building an RPM but with this proposed changed included and all is well: in that the %check scriptlet runs all tests just fine and there-after there is still no /etc/sysctl.d/99-tailscale.conf file.

A further run of all tests via cli from the RPM installed code also succeeds and again without this file having been created.

So all good. Well done on the minimalist mock re MockDirEntry. There is a facility to mock an entire filesystem I believe - but never used it and can't remember what it was! Maybe with our recent Python updates we could look to this for some of our file interactions.

FroggyFlox:
Thanks for the confirmation; I'm glad it holds up during the %check step!

> Well done on the minimalist mock re MockDirEntry.

You've got the collective mind of the web to thank for that one.

> There is a facility to mock an entire filesystem I believe - but never used it and can't remember what it was! Maybe with our recent Python updates we could look to this for some of our file interactions.

I've seen references to `pyfakefs` (https://pypi.org/project/pyfakefs/) when trying to figure these tests. Maybe that's the one you're thinking of? It seems to work with both pytest and unittest.

phillxnet:
@FroggyFlox Re:

> I've seen references to `pyfakefs` (https://pypi.org/project/pyfakefs/) when trying to figure these tests. Maybe that's the one you're thinking of? It seems to work with both pytest and unittest

Yes that sounds like it, not certain though. I heard a mention of such a thing on PythonBytes or TalkPython and Mr Python testing, Brian Okken, @brianokken@fosstodon.org made mention.
---
OK just found at least one mention:
Apparently in the following episode:
https://pythonbytes.fm/episodes/show/277/its-a-python-package-showdown
Special guest: Thomas Gaigher makes mention.

But I expect Mr Okken has mentioned it on other occasions :) .

PR #2719, opened 2023-10-20
Use regular expressions to validate tailscale hostname #2714

### LINKED ISSUE #2714: [Tailscale]: consider regex filtering for hostname validation/adjustments
Follow up to the excellent initial pull request, regarding how hostnames are evaluated and adjusted in the corresponding python code.

              quick question, can a hostname not have unicode characters in it?  If that's the case, wouldn't some "inclusive" regex be better than excluding what we currently know as "special" characters"?

_Originally posted by @Hooverdan96 in https://github.com/rockstor/rockstor-core/pull/2712#discussion_r1364553929_

FroggyFlox:
Fixes #2714 
@Hooverdan96, @phillxnet: ready for review.

Following up on @Hooverdan96's suggestion to leverage regular expressions to better validate the hostname key in the Tailscale service config.

The current mechanism uses a hard-coded list of characters to exclude them from a custom tailscale hostname if present. This is rather ugly and does not scale to all the characters (such as unicode) that need to be excluded.

Switch to using regular expressions to exclude only characters from the hostname that do not belong to that regular expression.

All tests still pass...
```
......................................................................................................................................................................................................................................................................................
Ran 278 tests in 15.039s

OK
```

... and I was able to configure and turn ON the Tailscale service with a custom hostname being taken in effect (as seen on the Tailscale machines dashboard).

### DIFF
--- a/src/rockstor/system/tailscale.py
+++ b/src/rockstor/system/tailscale.py
@@ -104,32 +104,20 @@ def extract_param(param: str) -> str:
 
 
 def validate_ts_hostname(config: dict) -> dict:
-    """Ensure hostname is alphanumeric with hyphens only
-    No special character (including hyphens) is allowed as a custom hostname.
+    """Applies a set of rules to validate hostname
+    Keep each character in hostname only if:
+    - not a special character
+    - not a unicode character
+    - is hyphen
+    - replace underscore with hyphen
 
-    "hostname": "rock-dev_@#~!$%^&*()+123"
+    "hostname": "rock-dev_@#~!$%^&*()+123ü"
     should return
     "hostname": "rock-dev-123"
     """
     if "hostname" in config:
         config["hostname"] = re.sub("_", "-", config["hostname"])
-        to_exclude = [
-            "@",
-            "#",
-            "~",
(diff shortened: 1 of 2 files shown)


phillxnet (approved):
@FroggyFlox Thanks for this follow-up. Much appreciated.

RPM built OK with this included so all tests passed via %check scriptlet. And your addition to the test data confirms the added sensitivity and move to inclusion rather than list of chars to exclude.

Web-UI test irrelevant as we already have your prior equivalent character inclusion via the following form validator:

https://github.com/rockstor/rockstor-core/blob/ee4af7c1c632eb61434e2560c5a38ed9944f0099/src/rockstor/storageadmin/static/storageadmin/js/views/configure_service.js#L49-L52

FroggyFlox:
Thanks a lot for the review and RPM build testing, @phillxnet.
Given your approval and passing tests, I'll go ahead and merge this.
@Hooverdan96, let me know if you see something needing validation that I missed and I'll create a dedicated PR. I did not think about the unicode characters so thank you very much for bringing it up in your previous review!

PR #2721, opened 2023-10-21
Add/Update help icon linking to docs #2720

### LINKED ISSUE #2720: (t) Update HELP button link to Tailscale docs
In #2712, a HELP icon was added to the configuration modal header (see below):
image

Clicking this icon currently links to the documentation root but now that a dedicated documentation for the Tailscale service exists, we should point to that instead: https://rockstor.com/docs/interface/system/services.html#tailscale

FroggyFlox:
Fixes #2720 

In the configuration modals for our Services, some include links to their respective docs while other do not.

This pull requests harmonizes these discrepancies by adding the same help icon linking to the Service's section in our docs.

Also includes:
  - minor update of Tailscale service configuration header so that it matches the other services
  - update copyright statements

The configuration modals for all affected services were verified:
  - to include the new/updated HELP icon
  - to link to the appropriate section of our docs

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/services/configure_active-directory.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/services/configure_active-directory.jst
@@ -1,6 +1,6 @@
 <script>
   /*
-  * Copyright (c) 2012-2020 RockStor, Inc. <http://rockstor.com>
+  * Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
   * This file is part of RockStor.
   *
   * RockStor is free software; you can redistribute it and/or modify
@@ -14,7 +14,7 @@
   * General Public License for more details.
   *
   * You should have received a copy of the GNU General Public License
-  * along with this program. If not, see <http://www.gnu.org/licenses/>.
+  * along with this program. If not, see <https://www.gnu.org/licenses/>.
   *
   */
 </script>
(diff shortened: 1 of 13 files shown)


phillxnet (approved):
@FroggyFlox just done a test RPM build and all links work as intended:

new-help-links

But we are missing a link for replication. But then we don't have a proper replication doc entry yet anyway. Akin to the ShellInaBox & SNMP service also.

S.M.A.R.T? but I think you are working on that as we speak. Or we addressing as a follow-up.

FroggyFlox:
> But we are missing a link for replication. But then we don't have a proper replication doc entry yet anyway. Akin to the ShellInaBox & SNMP service also.
> 
> S.M.A.R.T? but I think you are working on that as we speak. Or we addressing as a follow-up.

Correct...
For S.M.A.R.T., we do have a nice write-up for its configuration (your doing, probably) but we do not link to it in the Services section of the docs. I think we should correct that first and then I'll amend that PR. There is no rush in getting this current PR in so I'll update the docs one first if, I think.

FroggyFlox:
@phillxnet, I've now pushed a new commit (https://github.com/rockstor/rockstor-core/pull/2721/commits/6e5b5aa74c037ddd2d0922566b306c9cbdba448c) with the same help icons-links for the configurable services that were missing a section in our docs previously.
Let me know if you prefer a squash + force-push.

phillxnet:
> Let me know if you prefer a squash + force-push.

I generally squash locallly and delete GitHub version of the multi-commit branch, which will auto close this branch. Then create a new PR from the freshly pushed to GitHub branch. This avoids a tricky situation where our back-end test setup can sometimes fail to merge force pushed changes. This enables me to easily test build to RPM from a PR.

Hope that's OK at your end. We also then get back to a single commit and cleaner history. As in this case there is not need for attribution of other collaborators as it's all your own work to date anyway in this branch.

PR #2727, opened 2023-10-30
Update django-oauth-toolkit #2710

phillxnet:
Remove pinning for django-oauth-toolkit and remove explicit declaration of oauthlib as it is a dependency of django-oauth-toolkit.

Re-address prior work-around for older oauth2_provider migration file silently failing to apply, and holding up all subsequent oauth2_provider migrations, as this migration file, and a few subsequent ones, have now been squashed upstream. "oauth2_provider" is part of django-oauth-toolkit.

# Includes:
- Added logging for before & after django-oauth-toolkit migration.
- Adopt dynamic client_secret for internal Oauth app. As from Django Oauth Toolkit 2.x onwards, Oauth app client_secret is hashed within Django's database, dictating that we can no longer source this secret from the db for our internal cli client app token requests. Move to establishing a dynamic Oathapp client_secret, established in settings.py, and reset by rockstor-bootstrap.service, i.e. on each service restart/reboot.
- Adding a requests timeouts to client token requests.
- Arbitrary fsting application.
- Update disk, pool, share, snap state every 20s not every minute.
- Abandon rockstor-bootstrap.service start (boostrap scritp) after 10, not 15 attempts.

Fixes #2710 
See issue text, and this pull requests prior draft #2726 for development context.

PR #2732, opened 2023-11-04
Address redundancy re database setup #2729

### LINKED ISSUE #2729: Address redundancy re default db setup
As highlighed by @FroggyFlox & @Hooverdan96 as part of #2723, which concerns the smart_manager database and our services setup, we may similarly have a redundancy in how we establish, after dropping storageadmin, a new setup of the default databases "storageadmin".

As Rockstor pre-dates south's integration into Django we have some layers of legacy here. We also have. as part of our DB_SETUP in initrock (for new database instances only), a straight dump of all that has gone before via:
```
"psql storageadmin -w -f {}/storageadmin.sql.in".format(CONF_DIR)
```
However this dump contains many now redundant, and conflicting, migration details. For example, in this dump/restore file we have the following south migration detail as having been already applied within this sql dump:

> 6	oauth2_provider	0001_initial	2016-11-24 15:00:22.623774-08

And possibly then responsible for the indicated applied state for oauth2_provider's migrations. See the following comment for details:
https://github.com/rockstor/rockstor-core/pull/2727#issuecomment-1785866753

Which pertains to an older version of the initial db setup file of oauth2_provider (part of Django-oath-toolkit - an external app).
However as detailed in #2710 comment: https://github.com/rockstor/rockstor-core/issues/2710#issuecomment-1779566523
This initial setup/migration file for oauth2_provider has since been re-used as a squashed version now containing many of the follow-up model additions/alterations. Leading, in another part, to this issue's creation, to examine if our initial database setup, consisting of an initial (now old) db snapshot/dump file's restore, should be dropped in favour of using only migrations files to achieve the same. Especially give the dump contains models that are the responsibility of external applications. Inevitably leading to breakage/fragility when those projects modify/squash their migration files.

South migration files, and now Django's also, (over the life time of Rockstor) are 'tagged' as having been applied by their name only via a dedicated db table. This is thought to have now lead to our current inability to re-create a complete/sane oauth2_provider db setup, regarding new db setups only, our supposedly pre-applied (via dump file) migration now contains less than the post squash (into 0001_initial) migrations assume. And so some interim migrations are now missing from our external Django-oauth-toolkit library. But only on fresh db setups, as pre-existing databases prior to our ongoing work in #2727 used versions of this same library prior to this upstream squashing event.

phillxnet:
We currently use both an initial (years old) pg_dump, & Django's native migration system to setup our initial database structure. This patch proposes that we drop the initial setup via sql.in files and normalise on using only the Django migration files. The fix purpose here is to accommodate upstream squash events such as one observed in Django-oath-toolkit, which introduced a migration conflict with our existing initial/static db setup via the old storageadmin.sql.in dump file.

# Includes
- Revised database router to ensure the existing model/db separation regarding the smart_manager app and its use of only the dedicated smart_manager database. Along with migration awareness, so that model metadata 'app_label' informs each migration appropriately: wise to the currently specified detabase during `django-admin migrate` runs.
- Removed old/legacy and now redundant smartdb.sql.in & storageadmin.sql.in files, and south_migrations archive
directories for both databases.
- Move db setup/migration code in initrock script to fstrings.
- DB migration mechanisms not associated with the initial DB setup are not functionally modified, as this is outside the
scope intended for this patch.

Fixes #2729


phillxnet:
# Testing
Initial DB setup (no .initrock file) looks to be as intended; under the Django migrations only setup arrangement.
And our existing (no functional change) post db existence (.initrock file exists) migration complex (needs focused issue) appears to still function without error on this new migrations only instantiated DB.

## TODO Testing
RPM updates (using this new initrock) from prior/ongoing database instances (i.e. v4.6.1-0) have yet to be proven.
Similarly these modification to initrock have also not been proven for the related PR of #2727 regarding the instantiation of the initial DBs with an updated Django-oath-toolkit, where our prior pg_dump setup method had reached it's limits and represented a blocker to #2727: hence the related nature of this PR.

phillxnet:
On a Leap 15.5 base amd64 arch we have:
- A successful RPM build which then installs successfully on the build host with no prior database in place: a clean install.
- The same host with 5.0.5-0 rpm installed, the last testing rpm released in this phase, successfully updated to the rpm build against this PR.
- A 15.4 host likewise built and installed a-fresh an RPM build against this PR.
- A 15.4 host was similarly updated from a fresh install of a 4.6.1-0 stable RPM to the 5.0.5-2732 RPM built against his PR.

phillxnet:
Moving to squash this Draft PR's commits and present the same in a fresh PR under a single commit.

PR #2736, opened 2023-11-10
Ease database diagnosis via local IP access configuration #2730

### LINKED ISSUE #2730: Ease database diagnosis via local IP access configuration
We currently limit our Host Based Authentication (`./conf/pg_hba.conf`) to what is required functionally and for running our tests. That is unix socket access only. It is proposed that we add the following, **but remarked out**, configurations that enable local IP host access that can be invaluable while working on database related issues and using for example a remove PyCharm setup where a IDE client exists locally on the development VM instance:

```
host   storageadmin    rocky              127.0.0.1/32            md5
host   smartdb         rocky              127.0.0.1/32            md5
host   storageadmin    rocky              ::1/128                 md5
host   smartdb         rocky              ::1/128                 md5
```

phillxnet:
Add to Postgresql's Host Based Authentication config (pg_hba.conf), remarked out examples of enabling IPv4 & IPv6 based access, from localhost. Typical DB interfaces found in IDEs are limited to IP access only, and so cannot connect using the unix socket only configuration used by Rockstor.

Fixes #2730

### DIFF
--- a/conf/pg_hba.conf
+++ b/conf/pg_hba.conf
@@ -96,3 +96,9 @@ local   test_smartdb         rocky                                   md5
 # local   replication     all                                     peer
 # host    replication     all             127.0.0.1/32            ident
 # host    replication     all             ::1/128                 ident
+
+# "host" IPv4 & IPv6 connections - e.g. IDE localhost only DB diagnostics
+# host   storageadmin    rocky              127.0.0.1/32            md5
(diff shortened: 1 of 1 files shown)


phillxnet:
As this is a 'no functional change' addition I'll merge directly to get this developer ease addition in-play.

PR #2745, opened 2023-11-14
Update dbus python dependency to latest #2744

### LINKED ISSUE #2744: Update dbus python dependency to latest
Our dbus-python dependency was previously constrained by our python version. This is not longer the case. We currently have:

> dbus-python = "==1.2.18"

It is proposed, in the interests for issues such as

"Avoid system calls to interact with systemd services" #2680 where we have:

> A better approach would most likely be to directly interact with systemd DBus. While we already have the dbus library as a dependency for interacting with the SSSD DBus responder InfoPipe, there is a python library to simplify interacting with systemd units and services: pystemd

While the linked issue, and proposed systemd dbus library, does not currently [depend](https://github.com/systemd/pystemd/blob/main/pyproject.toml) on dbus-python; it would never-the-less be advisable to unpin this library to enable our ongoing Python updates.

PyPi: https://pypi.org/project/dbus-python/
Project source: https://gitlab.freedesktop.org/dbus/dbus-python/
Project docs: https://dbus.freedesktop.org/doc/dbus-python/
Release history: https://dbus.freedesktop.org/doc/dbus-python/news.html#release-history

phillxnet:
Remove explicit pinning of dbus-python, enabling: 1.2.18 to 1.3.2 update as per poetry.lock.

Fixes #2744 

## Testing
See referenced issue commit: https://github.com/rockstor/rockstor-core/issues/2744#issuecomment-1810044651
All unit test pass.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -75,11 +75,15 @@ test-randomorder = ["pytest-randomly"]
 
 [[package]]
 name = "dbus-python"
-version = "1.2.18"
+version = "1.3.2"
 description = "Python bindings for libdbus"
 category = "main"
 optional = false
-python-versions = "*"
+python-versions = ">=3.7"
+
+[package.extras]
+doc = ["sphinx", "sphinx-rtd-theme"]
+test = ["tap.py"]
 
 [[package]]
 name = "deprecated"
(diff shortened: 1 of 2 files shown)


phillxnet:
A test 5.0.5-2745 RPM was build and installed successfully.

As this PR requires an sssd setup, a current omission in my test environment, I'm not able to test this adequately as it is my understanding that we only use dbus-python when interacting with sssd for its users/groups.

I was able to create & delete both a user and group using the PR RPM instantiated install.

@FroggyFlox & @Hooverdan96
I propose that we merge this to avoid hold-up of our pending Python/Django updates and from a quick look there doesn't appear to have be any breaking changes in this update (from changelog) - bar some Python 2 specific calls types. I would have expected them to error out during my exercise of users.py however, as they were removed: which should have broken the generation of a pyc file during the initial Python parsing of users.py.

FroggyFlox:
Agreed. Thanks a lot for that work!

Issue #2728, opened 2023-10-30
Adopt dedicated secrets management library


phillxnet:
We have an ongoing need to centralise our secrets management. This need has more recently been highlighted within the following PR/Issue: "Update django-oauth-toolkit #2710" #2727 where we had need to establish a new means to maintain an install-persistent OAUTH_INTERNAL_APP client secret. As this function was deemed to be outside the scope of the linked issue/PR the proposal is being made in this issue. The temporary work-around in #2727 was to establish a per-boot/rockstor-bootstrap.service  persistent secret. This however is inappropriate for such facilities as replication where we require install persistence for secrets.

Popular candidates would be:
- https://pypi.org/project/python-dotenv/
- https://pypi.org/project/pysecret/

Proposed initial focus would be our settings.py file.

phillxnet:
An alternative, but a little more complex/heavy-weight approach here, could be to 'standardise' on keyring: https://github.com/jaraco/keyring

> The Python keyring library provides an easy way to access the system keyring service from python. It can be used in any application that needs safe password storage.

This library is more naively concerned with desktop associated keystores, i.e. from the README.rst again:

> These recommended keyring backends are supported:
> 
> - macOS [Keychain](https://en.wikipedia.org/wiki/Keychain_%28software%29)
> - Freedesktop [Secret Service](http://standards.freedesktop.org/secret-service/) supports many DE including GNOME (requires [secretstorage](https://pypi.python.org/pypi/secretstorage))
> - KDE4 & KDE5 [KWallet](https://en.wikipedia.org/wiki/KWallet) (requires [dbus](https://pypi.python.org/pypi/dbus-python))
> - [Windows Credential Locker](https://docs.microsoft.com/en-us/windows/uwp/security/credential-locker)

However we have also the option of:

> Other keyring implementations are available through [Third-Party Backends](https://github.com/jaraco/keyring#third-party-backends).

Among these indicated "third-party-backends" we have a potential candidate of:

> - [keyrings.cryptfile](https://pypi.org/project/keyrings.cryptfile) - Encrypted text file storage.

Where from the linked (in PyPi) GitHub homepage of this project we have:

https://github.com/frispete/keyrings.cryptfile#notes

> You can avoid the interactive getpass() request for the keyring password by supplying kr.keyring_key = "your keyring password" before calling any other methods on the keyring. The following example shows a simple way to retrieve the password from an environment variable KEYRING_CRYPTFILE_PASSWORD, when present:

[...]

PR #2747, opened 2023-11-14
Update pyzmq dependency to latest #2746

### LINKED ISSUE #2746: Update pyzmq dependency to latest
Our current explicit pinning was due to a Py2.7 dependency, however we have now exceeded the Py3.6 stated newest dependency for this pinning:

> # pyzmq requires libzmq5 on system unless in wheel form.
> pyzmq = "==19.0.2"  # Last specifying Python 2 on PyPi page.

(19.0.2 released July 2020)

It is proposed that we move to the latest version to remove blocker for ongoing Python and Django updates:

PyPi: https://pypi.org/project/pyzmq/
Project page: https://pyzmq.readthedocs.io/en/latest/
> Binary distributions (wheels on PyPI) of PyZMQ ship with the stable version of libzmq at the time of release, ...

Changelog: https://pyzmq.readthedocs.io/en/latest/changelog.html

Latest: (25.1.1 released August 2023)

phillxnet:
Remove explicit pinning of pyzmq, enabling: 19.0.2 to 25.1.1 update as per poetry.lock.

Fixes #2746

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -347,11 +347,14 @@ python-versions = "*"
 
 [[package]]
 name = "pyzmq"
-version = "19.0.2"
+version = "25.1.1"
 description = "Python bindings for 0MQ"
 category = "main"
 optional = false
-python-versions = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*"
+python-versions = ">=3.6"
+
+[package.dependencies]
+cffi = {version = "*", markers = "implementation_name == \"pypy\""}
 
 [[package]]
 name = "requests"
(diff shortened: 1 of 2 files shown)


phillxnet:
The 0mz library is used exclusively by our replication sub-system that is currently suffering a regression regarding our recent Django-oath-toolkit update (in current testing phase) where upstream no longer provide a method to retrieve, from DB, our cliapp secret. We are as such awaiting development on the following issues:

See: Adopt dedicated secrets management library #2728

Replication regression issue: "Replication regressions" #2748

phillxnet:
@FroggyFlox @Hooverdan96 we have not our first/last chicken and egg scenario. I'm unable to test this as we have outstanding breaking updates already: however we also have a pressing need to move our base Python and Django along. And this library is one of our oldest yet to be updated.

This PR does not break rpmbuild on a 15.5 amd64 of a 5.0.5-2747 RPM, which then install on the host OK, but given the above I'm then not able to test replication anyway. As such I've broken out the following issue:

https://github.com/rockstor/rockstor-core/issues/2748

To collect current replication breaking/suspected-breaking changes.

But we still have to move-along with our pending Python and Django (and Poetry) updates: all of which are intertwined, so I propose that we merge this PR to at least remove all/any blockers on those more fundamental fronts. These temporary breakages are, I think, an inevitability give our massive changes of late. But to date we have only Replication as a known casualty.

FroggyFlox:
@phillxnet , I completely agree.

Updating all of this is a requirement to fix replication anyway so we should get that in first so that we can fix it.

PR #2751, opened 2023-11-16
Update Django to latest 4.2 LTS #2750

phillxnet:
Update Django from 3.2.23 to 4.2.7 via `poetry update` and pyproject.toml.
## Includes:
- Move from django.utils.timezone utc to datetime.timezone.utc: RemovedInDjango50Warning
- Remove to-be-deprecated USE_L10N setting, prior setting now enforced: RemovedInDjango50Warning
- Add comment re Django 5.0 change in default USE_TZ value: Future default already adopted.
- ~~Remove non-existent/unused STATICFILES_DIRS entry.~~
- Adopt new-in-Django-4.2 STORAGES setting to replace STATICFILES_STORAGE setting.

Fixes #2750

Issue #39, opened 2013-05-08
nginx and gunicorn timeouts


schakrava:
Nginx and Gunicorn need to be configured properly to avoid nginx gateway timeout.

I don't have a reproducible case but seen it every time a certain operation, share creation for example, takes a while.

Until we change the api implementation to return immediately, it worth generously increasing necessary timeout variables in nginx and gunicorn configuration.

PR #2752, opened 2023-11-20
Update Django to latest 4.2 LTS #2750

### LINKED ISSUE #2750: Update Django to latest 4.2 LTS
Our current testing phase is focused on updating our dependencies. The second largest of which is Django itself (Python has mostly been addressed but is awaiting a Py3.10 update).

Canonical issue for Milestone: https://github.com/rockstor/rockstor-core/milestone/26

From: https://docs.djangoproject.com/en/4.2/faq/install/#what-python-version-can-i-use-with-django

Django version | Python versions
| - | - |
4.1	| 3.8, 3.9, 3.10, 3.11 (added in 4.1.3)
4.2	| 3.8, 3.9, 3.10, 3.11

phillxnet:
Update Django from 3.2.23 to 4.2.7 via `poetry update` and pyproject.toml.
## Includes:
- Move from django.utils.timezone utc to datetime.timezone.utc: RemovedInDjango50Warning
- Remove to-be-deprecated USE_L10N setting, prior setting now enforced: RemovedInDjango50Warning
- Add comment re Django 5.0 change in default USE_TZ value: Future default already adopted.
- Comment STATICFILES_DIRS setting re build.sh & rockstor-jslibs.
- Adopt new-in-Django-4.2 STORAGES setting to replace STATICFILES_STORAGE setting.

Fixes #2750 

For development history see also draft PR #2751

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -109,16 +109,16 @@ python-versions = ">=3.6"
 
 [[package]]
 name = "django"
-version = "3.2.23"
-description = "A high-level Python Web framework that encourages rapid development and clean, pragmatic design."
+version = "4.2.7"
+description = "A high-level Python web framework that encourages rapid development and clean, pragmatic design."
 category = "main"
 optional = false
-python-versions = ">=3.6"
+python-versions = ">=3.8"
 
 [package.dependencies]
-asgiref = ">=3.3.2,<4"
-pytz = "*"
-sqlparse = ">=0.2.2"
+asgiref = ">=3.6.0,<4"
+sqlparse = ">=0.3.1"
+tzdata = {version = "*", markers = "sys_platform == \"win32\""}
 
 [package.extras]
 argon2 = ["argon2-cffi (>=19.1.0)"]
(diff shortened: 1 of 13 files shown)


phillxnet:
As per development details in linked draft PR, these changes depend upon the current testing channel of rockstor-jslibs as we have the following merged, but not yet in-release tar, PR dependency:
[Add socket.io.min.js.map file #39](https://github.com/rockstor/rockstor-jslibs/pull/40)

phillxnet:
Our rockstor-core PR test back-end fakes latest matched rockstor-jslibs repo branch, testing in this case, to allow for pre-tagged GitHub releases. Given this mechanism we have a successful rpmbuild and install of a test 5.0.5-2752 rpm (15.5 amd64 host), along with a preliminary Web-UI setup and basic/scant by-hand test under web browser development tools -> network -> disable cache to ensure we are using the rpm provided jslibs only.

The resulting install also rebooted successfully with all core rockstor services stating as expected.

phillxnet:
@FroggyFlox & @Hooverdan96 I'll go ahead and merge this give the tests here and our unit tests executed during rpmbuild and the work done in the referenced draft PR. That jslibs issue was quite the surprise but we can re-approach that after we have our HTMX setup in place and are ready for Server Sent Events or the like.

Plus we need this in place as part of our ongoing Python updates, and related pending Poetry update.

PR #2762, opened 2023-12-12
Make explicit to systemd our NetworkManager dependency #2685

### LINKED ISSUE #2685: Make explicit to systemd our NetworkManager dependency
We use nmcli (Network Manager cli) and expect the associated systemd service, NetworkManager, to be enabled and running. As such, it is proposed that we explicitly 'Require' this service within our own systemd services.

## Current explicit 'Require's in our systemd services
| service | Requires |
| -- | -- |
| rockstor-pre | postgresql.service |
| rockstor | rockstor-pre |
| rockstor-bootstrap | rockstor |
| rockstor-hdparm | systemd-udev-settle |

Our installer already establishes this service, but it can only help to be explicit. This may also help with speeding up on-boarding for those doing rpm installs without first referencing our following doc entry: 
**Install on Vanilla openSUSE/SuSE SLES**: https://rockstor.com/docs/howtos/rpm_install.html
And in ground-up developmental setups, or deployments experimenting with new OS bases.

N.B. the postgresql.service `/usr/lib/systemd/system/postgresql.service` has the following:
`After=network.target`
however on some of our current and likely future platforms this does not imply/guarantee the NetworkManager.service but could for example still be provided by the wicked.service.

phillxnet:
Adds `After` &  `Requires` for NetworkManager.service to rockstor-pre.service. Primarily to surface to admins our dependency on nmcli.

Fixes #2685 

### Before:
```
rleap15-5:~ # systemd-analyze dot --require rockstor-pre.service
digraph systemd {
        "rockstor-pre.service"->"-.mount" [color="black"];
        "rockstor-pre.service"->"sysinit.target" [color="black"];
        "rockstor-pre.service"->"system.slice" [color="black"];
        "rockstor-pre.service"->"postgresql.service" [color="black"];
        "rockstor-pre.service"->"opt.mount" [color="black"];
        "rockstor-pre.service"->"shutdown.target" [color="red"];
        "rockstor.service"->"rockstor-pre.service" [color="black"];
}
...
```

### After:
```
lbuildvm:~ # systemd-analyze dot --require rockstor-pre.service
digraph systemd {
        "rockstor-pre.service"->"NetworkManager.service" [color="black"];
        "rockstor-pre.service"->"system.slice" [color="black"];
        "rockstor-pre.service"->"postgresql.service" [color="black"];
        "rockstor-pre.service"->"sysinit.target" [color="black"];
        "rockstor-pre.service"->"-.mount" [color="black"];
        "rockstor-pre.service"->"opt.mount" [color="black"];
        "rockstor-pre.service"->"shutdown.target" [color="red"];
        "rockstor.service"->"rockstor-pre.service" [color="black"];
        "multi-user.target"->"rockstor-pre.service" [color="grey66"];
}
...
```
# Caveat
We seem to have now inherited, via the new association/requires,  the multi-user.target. This may well put us a little later on in the boot sequence.

### DIFF
--- a/conf/rockstor-pre.service
+++ b/conf/rockstor-pre.service
@@ -1,7 +1,9 @@
 [Unit]
 Description=Tasks required prior to starting Rockstor
 After=postgresql.service
+After=NetworkManager.service
 Requires=postgresql.service
+Requires=NetworkManager.service
 
 [Service]
 Environment="DJANGO_SETTINGS_MODULE=settings"


phillxnet:
An rpm was build & installed successfully on a 15.4 x86_64 host, and was successfully rebooted.

Curiously there is no multi-user (see caveat) indicated in this instance.
```
rleap15-4:~ # systemd-analyze dot --require rockstor-pre.service
digraph systemd {
        "rockstor-pre.service"->"postgresql.service" [color="black"];
        "rockstor-pre.service"->"sysinit.target" [color="black"];
        "rockstor-pre.service"->"NetworkManager.service" [color="black"];
        "rockstor-pre.service"->"opt.mount" [color="black"];
        "rockstor-pre.service"->"system.slice" [color="black"];
        "rockstor-pre.service"->"-.mount" [color="black"];
        "rockstor-pre.service"->"shutdown.target" [color="red"];
        "rockstor.service"->"rockstor-pre.service" [color="black"];
}
...
```

FroggyFlox:
> We seem to have now inherited, via the new association/requires, the multi-user.target. This may well put us a little later on in the boot sequence.

Maybe irrelevant if that new target is actually not seen on an install, but would it mean we could see the rockstor services not being started on systems that fail to start their network? In itself it might not be a problem as a NAS without network is not really a thing, but I'm thinking of rate instances wherewe could have a delay/timeout in the network manager service as we've seen once (https://github.com/rockstor/rockstor-installer/issues/32).

phillxnet:
This PR has now been tested on all OS/arch targets re:
- rpmbuild
- resulting rpm install

Issue #2763, opened 2023-12-13
Surface auto-generated access key secrets prior to save


phillxnet:
Since "Update django-oauth-toolkit #2710" #2727 we can no longer retrieve the raw secret we auto-generate when adding a new Access Key: - rendering this capability unusable. To re-enable this facility, it is proposed that we add the ability to present within the Web-UI, just prior to save, the proposed secret; with the ability to enter/edit.

Currently, during entry, we have:
![add-access-key-dialog](https://github.com/rockstor/rockstor-core/assets/2521585/dc30455c-bfce-49ef-a2f1-26bef83be463)

where the associated access key ID & secret are auto-generated.

However give  django-oauth-toolkits change of stance re it's storage & retrieval of this key, we must now manage these secrets ourselves. In the special case of our internal `cliapp` app we now alternatively manage the secret via OS provided `pass`: see:

Adopt dedicated secrets management library #2728 #2758

And ongoing issue: Replication secret encrypted in Web-UI #2759

But for external/custom Access Key additions we could simply present the proposed new Access Key secret (with enter/edit option) along with a warning indicating that its retrieval will: there-after no longer be possible. This is more in line with modern practice and informs the user that they are responsible for recording the generated/edited/entered secret, so that it might be transferred to the relevant external application; while hence forth no longer being accessible from the Rockstor instance.

phillxnet:
The relevant UI template for this issue is as follows:

> src/rockstor/storageadmin/static/storageadmin/js/templates/access_keys/add_access_key.jst

PR #2764, opened 2023-12-13
Replication secret encrypted in Web-UI #2759

### LINKED ISSUE #2759: Replication secret encrypted in Web-UI
Prior to "Update django-oauth-toolkit #2710" #2727 we stored our CLI app secret in the database: and exposed that value within the Web-UI (to admin Web-UI users) for use in replication. However the linked PR introduced a replication regression contextualised in "Replication regressions #2748".

It is proposed that, given the recent resolution of:

Issue: "Adopt dedicated secrets management library" #2728
PR: "Adopt dedicated secrets management library #2728" #2758

has again returned us to an install static CLIENT_SERCRET, we can approach the issue of surfacing this secret within the Web-UI again, and not in its current hashed form: an artifact of upstream changes in django-oauth-toolkit.

phillxnet:
Special-case our internal cliapp re secret availability. Indicate all non-internal Access Key secrets as only available during creation.

Closes #2759 

## Includes
- Surfacing our pass/PGP encrypted raw secret to authenticated Web-UI logins.
- Indicate all other API credentials as not available.
- Brevity improvements re deletion attempt on cliapp message.

show-only-internal-cliapp-secret

### DIFF
--- a/src/rockstor/storageadmin/models/oauth_app.py
+++ b/src/rockstor/storageadmin/models/oauth_app.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2020 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2023 RockStor, Inc. <https://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


phillxnet:
This pending fix is required by / is a dependency for, the development of draft PR:
(t) replication spawn error #2766 #2769

I've now build a test rpm including this PR on a 15.5 base and all looks to be as intended. I.e. we now surface this 'special case', with the caveat that we need to follow this up re:

Surface auto-generated access key secrets prior to save #2763

PR #2781, opened 2024-01-21
Establish Postgres database format upgrade #2780

### LINKED ISSUE #2780: Establish Postgres database format upgrade
Thanks to @FroggyFlox & KarstenV on the forum for identifying this need. From  testing channel/branch v5.0.6.0 onwards we have fully updated our Django to 4.2 LTS, and the related python interface library of psycopg to V3: this has surfaced a Django/db-format incompatibility requiring an in-place db update capability.

From @FroggyFlox's forum comment here: https://forum.rockstor.com/t/after-updating-to-5-0-6-0-web-interface-is-not-accesible-and-network-shares-are-also-not-accesible/9226/29

We see that although we have previously update our Postgres dependency to 13, and enforce that binaries use via the alternatives system when doing an rpm update, this does not in-itself adjust the underlying DB format, and the newer start up scripts, if they find an older database, will in-fact invoke the relevant older binary of Postgres, regardless of the alternatives system.

New, or more recent installs are not affected assuming the use of our official installer as this always wipes the entire system disk, and so always establish a fresh and alternatives guided postgresql version. But Rockstor installs that have been updated from older Leap instances will inherit DB formats that were default at that time: to date we have reports of Postgres version 10 databases.

It is proposed that for the next Stable release we develop a script/mechanism to identify and then migrate, in-place, older db versions to the newer 13 version that we currently favour. From the reports in our ongoing testing phase we have the following from KarstenV's testing efforts with 5.0.6-0 (RC1) rpm. Where we significantly updated our Django (twice), as well as adopting psycopg3 which are both very significant update.

```
NotSupportedError(', ‘django.db.utils.NotSupportedError: PostgreSQL 12 or later is required (found 10.23).’, ‘’]
```
Due to the Postgresql startup scripts failing over to the older Postgres to accomodate the older DB format found:
```
Jan 18 22:28:33 RockstorNAS postgresql-script[7819]: Your database files were created by PostgreSQL version 10.
```
The latter resulted when removing the co-resident postgres v10 as openSUSE can carry multiple postgres package installs and auto-selects the newest able to read the database in question.

I.e. form @FroggyFlox's exposition of this issue on the linked form thread we have:

> This happens when the postgresql.service systemd service tries to start. If we look at that service (systemctl cat postgresql.service), we can see it tries to run the following:
> 
> ExecStart=/usr/share/postgresql/postgresql-script start
> That file is provided by the system postgresql-server package

and

> That script is just a bash script that does some checks, such as compatibility between postgresql versions. The part that is failing on your machine is:

```bash
if test -r $DATADIR/PG_VERSION ; then
    DATA_VERSION=$(cat $DATADIR/PG_VERSION)
    POSTGRES=/usr/lib/postgresql$(echo -n $DATA_VERSION | tr -d .)/bin/postgres
fi
if test -x /usr/bin/postgres; then
    ACTIVE=$(readlink -q -f /usr/bin/postgres)
    test -z "$POSTGRES" && POSTGRES="$ACTIVE"
fi
if test -n "$DATA_VERSION"; then
    if test -z "$ACTIVE" -o "$ACTIVE" != "$POSTGRES"; then
        echo " Your database files were created by PostgreSQL version $DATA_VERSION."
        if test -x "$POSTGRES"; then
            echo " Using the executables in $(dirname $POSTGRES)."
        else
            echo " Could not find executables for this version."
            echo " Please install the PostgreSQL server package for version $DATA_VERSION."
        fi
    fi
```

Similarly we might establish our own script to identify older 10 db instances: established by old official installers, and provide an in-place update mechanism via presumably: [pg_upgrade](https://www.postgresql.org/docs/13/pgupgrade.html) which must be run under the newer/target Postgres version involved in the migration. Note that a prior name for pg_upgrade was pg_migrator.

phillxnet:
Adds a shell script, to be run by rpm during update, to ensure any Postgres 10 (or later) formated DBs are updated to our current Postgres target version: established by rpm dependencies and the 'alternatives' system.

Fixes #2780 

# Testing

This script was developed on, and has had initial testing with, an install originally derived from the following installer:
> Rockstor-Leap15.3-generic.x86_64-4.1.0-0.install.iso Jan 14 2022

- The resulting install was then subscribed to the Stable updates channel.
- An underlying OS update to a 15.4 base was then enacted via:
**Distribution update from 15.3 to 15.4**: https://rockstor.com/docs/howtos/15-3_to_15-4.html
Resulting also in an automatic update to our last stable release of 4.6.1-0.

The above system then represents a reproducer for a pending stable update issue found in the current testing channel; where the far newer Django and psycopg3 there-in are no longer compatible with a Postgres 10 formatted DB: such as a system with the above, or older heritage, will have.

### DIFF
--- a/src/rockstor/scripts/db_upgrade.sh
+++ b/src/rockstor/scripts/db_upgrade.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+# exit on error
+set -o errexit
+#
+# Upgrade the database format from 10 to 13.
+# Testing RPM version 5.0.6-0 updated to Django LTS 4.2 & psycopg3.
+# These are no longer compatibility with databases created
+# with Postgres version 10, i.e. installs originally created from:
+# Rockstor-Leap15.3-generic.x86_64-4.1.0-0.install.iso Jan 14 2022
+# or earlier.
+# - RPM version 4.5.0-0 added a Postgresql V13 dependency.
+# - RPM version 5.0.2-0 enforced Postgresql alternative to V13.
+#
+# Uses pg_upgrade https://www.postgresql.org/docs/13/pgupgrade.html
+# Informed by:
+# - https://en.opensuse.org/SDB:PostgreSQL#Upgrading_major_PostgreSQL_version
+# - https://progress.opensuse.org/news/113
+# ps aufx | grep '^postgres.* -D'
+# V10: /usr/lib/postgresql10/bin/postgres -D /var/lib/pgsql/data
+#
+FROM_VERSION="$1"
+TO_VERSION="$2"
+
+if [ $# -ne 2 ]; then
+    echo "This script upgrades the Rockstor Postgres database format."
+    echo "It requires two integer numbers: the 'current' and 'target' DB formats."
+    echo "Example: 10 13"
+    exit 1
+fi
+# OpenSUSE postgres user home (~) is: /var/lib/pgsql/
+# By default databases are created in the "data" subdirectory.
+DATA_BASEDIR="/var/lib/pgsql"
+# Get current database format version via PG_VERSION file.
+if test -r $DATA_BASEDIR/data/PG_VERSION ; then
+    CURRENT_DATA_VERSION=$(cat $DATA_BASEDIR/data/PG_VERSION)
+else
+  echo "No DB PG_VERSION file found."
+  exit 0
+fi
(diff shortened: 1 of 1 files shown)


phillxnet:
Superseded-by #2785
as a rebase was required.

PR #2783, opened 2024-01-22
Failure to remove legacy poetry version in 5.0.6-0 rpm #2782

phillxnet:
Add legacy poetry install path to avoid failing its uninstall. Without this additional path, when build.sh is run from within RPM's constrained %posttrans scriptlet, we fail to find and uninstall our prior 1.1.15 poetry version. This leads to having both poetry 1.1.15 (via upstream installer) and our new pipx installed poetry 1.7.1. After having checked-for/uninstalled this legacy version we then need to remove the associated legacy path.

Fixes #2782

PR #2784, opened 2024-01-23
Failure to remove legacy poetry version in 5.0.6-0 rpm #2782

### LINKED ISSUE #2782: Failure to remove legacy poetry version in 5.0.6-0 rpm
Thanks to @FroggyFlox for highlighting this issue. During rpm update, we execute our build.sh script from within rpmbuild's %posttrans scriptlet. As from rpm 5.0.6-0 the build.sh shell script also:

- removes our legacy Poetry 1.1.15, installed (in **/root/.local/bin/**) via upstream installer script,
- and installs (in **/usr/local/bin/**) a far newer Poetry 1.7.1 under Py3.11 via OS supplied python3.11 pipx ,

This new Poetry update mechanism was part of the following pull request:
Update Poetry build system & normalise on Python 3.11 #2703 #2754 #2693 by @phillxnet @FroggyFlox in #2755

However; during more recent update testing concerning our next Stable release (from  our last), given we are now in Release Candidate phase in testin updates/branch, a 4.6.1-0 machine failed to update to a pre-release rpm version of 5.0.6-2781.

The  `/var/log/zypp/history` rpm log had the following contents.
```
# 2024-01-22 13:53:58 %posttrans(rockstor-5.0.6-2781.x86_64) script output:
# Current DB format is already 13.
# which: no poetry in (/usr/local/bin:/opt/rockstor/.venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
# creating virtual environment...
# creating shared libraries...
# upgrading shared libraries...
# installing poetry from spec 'poetry==1.7.1'...
# done! ✨ 🌟 ✨#   installed package poetry 1.7.1, installed using Python 3.11.5
#   These apps are now globally available
#     - poetry
# No config-backups-rpmsave directory found.
``
Noteworthy here is no the **no poetry in ...** line. When the prior install was a Poetry variant: 4.6.1-0 with the following poetry-install-txt:

```bash
installer:~ # cat /opt/rockstor/poetry-install.txt 
The currently activated Python version 3.6.15 is not supported by the project (~2.7).
Trying to find and use a compatible version. 
Using python2 (2.7.18)
Creating virtualenv rockstor in /opt/rockstor/.venv
Installing dependencies from lock file

Package operations: 29 installs, 0 updates, 0 removals

  • Installing certifi (2021.10.8)
  • Installing chardet (4.0.0)
  • Installing greenlet (2.0.2)
  • Installing idna (2.10)
  • Installing pytz (2022.6)
  • Installing six (1.16.0)
  • Installing urllib3 (1.26.12)
  • Installing django (1.11.29)
  • Installing gevent (1.1.2)
  • Installing oauthlib (3.1.0)
  • Installing python-engineio (2.3.2)
  • Installing requests (2.27.1)
  • Installing dbus-python (1.2.18)
  • Installing distro (1.6.0)
  • Installing django-oauth-toolkit (1.1.2)
  • Installing djangorestframework (3.9.3)
  • Installing django-pipeline (1.6.9)
  • Installing django-braces (1.13.0)
  • Installing gevent-websocket (0.9.5)
  • Installing gunicorn (19.10.0)
  • Installing huey (2.3.0)
  • Installing mock (1.0.1)
  • Installing psutil (5.9.4)
  • Installing psycogreen (1.0)
  • Installing psycopg2 (2.8.6)
  • Installing python-socketio (1.6.0)
  • Installing pyzmq (19.0.2)
  • Installing supervisor (4.2.4)
  • Installing urlobject (2.1.1)

Installing the current project: rockstor (4.6.1)
```

Where-as post this Leap 15.4 reproducer system (``zypper dup``'ed from 15.3) we have the following contents:

```
installer:~ # cat /opt/rockstor/poetry-install.txt 

Command ['python', '-I', '-W', 'ignore', '-c', 'import sys\n\nif hasattr(sys, "real_prefix"):\n    print(sys.real_prefix)\nelif hasattr(sys, "base_prefix"):\n    print(sys.base_prefix)\nelse:\n    print(sys.prefix)\n'] errored with the following return code 2

Error output:
Unknown option: -I
usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Try `python -h' for more information.
```

So we see the failure of the legacy Poerty to:
1. be found by the rpm invoked %posttrans env.
2. consequently of 1. not actually be removed.
```
linstaller:~ # which poetry
/root/.local/bin/poetry
```
and we also have the newer version inappropriately co-resident, as per pipx's feedback within the above log excerpt, and from:
```
installer:~ # ls -la /usr/local/bin/
total 4
drwxr-xr-x 1 root root 12 Jan 22 13:54 .
drwxr-xr-x 1 root root 66 Mar 15  2022 ..
lrwxrwxrwx 1 root root 33 Jan 22 13:54 poetry -> /opt/pipx/venvs/poetry/bin/poetry
```

It is proposed that in this setting we have an insufficient path to identify the existence of the legacy Poetry version and thus skip entirely running its uninstaller.

phillxnet:
Add legacy poetry install path to avoid failing its uninstall. Without this additional path, when build.sh is run from within RPM's constrained %posttrans scriptlet, we fail to find and uninstall our prior 1.1.15 poetry version. This leads to having both poetry 1.1.15 (via upstream installer) and our new pipx installed poetry 1.7.1.

## Includes:
- Remove dangling legacy poetry symlink - apparently neglected by legacy poetry installer/uninstaller.
- After the above there is an in-memory 'poetry' resolution to `/root/.local/bin/poetry' until path is reset. So use reset as opportunity to remove legacy paths.
- Add echo of PATH used by remainder of build.sh.
- Remove .venv related path from PATH as the entire directory is deleted in the %post RPM scriptlet. And build.sh (outside of development) is run after this: from %posttrans.

## Testing
See the following comment in this PR's draft predecessor:
https://github.com/rockstor/rockstor-core/pull/2783#issuecomment-1906651983

Fixes #2782
In that we now remove the legacy Poetry, where-as previously, the mechanism to update Poetry introduced in rpm 5.0.6-0 failed in this regard. This did no show-up in development as we prove each testing rpm's install (not update) capability on the build host, and in our rpm %build scriptlet the required legacy poetry path was established before hand.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,10 +4,17 @@ set -o errexit
 
 # Install Poetry, a dependency management, packaging, and build system.
 # Uninstall legacy/transitional Poetry version of 1.1.15
+PATH="$HOME/.local/bin:$PATH"  # account for more constrained environments.
 if which poetry && poetry --version | grep -q "1.1.15"; then
   echo "Poetry version 1.1.15 found - UNINSTALLING"
   curl -sSL https://install.python-poetry.org | python3 - --uninstall
+  rm --force /root/.local/bin/poetry  # remove dangling dead link.
 fi
+PATH="${PATH//'/root/.local/bin:'/''}" # null all legacy poetry paths
+# We are run, outside of development, only by RPM's %posttrans.
+# As such our .venv dir has already been removed in %post (update mode).
+PATH="${PATH//'/opt/rockstor/.venv/bin:'/''}" # null now removed .venv from path.
+echo "build.sh has PATH=$PATH"
 # Install Poetry via PIPX as a global app
 # https://peps.python.org/pep-0668/#guide-users-towards-virtual-environments
 export PIPX_HOME=/opt/pipx  # virtual environment location, default ~/.local/pipx
@@ -24,6 +31,7 @@ python3.11 -m pipx install poetry==1.7.1
 # https://github.com/python-poetry/poetry/issues/3078
 export LANG=C.UTF-8
 export PYTHONIOENCODING=utf8
+# /usr/local/bin/poetry -> /opt/pipx/venvs/poetry
 /usr/local/bin/poetry install --no-interaction --no-ansi > poetry-install.txt 2>&1
 echo
 


phillxnet:
@FroggyFlox & @Hooverdan96
As this essentially fixes only one part of other ongoing show-stoppers in the testing channel, i.e. the also pending db_update script PR, and its rpmbuild intergration, and they, in turn, reply on this PR's improvements/fix; I'll go ahead and merge.

We have more work to do in this area: but that is never news :).

PR #2804, opened 2024-02-26
Replace raw_input() with input()

### LINKED ISSUE #2803: [t] OpenVPN scripts for key generation and client enumeration don't work anymore with python3.11+
Thanks to user [ZeusMR](https://forum.rockstor.com/u/zeusmr) who uncovered a renamed python 2.x function. The `raw_input`  function was used in one of the specialized scripts connected to the OpenVPN Rockon to generate keys, etc.

Unfortunately, this function has been renamed to `input()` so the interpreted won't recognize the `raw_input` function name anymore.

https://peps.python.org/pep-3111/

This affect these two script functions that were created for the OpenVPN Rockon:

https://github.com/rockstor/rockstor-core/blob/6c86e3858079ce8edc62c62684a7f2cf50fb6d79/src/rockstor/scripts/ovpn_util.py#L33-L50

It will hopefully be as easy as replacing the renamed function name with the new one to get it to work again.

Hooverdan96:
Fixes #2803.

Replacing raw_input() function with input() instead.

Testing done in a test install manually replacing the commands.

### DIFF
--- a/src/rockstor/scripts/ovpn_util.py
+++ b/src/rockstor/scripts/ovpn_util.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2012-2020 RockStor, Inc. <http://rockstor.com>
+Copyright (c) 2012-2024 RockStor, Inc. <http://rockstor.com>
 This file is part of RockStor.
 
 RockStor is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


phillxnet:
@Hooverdan96 Thanks for seeing to this. My apologies for missing this previously. And thanks for testing this substitution.

Quoting you here from your linked issue:

> this function has been renamed to input() so the interpreted won't recognize the raw_input function name anymore.

> https://peps.python.org/pep-3111/

Where the upstream reference states:

> The existing raw_input() function will be renamed to input().

Merging as a consequence: as this looks like, as stated, a straight-forward omission in our Python 2.7 to now 3.11 work.

Issue #2807, opened 2024-03-12
[t] Assigning custom GID during User Group creation, results in syntax error


Hooverdan96:
when trying to create a new User Group using a custom GID (e.g., 1001), the system will throw a syntax error. Here is the traceback:
```
            Traceback (most recent call last):
  File "/opt/rockstor/src/rockstor/system/osi.py", line 236, in run_command
    p = subprocess.Popen(
        ^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib64/python3.11/subprocess.py", line 1883, in _execute_child
    self.pid = _fork_exec(
               ^^^^^^^^^^^
TypeError: expected str, bytes or os.PathLike object, not int

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/rockstor/src/rockstor/rest_framework_custom/generic_view.py", line 41, in _handle_exception
    yield
  File "/opt/rockstor/src/rockstor/storageadmin/views/group.py", line 74, in post
    groupadd(groupname, gid)
  File "/opt/rockstor/src/rockstor/system/users.py", line 254, in groupadd
    return run_command(cmd)
           ^^^^^^^^^^^^^^^^
  File "/opt/rockstor/src/rockstor/system/osi.py", line 253, in run_command
    raise Exception("Exception while running command({}): {}".format(cmd, e))
Exception: Exception while running command(['/usr/sbin/groupadd', '-g', 1001, 'plex']): expected str, bytes or os.PathLike object, not int
```
 when leaving the field blank, i.e. allow AutoGenerate, this is not an issue. I assume, in that case the `-g` parameter is not passed and the OS assigns a new number.

Note, when creating a new user with a custom GID, this error does not occur.

FroggyFlox:
Nice find!
That looks like a Python3 requirement that was missed in the transition so far.so I've added this issue to our next Stable milestone.

Hooverdan96:
I suspect we might have some more candidates that take advantage of the run_command(cmd) or similar, where usually integers are passed instead of strings? ... Non-exhaustive list mind you:

like this:
https://github.com/rockstor/rockstor-core/blob/4d60a2c683140aaa10a8e5b201157163cc9a898b/src/rockstor/system/network.py#L344-L349

possibly this:
[...]

PR #2811, opened 2024-03-20
[t] Add Group with custom GID fails with type error #2807

phillxnet:
Add type hints to run_command() to ensure callers are passing the expected argument types.

Linking to #2807


phillxnet:
After the first commit: adding only type hints to run_command() we look to still have all tests passing:
```
Ran 279 tests in 30.053s

OK
```

phillxnet:
With this dev branch we now have functional 'Add Group' with for example a custom Web-UI entry of  GUID `1100`.

PR #2815, opened 2024-03-22
[t] Add Group with custom GID fails with type error #2807

### LINKED ISSUE #2807: [t] Assigning custom GID during User Group creation, results in syntax error

phillxnet:
Add type hints to run_command() to ensure callers are passing the expected argument types, and add some type hints &/or typecasts to run_command callers in:
- user.py
- active_directory.py
- qgourp_ids list to enforce str() members only in qgroup_maxout_limit.py.
- acl.py to enforce str() members of run_command() args.
- minor refactor of parameter name in run_command() as it shadowed another context.

Fixes #2807

### DIFF
--- a/src/rockstor/scripts/qgroup_maxout_limit.py
+++ b/src/rockstor/scripts/qgroup_maxout_limit.py
@@ -34,7 +34,7 @@ def main():
                 print("Quotas not enabled on pool(%s). Skipping it." % p.name)
                 continue
 
-            qgroup_ids = []
+            qgroup_ids: list[str] = []
             for l in o:
                 if (
                     re.match("qgroupid", l) is not None
--- a/src/rockstor/smart_manager/views/active_directory.py
+++ b/src/rockstor/smart_manager/views/active_directory.py
@@ -100,9 +100,9 @@ class ActiveDirectoryServiceView(BaseServiceDetailView):
                 self._resolve_check(config.get("domain"), request)
 
                 # 2. realm discover check?
-                domain = config.get("domain")
+                domain = str(config.get("domain"))
                 try:
-                    cmd = [REALM, "discover", "--name-only", domain]
+                    cmd: list[str] = [REALM, "discover", "--name-only", domain]
                     o, e, rc = run_command(cmd)
                 except Exception as e:
                     e_msg = (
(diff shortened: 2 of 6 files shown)


phillxnet:
# Testing
An rpmbuild resulted in the following single failure:

```
FAIL: test_post_requests (rockstor.storageadmin.tests.test_user.UserTests.test_post_requests)
  File "/home/buildbot/rockstor-buildbot/worker/Poetry-Build-on-Leap15-5/rpmbuild/rockstor-core-5.0.8-2815/src/rockstor/storageadmin/tests/test_user.py", line 195, in test_post_requests
    self.assertEqual(
AssertionError: 200 != 500 : {'id': 3, 'groupname': 'users', 'managed_user': True, 'has_pincard': False, 'pincard_allowed': 'no', 'username': 'newUser', 'uid': 3, 'gid': 5, 'public_key': None, 'shell': '/bin/bash', 'homedir': '/home/newUser', 'email': None, 'admin': True, 'user': 5, 'group': 1, 'smb_shares': []}
Ran 279 tests in 30.602s
FAILED (failures=1)
```

phillxnet:
# Testing 2

Post the proposed changes, an install of an rpm built using the proposed issue/pr branch no longer fails to create legitimate new groups with custom GID:

System - Groups - Add Group (button)


Groupname: PR2815
Put this group under Rockstor management? (unticked)
GID 3333

Custom-group-guid-creation

Resulting in the expected creation of this group with custom GID:

Groupname | GID | Actions
-- | -- | --
nobody | 65534 |  
nogroup | 65533 |  
PR2815 | 3333 |  

And we have the expected new group on the underlying OS:
```
rleap15-5:~ # cat /etc/group | grep PR2815
PR2815:x:3333:
```

PR #2818, opened 2024-03-26
Change Quota Status Display Wording #2810

### LINKED ISSUE #2810: Change "Disabled" to "Quotas Disabled" in Web-UI
We have the following header & example content in shares overview:

| `Pool (Active mount options / Status) Quotas` |
| -- |
| pool-name (rw,relatime,space_cache,subvolid=5,subvol=/) Enabled |

Which is meant to indicate the 3 infomational items: Pool (name), mount options/status, Quota status.

However when the RED Disabled appears in the 'Quotas' position:

| `Pool (Active mount options / Status) Quotas` |
| -- |
| pool-name (rw,relatime,space_cache,subvolid=5,subvol=/) Disabled |

 it can too easily lead to an inaccurate/misleading inference that the Share itself is disabled.

It is proposed that we instead specify that it is the Quotas that are disabled by the explicit use of **Quotas Disabled** in this warning message text.

Hooverdan96:
Fixes #2810.

update to shares_table.jst

Testing:
Changed wording (including Quotas to clarify what this status refers to) shows up on screen now:

image

For good measure, also ran test suite.

```
# poetry run django-admin test -v 2

Ran 279 tests in 46.885s

OK
```

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/share/shares_table.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/share/shares_table.jst
@@ -41,9 +41,9 @@
                 (<strong><span style="color:red">{{this.pool.mount_status}}</span></strong>)
             {{/if}}
             {{# if this.pool.quotas_enabled}}
-                Enabled
+                Quotas Enabled
             {{else}}
-                <strong><span style="color:red">Disabled</span></strong>
+                <strong><span style="color:red">Quotas Disabled</span></strong>
             {{/if}}
         </td>
         <td>


FroggyFlox:
Probably off topic here, but what do you guys think about the red font used here? I wonder if that could lead some to think it's an actual error more than a setting for the share/pool.

phillxnet:
Reviewing this now: as per colour change: lets address that in it's own issue as there are other locations affected by that. Here I think we just need for the warning (ergo red) to be more clear on what it is warning about. And this more explicit wording does just that I think/hope.

phillxnet (approved):
@Hooverdan96 Thanks for seeing to this change: much appreciated.

When I originally added this wording I had not appreciated it's potential to mislead via its brevity; and we now have longer cell entries in this column as a result: but better over all and with some improvements in our tables we can always reclaim some of the wasted space in many of the other columns.

With an rpm built including this fix we have the expected outcome of better wording, after having refreshed browser cache that is:

quotas-disabled-wording-change-pr


Cheers, I'll go ahead and get this merged ready for our next testing rpm.

Hooverdan96:
Just as a sidenote, if we choose to do away with the red color **just** in this instance (and/or the bolding), it's in the same line as what this PR's change did:

do away with the tag around the text:

```
<strong><span style="color:red">Quotas Disabled</span></strong>
```

phillxnet:
@Hooverdan96 
Re:

> if we choose to do away with the red color **just** in this instance

In the context that we show folks data as 0 bytes when quotas are disabled: I'm very much inclined to still treat this as a 'red' condition. I now we have grown accustom to quotas disabled but I think we need to move towards actually enforcing them soon. And we do still have quotas enabled as a default. We have after all just seen on the forum a discussion where folks were to delete shares as no data, i.e. zero bytes, was occupied: ergo zero loss - or data already gone. This was the motivation for me creating the issue. To emphasis that it is only the quotas that was disabled.

We do have the mouse-over explaining the zero space reading with quotas disabled - but bit-by-bit. Maybe a parallel UI improvement here is to replace the zero (when quotas disabled) with something like "unknown"  with mouse-over to indicate quotas disabled. But this is all for the next testing phase I think, as the hope is we can make such changes after updating/easing our Web-UI out from under near raw JS: and have more live elements.

PR #2822, opened 2024-03-27
Adapt net interface delete to 'rockstor' service null config #2819

### LINKED ISSUE #2819: Adapt net interface delete to 'rockstor' service null config default
Almost all services have a default config of null: including the Web-UI service known as 'rockstor'. However when deleting an existing interface, with a default null config for the 'rockstor' service: we get the following log entry indicating a silent failure in our fencing to guard against inadvertent removal of the Web-UI associated 'management' interface.

```
[26/Mar/2024 14:56:39] ERROR [storageadmin.views.network:564] the JSON object must be str, bytes or bytearray, not NoneType
  File "/opt/rockstor/src/rockstor/storageadmin/views/network.py", line 560, in delete
    config = json.loads(so.config)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/json/__init__.py", line 339, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not NoneType

```

---

See the following reproducer comment that arouse while testing a recent fix for network interface removal via Web-UI:
https://github.com/rockstor/rockstor-core/pull/2817#issuecomment-2020679498

And the following contextual issues/design directions:
- Ability to reset services to original status/settings #2332
- Failure to start Shellinabox from System or breadcumbs links if unconfigured #2773 where the following comment indicates a related inelegance (not not a silent failure) relating to default null config: https://github.com/rockstor/rockstor-core/issues/2773#issuecomment-1872170961

phillxnet:
Block non-rocknet interface deletions for the default/no management/Web-UI interface configuration. Raising a user visible exception explaining the interface delete rejection, and advising the need to configure the 'rockstor' service first. This way, for default/no 'rockstor'/Web-UI service configurations (no management interface defined) we avoid inadvertent Web-UI outage. Once a non default 'rockstor' service/Web-UI Network interface is established, only that interface will be bared from delete.

Fixes Web-UI silent but logged failure to properly process/fence an interface delete request with default 'rockstor'/Web-UI configuration.

Fixes #2819

### DIFF
--- a/src/rockstor/storageadmin/views/network.py
+++ b/src/rockstor/storageadmin/views/network.py
@@ -554,21 +554,32 @@ class NetworkConnectionDetailView(rfc.GenericView, NetworkMixin):
                         dnet_disconnect(c, brco.docker_name)
                 dnet_remove(network=brco.docker_name)
             else:
-                restricted = False
+                restricted: bool = False
+                unknown_restricted: bool = False
                 try:
                     so = Service.objects.get(name="rockstor")
-                    config = json.loads(so.config)
-                    if config["network_interface"] == nco.name:
-                        restricted = True
+                    if so.config is None:
+                        unknown_restricted = True
+                    else:
+                        config = json.loads(so.config)
+                        if config["network_interface"] == nco.name:
+                            restricted = True
                 except Exception as e:
                     logger.exception(e)
                 if restricted:
                     e_msg = (
-                        "This connection ({}) is designated for "
-                        "management and cannot be deleted. If you really "
+                        f"This connection ({nco.name}) is designated for "
+                        "management/Web-UI and cannot be deleted. If you really "
                         "need to delete it, change the Rockstor service "
                         "configuration and try again."
-                    ).format(nco.name)
+                    )
+                    handle_exception(Exception(e_msg), request)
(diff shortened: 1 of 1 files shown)


phillxnet:
# Testing

See issue for pre patch behaviour: a silent failure in our fencing to guard against inadvertent removal of the Web-UI associated 'management' interface; but only when a management interface has yet to be defined: i.e. default 'rockstor' service configuration.

**Resulting in Web-UI inaccessibility** and the requirement to re-connect that interface via, for example, `nmcli dev connet eth0`.

After the proposed patch, for a default configuration of the 'rockstor'/Web-UI network interface, we disable/block all non rocknet interface deletions with the following Web-UI/user-visible exception:

delete-only-interface-null-rockstor-service-config

On the grounds that we have no configuration to advise us otherwise.

Once a non-default 'rockstor' service config has been established: i.e. an interface for the Web-UI has been selected, this interface only is excluded from all deletion requests.

delete-WebUI-interface-as-per-rockstor-service-config

Noting that we can, with a Web-UI config, now delete other non-Web-UI associated interfaces as expected.

Issue #2824, opened 2024-03-29
Un special-case system drive btrfs-in-partition treatment


phillxnet:
In the olden days, pre:
- "improve root not in partition behaviour and mdraid member labeling. Fixes #1164" #1166
- "enhance disk role/management subsystem. Fixes #1494" #1622
- "support full disk luks. Fixes #550" #1716

we managed only whole disk btrfs, and **special cased** our system drive as it was partitioned. This is no longer necessary and represents an inconsistency that adds complexity - bringing with it a maintenance burden.

It is proposed that we leverage the work in the interim years re our data dives to apply similarly to our system drive (drives in the future). With the aim of removing legacy special treatment that was originally intended solely to present a partitioned system pool member as if it was a whole disk: the only entity we understood at that time.

I.e. our low-level internal representation from system.osi.scan_disks() for the system disk is presented below.
The following are older real-system test data from when our OS pool was labelled `rockstor_rockstor` but otherwise similar.
Test reference: https://github.com/rockstor/rockstor-core/blob/4d60a2c683140aaa10a8e5b201157163cc9a898b/src/rockstor/system/tests/test_osi.py#L1072-L1076

```
Disk(
                    name="/dev/sda3",
                    model="QEMU HARDDISK",
                    serial="QM00005",
                    size=7025459,
                    transport="sata",
                    vendor="ATA",
                    hctl="2:0:0:0",
                    type="part",
                    fstype="btrfs",
                    label="rockstor_rockstor",
                    uuid="355f53a4-24e1-465e-95f3-7c422898f542",
                    parted=True,
                    root=True,
                    partitions={},
                )
```
Where-as the equivalent btrfs-in-partition pool member would be represented as follows:
```
                # Note partitions entry within vda, consistent with cli prep.
                Disk(
                    name="/dev/vda",
                    model=None,
                    serial="serial-1",
                    size=4194304,
                    transport=None,
                    vendor="0x1af4",
                    hctl=None,
                    type="disk",
                    fstype="btrfs",
                    label="btrfs-in-partition",
                    uuid="55284332-af66-4ca0-9647-99d9afbe0ec5",
                    parted=True,
                    root=False,
                    partitions={"/dev/vda1": "vfat", "/dev/vda2": "btrfs"},
                )
```

[...]

Issue #2749, opened 2023-11-15
Disk miss-attribution to ROOT pool


ubenmackin:
Hello again!

I was going to make a forum post, but for some reason was having issues loading the forum.

At any rate, I have a strange issue. My system has a lot of disks (currently 35) spread across a number of pools. This has caused my drive disk labels to cycle from sda to sdz, and then to sdaa, sdab, sdac, etc.

Recently I noticed that my ROOT pool, which consists of a single disk was showing with two disks in the Pool viewer. Also, when looking at the Disks page, the ROOT pool shows with two disks, which is incorrect.

<img width="1205" alt="image" src="https://github.com/rockstor/rockstor-core/assets/11615536/9a39318f-2f2a-4722-ae62-3b5ce6b42a75">

Not only is the pool listed wrong, but so is the Capacity. This is a 10 TB disk, but it is being show on the disk page as only 109 GB, the size of my ROOT disk.

Logging into the shell, and doing a

`btrfs fi show`

Shows everything correctly. So it seems to be just a display issue in Rockstor.

I am guessing the issue is due to the drive lettering. Here is my ROOT pool:

```
rocknas:~ # btrfs fi show /
Label: 'ROOT'  uuid: 4ac51b0f-afeb-4946-aad1-975a2a26c941
	Total devices 1 FS bytes used 19.03GiB
	devid    1 size 109.72GiB used 37.60GiB path /dev/sdag4
```

And here is the pool that is having a disk mislabeled:

```
rocknas:~ # btrfs fi show /mnt2/JBOD
Label: 'JBOD'  uuid: b86538f8-e447-48e5-84ec-b72a25c65282
	Total devices 10 FS bytes used 32.06TiB
	devid    1 size 7.28TiB used 6.32TiB path /dev/sdl
	devid    2 size 9.10TiB used 6.98TiB path /dev/sda
	devid    3 size 7.28TiB used 6.93TiB path /dev/sdb
	devid    4 size 7.28TiB used 6.93TiB path /dev/sde
	devid    5 size 7.28TiB used 6.89TiB path /dev/sdn
	devid    6 size 7.28TiB used 6.89TiB path /dev/sdm
	devid    7 size 7.28TiB used 6.89TiB path /dev/sdo
	devid    8 size 7.28TiB used 6.89TiB path /dev/sdk
	devid    9 size 7.28TiB used 6.89TiB path /dev/sdp
[...]

PR #2825, opened 2024-03-29
Un special-case system drive btrfs-in-partition treatment #2824

phillxnet:
No functional change on first commit. Some basic code improvements to clear the way for the planed simplification/modification later in this draft PR.

Includes:
- Modernisation re type hints in scan_disks().
- Programmatically establish `lsblk` field derived Disk namedtuple fields.
- Refactoring for readability.
- Black reformatting.

See issue text #2824

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -302,8 +302,8 @@ def scan_disks(min_size: int, test_mode: bool = False) -> list[Disk]:
         "NAME,MODEL,SERIAL,SIZE,TRAN,VENDOR,HCTL,TYPE,FSTYPE,LABEL,UUID",
     ]
     o, e, rc = run_command(cmd)
-    dnames: dict = {}  # Working dictionary of devices.
-    disks: list[Disk] = []  # List derived from the final working dictionary of devices.
+    # Working dictionary of Disk values: indexed by a copy of their .name field.
+    dnames: dict = {}
     serials_seen: list[str] = []  # List tally of serials seen during this scan.
     # Stash variables to pass base info on root_disk to root device proper.
     root_serial: str | None = None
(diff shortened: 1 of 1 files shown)


src/rockstor/system/osi.py
```diff
@@ -108,12 +108,25 @@
     "0123456789ABCDEF",  # No-name USB external multi-bay.
 ]
 
-Disk = collections.namedtuple(
-    "Disk",
-    "name model serial size transport vendor "
-    "hctl type fstype label uuid parted root "
-    "partitions",
-)
+# Requried to assist parsing of lsblk output to Disk field names.
```
Hooverdan96, line 111: Small typo:
# Required to assist parsing of lsblk output to Disk field names.

src/rockstor/system/osi.py
```diff
@@ -265,7 +278,20 @@ def run_command(
     return out, err, rc
 
 
-def scan_disks(min_size, test_mode=False):
+# def lsblk_line_parse(line: str = "") -> Disk:
+#     """
+#     Parse single lsblk output line into a Disk named tupile
```
Hooverdan96, line 283: small typo:
```
# Parse single lsblk output line into a Disk named tuple
```

phillxnet:
@Hooverdan96 Thanks for the review: but this is very much a draft still: I've not actually begun the functional changes. And I'll have to squash these and many other changes to come so your contributions/attribution will lost.

But thanks for the typos anyway. I'll incorporate when next working on this draft: soon hopefully. Pushing my work as I go here though just in case - and easier to ask advice etc if I get stuck. Or do end-to-end testing via our back-end rpmbuild against PRs (draft or otherwise). I think I have a few simplifications on the way, and hopefully the end result will be more maintainable. Pretty sure we can simplify quite a bit: as what we have currently is definitely super stretched. I also plan to make more use of our Disk namedtuple in scan_disks() so the code can be more self documenting, and to help guard against accidental changes we don't intend: named tuples are immutable.

Hooverdan96:
Understood absolutely, and I'm not craving attribution for those cosmetic things, but figured if I mention them now, you can get rid of that, especially since you will now start focusing on making actual functional changes...

Godspeed on the simplification, and definitely makes sense to me if you will be able to put a more complete dataset/logic on the `scan_disks()` to simplify subsequent functions and overall usage for other pieces.

phillxnet:
At this point in this draft PR associated branch all existing osi.py tests again pass. And we have maintained the existing observed failure in the 2 new tests associated with draft PR: "Disk being misattributed to ROOT pool #2749" #2823.

So we are still at no functional change: as intended. But the code is now likely more robust and easier to read with many more type hints and way more obvious re internal changes and intent within scan_disks(). Next:

- [ ]  a little more tidy, and give the increased readability we can likely do away with many of the explanatory comments to further readability.
- [ ] modify all existing tests to expect the issue associated behaviour change.
- [ ] approach the functional changes in scan_disks() to return all tests to passing again.
- [ ] ensure we have accounted for and fixed the failure detailed in the 2 new tests associated with #2749 & #2823

phillxnet:
This Draft PR's work-to-date represents a significant enough modification to address a stepping stone defined by the following issue: Modernise scan_disks() - no functional change intended #2826 for which a new PR will be presented shorlty.

PR #2827, opened 2024-04-03
Modernise scan_disks() - no functional change intended #2826

### LINKED ISSUE #2826: Modernise scan_disks() - no functional change intended
Some basic code quality/readability improvements to assist with planned functional changes associated with the following issues:

- Un special-case system drive btrfs-in-partition treatment #2824
- Disk being misattributed to ROOT pool #2749

phillxnet:
By working directly with the Disk namedtuple during device collation, we gain readability and the advantages of an immutable type: i.e. we are then more explicit where writes/changes happen due to the requirement to use Disk._replace(). We also require less type transitions overall.

Includes:
- Modernisation re type hints in scan_disks().
- Improve lsblk line parser via modern Python builtins.
Replaces by-hand lsblk output line parser with dictionary comprehension.
Programmatically establishing `lsblk` field derived Disk namedtuple fields.
- Refactoring for readability.
- Employ defaults in Disk named tuple to ease creation using lsblk info, as these defaults pertain to our extra flag info.
- Normalise scan_disks()'s working dict keys to lowercase. By normalising on lowercase we can ease the transition to using our Disk named tuple as the working copy: to help readability and highlight where we write (via Disk.replace(root=True)). This should also simplify the assembly of our final return value. We also have a large number of tests that use the existing lowercase Disk fields.
- Update comments re now more readable code.
- Black reformatting.

Fixes #2826

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -102,17 +102,33 @@ EXCLUDED_MOUNT_DEVS = [
 # 5 bay no-name USB - forum member Miyuki report.
 
 EXCLUDED_SERIAL_NUMS = [
+    None,
     "",
     "000000000000",  # Many reports of multiple Orico models.
     "152D00539000",  # USB ID 152d:0567 a JMS567 based device.
     "0123456789ABCDEF",  # No-name USB external multi-bay.
 ]
 
+# Field_names correspond to all used lsblk properties.lower() bar 'TRANS' to 'transport'
 Disk = collections.namedtuple(
     "Disk",
-    "name model serial size transport vendor "
-    "hctl type fstype label uuid parted root "
-    "partitions",
+    [
+        "name",
+        "model",
+        "serial",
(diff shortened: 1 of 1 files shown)


phillxnet:
# Testing
An rpm was successfully build with these changes, ensuring all existing tests passed. The resulting Rockstor instance worked as expected regarding disk presentation. And two example pools were successfully imported and interpreted.

N.B. this PR is a stepping stone to contain/present the interim no function change intended improvements as per the linked issue text.

phillxnet:
@FroggyFlox & @Hooverdan96 Merging as a dependency / prerequisite to my ongoing work towards:

- Un special-case system drive btrfs-in-partition treatment #2824
- Disk being misattributed to ROOT pool #2749

As per linked issue text. This should then help with future assessments/understanding of why some changes were made to this rather sensitive and low-level piece of our puzzle.

PR #2857, opened 2024-06-22
Leap 15.6: SFTP share error - library paths changed #2856

### LINKED ISSUE #2856: Leap 15.6: SFTP share error - library paths changed
Thanks to forum user [mpidala](https://forum.rockstor.com/u/mpidala/summary), the following issue was reported during `sftp` share creation on a Leap 15.6 Rockstor instance (https://forum.rockstor.com/t/error-adding-sftp-share/9548/):
the following error message is thrown when trying to create an sftp share:

```
  File "/opt/rockstor/src/rockstor/rest_framework_custom/generic_view.py", line 41, in _handle_exception
    yield
  File "/opt/rockstor/src/rockstor/storageadmin/views/sftp.py", line 82, in post
    rsync_for_sftp(chroot_loc)
  File "/opt/rockstor/src/rockstor/system/ssh.py", line 303, in rsync_for_sftp
    copy(l, "{}{}".format(chroot_loc, l))
  File "/usr/lib64/python3.11/shutil.py", line 431, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib64/python3.11/shutil.py", line 256, in copyfile
    with open(src, 'rb') as fsrc:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/lib64/libz.so.1'

```

As @FroggyFlox already identified, some of the shared libraries that are symbolically linked during the sftp process can't be found because their physical location has changed with the Leap 15.6 release. Within the Rockstor code base they are maintained here:

https://github.com/rockstor/rockstor-core/blob/3f8d02b61f38c0a906423b6a371d17c837f52299/src/rockstor/system/ssh.py#L273-L285

the one that triggered the above error is assumed to be located in:
`/lib64/libz.so.1`

In Leap 15.6 the location has changed to:
`/usr/lib64/libz.so.1`

This likely has to be updated. The decision here is whether to remain backwards compatible to at least Leap 15.5 or not.

If no, then an update of the paths in the above code snippet should be sufficient.
If yes, we will need an additional way of identifying which `Leap` release is currently being used during the execution of the program.

Unfortunately, in the copying process only the distribution ID (`OS_DISTRO_ID`) from the `settings file` is used to validate:

https://github.com/rockstor/rockstor-core/blob/3f8d02b61f38c0a906423b6a371d17c837f52299/src/rockstor/system/ssh.py#L302-L304

As one can see, in the settings file the values for `OS_DISTRO_ID` do not distinguish by distribution release

https://github.com/rockstor/rockstor-core/blob/3f8d02b61f38c0a906423b6a371d17c837f52299/src/rockstor/settings.py#L466-L473

Unfortunately, just replacing the `OS_DISTRO_ID` with 'OS_DISTRO_VERSION' would likely break the "rolling" Tumbleweed version in this case. Since the version under Tumbleweed changes with every release which are very frequent (daily/weekly). Updating Rockstor that frequently is neither practical nor sustainable.

One approach could be to add another condition that essentially checks whether the Rockstor instance runs on Leap, and if so, check whether it's version 15.6 or earlier. If earlier the current definition shown above would be used, and if 15.6 (and presumably later) select a newly created section that contains the altered paths to create symbolic links from. If 15.6 or higher, concatenate ID and Version and the section the copy process refers would also have the concatenated version (e.g. `opensuse-leap15.6+` and for other leap flavors it remains what it was, `opensuse-leap`). That way, other Leap flavors are not impacted. Once support for 15.5 is dropped, that section can be dropped, and it can be similarly managed if a change comes with a `15.7` (if ever released) version ...

Or a completely different approach, of course.

Finally, I am surprised that the testing library would not have caught this, but I did not take a closer look at the unit tests. When fixing this, the unit test should probably be corrected, or if it doesn't exist should be created, since I suspect this will not be the last time that the location of these types of files will change. For reference:

https://github.com/rockstor/rockstor-core/blob/testing/src/rockstor/storageadmin/tests/test_sftp.py

happy path (which should be unhappy on 15.6 in my opinion):
https://github.com/rockstor/rockstor-core/blob/3f8d02b61f38c0a906423b6a371d17c837f52299/src/rockstor/storageadmin/tests/test_sftp.py#L158-L165

phillxnet:
Move to dynamic discovery of bash & rsync libraries when populating ssh / sftp share chroot.

Fixes #2856

### DIFF
--- a/src/rockstor/system/constants.py
+++ b/src/rockstor/system/constants.py
@@ -28,6 +28,7 @@ MOUNT = "/usr/bin/mount"
 UMOUNT = "/usr/bin/umount"
 
 USERMOD = "/usr/sbin/usermod"
+LDD = "/usr/bin/ldd"
 
 SYSTEMCTL = "/usr/bin/systemctl"
 
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -2006,7 +2006,7 @@ def get_libs(program_path: str) -> list[str]:
             line_fields: list = each_line.strip().split()
             match len(line_fields):
                 case 2:
-                    if re.match("linux-vdso", each_line) is not None:
+                    if re.match("linux-vdso", line_fields[0]) is not None:
                         continue
                     else:
                         libs.append(line_fields[0])
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -45,6 +45,7 @@ from system.constants import (
     DEFAULT_MNT_DIR,
     UDEVADM,
     SHUTDOWN,
+    LDD,
 )
 
 logger = logging.getLogger(__name__)
@@ -1990,6 +1991,28 @@ def get_devname(device_name, addPath=False):
             return fields[0]
     # a non one word reply was received on the first line from udevadm or
     return None
+
+
+def get_libs(program_path: str) -> list[str]:
+    """
+    Wrapper around `ldd program_path` 
+    @param program_path: Binary to query ldd about 
+    @return: list of OS paths or empty list if error or non found.
+    """
+    libs: list[str] = []
+    out, err, rc = run_command([LDD, f"{program_path}"], throw=False)
+    if len(out) > 0 and rc == 0:
+        for each_line in out:
+            line_fields: list = each_line.strip().split()
+            match len(line_fields):
+                case 2:
+                    if re.match("linux-vdso", each_line) is not None:
+                        continue
+                    else:
+                        libs.append(line_fields[0])
+                case 4:
+                    libs.append(line_fields[2])
+    return libs
 
 
 def update_hdparm_service(hdparm_command_list, comment):
--- a/src/rockstor/system/ssh.py
+++ b/src/rockstor/system/ssh.py
@@ -28,7 +28,7 @@ from tempfile import mkstemp
 import distro
 from django.conf import settings
 
-from system.osi import run_command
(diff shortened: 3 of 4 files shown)


phillxnet:
With this branch prior to setting up a sftp export we have the following chroot location:
```
rleap15-6:~ # tree /mnt3/sftp-user/
/mnt3/sftp-user/ [error opening dir]

0 directories, 0 files
```
I.e. it doesn't exist.

But creating a sftp export  of user owned `sftp-test-share` (along with our minimal chroot with bash and rsync) and we have the following created:

```
rleap15-6:~ # tree /mnt3/sftp-user/
/mnt3/sftp-user/
├── lib64
│   ├── ld-linux-x86-64.so.2
│   ├── libacl.so.1
│   ├── libattr.so.1
│   ├── libc.so.6
│   ├── libdl.so.2
│   ├── libreadline.so.7
│   └── libtinfo.so.6
├── sftp-test-share
└── usr
    ├── bin
    │   ├── bash
    │   └── rsync
    └── lib64
        ├── libcrypto.so.3
        ├── liblz4.so.1
        ├── libpopt.so.0
        ├── libslp.so.1
        ├── libz.so.1
        └── libzstd.so.1

5 directories, 15 files
```

This draft PR will under-go a little more testing of the resulting rpms and then be squashed and represented as this looks to have solved our #2857 issue with a little more adaptability than what we had before.

PR #2867, opened 2024-07-10
[t] Indicate Tailscale install requirement #2845

### LINKED ISSUE #2845: [t] Indicate Tailscale install requirement
Our attempt at having the Tailscale repos and its package pre-installed in our installer has unfortunately had to be stopped: https://github.com/rockstor/rockstor-installer/pull/173. This means we now need to ship our installer *without* the repo and package needed for the included Tailscale service (https://rockstor.com/docs/interface/system/services.html#tailscale).

We unfortunately do not surface the need for the user to first install the repo and the Tailscale package before configuring the service. This might have already misled a Testing channel user, unfortunately: [https://forum.rockstor.com/t/unknown-internal-error-doing-a-post-to-api-sm-services-tailscaled-config/9418/2](https://forum.rockstor.com/t/unknown-internal-error-doing-a-post-to-api-sm-services-tailscaled-config/9418/2?u=flox).

We should thus, to begin, surface this information to the user trying to configure the Tailscale service. This could be:
- 1: check that the tailscale repo is present on the system and show a banner in the configure pop-up if it isn't, along with disabling the "submit" button with explanation.
- 2: check that the tailscale repo is present on the system and if it isn't: disable the "configure" button and show some sort of warning icon or the like in the Services page itself next to the Tailscale repo.
- 3: simple banner reminding the user to install the package first, along with a link to our docs showing how to.

Any other idea(s) is welcome, of course.

phillxnet:
Introduces `tailscale` binary os.path.isfile sensitivity to inform the associated Web-UI errors: enabling more context regarding our NOT pre-installing tailscale.

Fixes #2845

### DIFF
--- a/src/rockstor/smart_manager/views/tailscaled_service.py
+++ b/src/rockstor/smart_manager/views/tailscaled_service.py
@@ -25,6 +25,7 @@ from smart_manager.models import Service
 from smart_manager.views.base_service import BaseServiceDetailView
 from storageadmin.exceptions import RockStorAPIException
 from storageadmin.util import handle_exception
+from system.exceptions import CommandException
 from system.services import systemctl
 from system.tailscale import (
     tailscale_up,
@@ -32,6 +33,7 @@ from system.tailscale import (
     get_ts_auth_url,
     tailscale_down,
     validate_ts_hostname,
+    TS_NOT_INSTALLED_E_MSG,
 )
 
 logger = logging.getLogger(__name__)
@@ -95,9 +97,20 @@ class TailscaledServiceView(BaseServiceDetailView):
             # to use it so we enable and start the tailscaled daemon.
             # This will create the tailscale tun device and allow us to get
             # an authURL during the "login" action
-            systemctl(self.name, "stop")
-            systemctl(self.name, "enable")
-            systemctl(self.name, "start")
+            try:
+                systemctl(self.name, "stop")
+                systemctl(self.name, "enable")
+                systemctl(self.name, "start")
+            except Exception as e:
+                logger.exception(e)
+                e_msg = "An error occurred with tailscaled.service."
+                if e.__class__ == CommandException:
+                    if e.err == [
+                        "Failed to stop tailscaled.service: Unit tailscaled.service not loaded.",
+                        "",
+                    ]:
+                        e_msg = TS_NOT_INSTALLED_E_MSG
+                handle_exception(Exception(e_msg), request)
 
         elif command == "start":
             config = None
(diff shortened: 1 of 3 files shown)


phillxnet:
# Testing

Behaviour appear as-was with a "Tailscale install" as per our recent how-to: https://rockstor.com/docs/howtos/tailscale_install.html

phillxnet:
> N.B. I was expecting a FileNotFound in one of the above. Looking into this now.

In order to trigger this we look to the code comment:

>            # Account for install with restored Tailscale config, but no Tailscale installed.

With a fresh install there is no config: and we error out with the prior exception when initially trying to adding one (with no Tailscale install) as per the last comments details.

1. Install Tailscale as per How-to in error message: https://rockstor.com/docs/howtos/tailscale_install.html
2. Tailscale spanner - submit config defaults: On/Off replaced by `Login` button.
3. `zypper --non-interactive remove tailscale`
4. Tailscale "Login" button is replaced by On/OFF (auto-refresh active on services page).
5. Switch service On (not blocked by config due to above arrangements) - but we trigger the missing `tailscale` binary fence: FileNotFoundError.

```
Houston, we've had a problem.
Tailscale is NOT installed by default. See docs How-tos & guides: `Tailscale install` https://rockstor.com/docs/howtos/tailscale_install.html

  File "/opt/rockstor/src/rockstor/smart_manager/views/tailscaled_service.py", line 130, in post
    tailscale_up(config=config)
  File "/opt/rockstor/src/rockstor/system/tailscale.py", line 69, in tailscale_up
    tailscale_not_found()
  File "/opt/rockstor/src/rockstor/system/tailscale.py", line 58, in tailscale_not_found
    raise FileNotFoundError
FileNotFoundError
```

phillxnet:
An rpm originating install was setup on both a 15.5 and 15.6 (both X86_64) install and both were Web-UI configured & enabled to participate in a Tailscale network. They were then able to ping one-another using the Tailscale assigned IPs. Serving as a test to ensure we still have successful Tailscale default config / enable / Login capability via the Web-UI only.

Thanks again to @FroggyFlox for this Tailscale UI implementation/addition. I'm not entirely happy with my modifications here: but they serve an informational purpose which was the aim of the linked issue. We can re-address the in-elegance of only catching exceptions here  once we have begun our HTMX 2.0 introduction and enabled SSEs to more interactively/proactively indicate a service's installed state. Enabling far smaller install options in the future. But for now we needed to inform folks of this services non-default install status: our first service of this kind.

FroggyFlox (approved):
@phillxnet,
I won't have time to test it myself but your extensive testing and report make me confident it's all good!
I do agreed it's still not the best user experience but it is definitely a big improvement over the current situation and it lays the foundation for a much better and dynamic user experience down the road once we have a proper messaging/notification system in place.

Thanks for taking care of this one too!

PR #2873, opened 2024-07-15
Scrub status "unknown" - Leap 15.6 OS base #2872

### LINKED ISSUE #2872: Scrub status "unknown" - Leap 15.6 OS base
Thanks to forum member Tex1954 for highlighting this issue. We currently report "unknown" for scrub status on a 15.6 base, but work as expected on a 15.5 base. Thanks to @FroggyFlox & @Hooverdan96 for helping to narrow down this issue.

N.B. @Hooverdan96 reports that a Stable Backport kernel does not exibity this error, this instance would also have a newer btrfs-progs userland as per: https://rockstor.com/docs/howtos/stable_kernel_backport.html

This issue is currently believed to be cosmetic in nature as the scrub itself is executed as expected. However cosmetic failure, in this context, translates to a Web-UI failure in a critical report: adding to our current Stable Milestone as a consequence.

- [x] Establish unit-test reproducer
- [x] Establish fix in the context of our existing tests and the new reproducer test.

---

Forum thread reference: https://forum.rockstor.com/t/5-0-11-scrub-not-working/9585

Seen last in 5.0.12-0

phillxnet:
Adopt packaging library as a direct dependency to enable more sophisticated btrfs-progs version assessment. Previously already an indirect dependency of gunicorn. Moving to explicit dependency to avoid issues re gunicorn's deprecation in the future.

Includes
- Additional test data pertaining to Leap 15.6 and newer, re btrfs-progs version.
- Additional test data re exception handling (fail through) re non PEP 440 compliant 'btrfs-progs version' output.

Fixes #2872

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ psycopg = "~3"
 psycogreen = "==1.0"
 gevent = "*"  # can be an extra dependency to gunicorn.
 gunicorn = "*"
+packaging = "*"  # 'btrfs version' parsing.
 
 # [tool.poetry.group.tools.dependencies]
 six = "==1.16.0"  # 1.14.0 (15 Jan 2020) Python 2/3 compat lib
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -38,6 +38,7 @@ from huey.contrib.djhuey import task
 from django.conf import settings
 import logging
 from datetime import datetime
+from packaging.version import Version, InvalidVersion
 
 """
 system level helper methods to interact with the btrfs filesystem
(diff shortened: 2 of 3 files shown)


phillxnet:
# Testing
On an issue-reproducer system (dev machine using Leap 15.6 X86_64) we have one before (see: https://github.com/rockstor/rockstor-core/issues/2872#issuecomment-2228515648) and one after scrub status reports:

finished-scrub-status-leap-15 6

FroggyFlox (approved):
Thanks @phillxnet for this proper fix!
I can confirm it fixes the issue in Leap 15.6 on my end too; Leap 15.5 is still working as intended as well.
All tests also pass on my Leap 15.6 VM.

phillxnet:
@FroggyFlox Thanks for the review & 15.5 check: much appreciated. I'll get this merged directly.

I know we have the question of need, for this legacy btrfs-progs support, as per https://github.com/rockstor/rockstor-core/issues/2872#issuecomment-2228544443: i.e. do we really need to support pre Leap 15.4 btrfs-progs versions. But we are in very late testing here so minimal changes to address indicated issue. We also have the potential to migrate scrub to Huey (or a more Django native Huey replacement) but our consensus appears to be that this is better tackled in the next or next-plus-one testing phase.

phillxnet:
An rpm was test built using this branch and looks to behave as-per the testing report so far. We also have the new dependency here which now shows up as a new top-level one re:
```
cd /opt/rockstor/
poetry show --tree
...
gunicorn 22.0.0 WSGI HTTP Server for UNIX
└── packaging *
...
packaging 24.1 Core utilities for Python packages
psutil 5.9.4 Cross-platform lib for process and system monitoring in Python.
...
```
Where-as previously we only saw the `packaging` dependency as secondary: i.e. via gunicorn as in the above.

PR #2882, opened 2024-07-26
Block creation of system reserved Share names #2881

### LINKED ISSUE #2881: Block creation of system reserved Share names
Thanks to forum member andyz for highlighting this issue. We have a hard limitation regarding unique btrfs vol (Pool) and btrfs subvol (Share) naming. But prior to (current testing phase):

Un special-case system drive btrfs-in-partition treatment #2824 #2835

we always auto imported the `ROOT` system Pool (btrfs vol), with it's 'special' `home` Share (btrfs subvol). And similarly identified this `home` Share by name as 'special': limiting the available functions as a result. But we are now missing a block against folks creating a `home` Share which would prior to the referenced changes always already exist. So our existing no-duplicate Share/Pool names block avoided the duplication.

Ergo we now need to guard against folks creating a `home`  Share via the Web-UI as this would then cause a potential clash if, at a later date, folks wanted to import the system Pool. It would also introduce a source of confusion re two subvols named the same - which we normally avoid via our existing hard limitation.

---

Forum thread reference: https://forum.rockstor.com/t/newly-created-test-share-named-home-cannot-be-resized-nor-deleted/9620

phillxnet:
Use/extend existing knowledge of system associated subvolumes to avoid breaking (optional) ROOT pool
import, and non-optional system unique share names.

Includes:
- Addition of CREATE_SUBVOL_EXCLUDE list.
- Block creation of all currently EXCLUDED subvol names, previously used only to avoid Web-UI surfacing.
- Incidental string.format to fstrings in share.py.
- Additional unit test to exercise added Share creation filter.

Fixes #2881

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -79,8 +79,14 @@ ROOT_SUBVOL_EXCLUDE = [
     "@/.snapshots",
     ".snapshots",
 ]
-# Note in the above we have a non symmetrical exclusions entry of '@/.snapshots
+# Note in the above we have a non-symmetrical exclusions entry of '@/.snapshots
 # this is to help distinguish our .snapshots from snapper's rollback subvol.
+
+# Create subvolume blacklist to avoid name clash with default ROOT pool.
+# Used in addition to ROOT_SUBVOL_EXCLUDE. From 5.0.9-0 we no longer auto-import
+# the system (ROOT) pool.
+CREATE_SUBVOL_EXCLUDE = ["home", "@/home"]
+
 # System-wide subvolume exclude list.
 SUBVOL_EXCLUDE = [".beeshome", "@/.beeshome"]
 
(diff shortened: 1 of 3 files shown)


phillxnet:
N.B. this new blocking mechanism is not intended to affect imports: only Web-UI Share creation.  Otherwise we would be unable to import our 'ROOT' Pool. An expectation from our past that is maintained, but no longer default.

Of all current upstream OS (openSUSE Leap/TW) subvols created by our kiwi-ng installer, we only surface `/home` within the Web-UI. All others are filtered out as a simplification/safety measure as they are considered for system use only. And as-from our new no-default-ROOT-Pool-import we also discourage use of the system drive entirely: easing on-boarding and further enforcing system/data separation.

phillxnet:
# Testing
An rpm was built (Leap 15.6 X86_64 host) and installed successfully on the build host. The issue highlighted `home` share was rejected as intended:

attempted-data-pool-home-share-creation

On a system where the ROOT pool had not been imported. The above error is an indicating of the newly added 'fence' to avoid possible conflicts with upstream established system subvolumes.

Similarly an attempt to create a Share named `var` or `opt` was also rejected. The same system was also able to import the `ROOT` Pool; assuming the imported data pool did not already have clashing share names with subvol names on the ROOT pool, this should be the case. Similarly we maintain our Web-UI surfacing of the 'real' /home subvol (ROOT Pool parent) after the system pool was imported (Advanced users only).

phillxnet:
Link, for context, to upstreams definition of the default subvolumes:
https://en.opensuse.org/SDB:BTRFS#Default_Subvolumes

PR #2885, opened 2024-08-12
Rockons: Invalid environment variable (...) #1588

phillxnet:
Based on #1688 by @anatox from 2017! lightly modified for current testing branch.

Rock-on front-end install wizard does not collate container info regarding Rock-on defined `environment` elements. Affects multi-container Rock-ons only. Results in back-end failure to successfully assign user-input environmental variables. Resolve by adding a containerid dimension to the environmental matrix created by the install wizard, and updating the back-end Rock-on instantiator to take advantage of this info. Previously a reverse engineering approach was taken: which cannot work for example, where two containers, within the same Rock-on definition, use the same environmental variable name.

See Issue #1588

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/rockons.js
@@ -807,7 +807,11 @@ RockonEnvironment = RockonCustomChoice.extend({
         }
         var env_map = {};
         var envars = this.custom_config.filter(function(cvar) {
-            env_map[cvar.get('key')] = this.$('#' + cvar.id).val();
+            co_id = cvar.get('container');
+            if(env_map[co_id] == undefined) {
+                env_map[co_id] = {};
+            }
+            env_map[co_id][cvar.get('key')] = this.$('#' + cvar.id).val();
             return cvar;
         }, this);
         this.model.set('env_map', env_map);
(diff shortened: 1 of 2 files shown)


phillxnet:
@FroggyFlox & @Hooverdan96 last commit looks to addresses issue comment: https://github.com/rockstor/rockstor-core/issues/1588#issuecomment-2284257824 with this PR's in-development branch:

> Fix Rock-on pre-install wizard env summary re now listing containers dimension, not the Env's we entered.

#  Before:
Rock-ons-install-wizard-env-summary-issue

# After:
Rock-ons-install-wizard-env-summary-proposal

In time we can modify these tables to account of container relevance but for now we do have larger issues a-foot. Open to simple enhancements where we can instead show the container name: but an approach to retrieving this escaped my JS time available for what is essentially a nice-to-have aesthetic that we can approach later. Mostly interested in establishing the new env per container capability with this PR. Thoughts/suggestions as always welcome.

phillxnet:
We also have, with this branch, successful install of a Rock-on with no environment entries, in this case Gollum: https://github.com/rockstor/rockon-registry/blob/master/gollum.json

And the same for a Rock-ons with only a single environment variable: 

Single-environment-entry-Rock-on
```
[13/Aug/2024 20:03:47] DEBUG [storageadmin.views.rockon_id:109] install request with share_map={}, port_map={'9980': 9980}, dev_map={}, cc_map={}, env_map={'4': {'domain': 'rleap15-6.lan'}}

```

phillxnet:
Moving to squash and re-present.

PR #2909, opened 2024-09-30
Refactor lsblk, resolves issue re leading space in values #2907

### LINKED ISSUE #2907: 5.0.9 to 5.0.14 Disk Scanning fails for leading space in lsblk values
Thanks to user cellisten on the community forum, there seems to be an issue when the model name contains a space within. Usual nomenclature found on model names usually does not contain any spaces.

https://forum.rockstor.com/t/scanning-disks-not-working/9781

There seems to already be a hint on how to address it, but that needs confirmation that it won't have an impact on other scenarios. From the message thread:

> I did some refactoring of the generation of the blk_dev_properties to separate the line in a manner where spaces don’t really matter and the strip is done differently:

```
line = [ item.strip() for item in re.split(r'["=]', line) ]
line = [ item for i, item in enumerate(line,-1) if i %3 != 0 ]
# Device information built from each lsblk line in turn.
blk_dev_properties: dict = {
key.lower() if key != "TRAN" else "transport":
value if value != "" else None for key, value in zip(line[::2], line[1::2])
}
```

cellisten:
scan_disks() raised ValueError when lsblk values had leading spaces refactored to strip leading and trailing spaces

Fixes #2907

### DIFF
--- a/src/rockstor/system/tests/test_osi.py
+++ b/src/rockstor/system/tests/test_osi.py
@@ -1650,11 +1650,6 @@ class OSITests(unittest.TestCase):
         cd /opt/rockstor/src/rockstor/system/tests
         /opt/rockstor/.venv/bin/python -m unittest test_osi.OSITests.test_scan_disks_lsblk_parse_fail
         """
-        # E.g 'VENDOR="        "' failed with: "ValueError: not enough values to unpack (expected 2, got 1)"
-        # We use dictionary comprehension to parse lsblk output and rely on: 'line.strip().split('" ')'
-        # this breaks when a disk property begins with a space.
-        # See: https://github.com/rockstor/rockstor-core/issues/2853
-        # Reported/encountered/addressed failure concerns values consisting solely of whitespace.
         out = [
             [
                 'NAME="/dev/fd0" MODEL="" SERIAL="" SIZE="4K" TRAN="" VENDOR="" HCTL="" TYPE="disk" FSTYPE="" LABEL="" UUID=""',  # noqa E501
@@ -1664,6 +1659,7 @@ class OSITests(unittest.TestCase):
                 'NAME="/dev/sdb2" MODEL="" SERIAL="" SIZE="64M" TRAN="" VENDOR="" HCTL="" TYPE="part" FSTYPE="vfat" LABEL="EFI" UUID="1381-2428"',  # noqa E501
                 'NAME="/dev/sdb3" MODEL="" SERIAL="" SIZE="2G" TRAN="" VENDOR="" HCTL="" TYPE="part" FSTYPE="swap" LABEL="SWAP" UUID="ee6de8d5-eb1c-4d19-943c-0d24ecee92a7"',  # noqa E501
                 'NAME="/dev/sdb4" MODEL="" SERIAL="" SIZE="29.9G" TRAN="" VENDOR="" HCTL="" TYPE="part" FSTYPE="btrfs" LABEL="ROOT" UUID="46befa9e-b399-480d-b10d-338a4c875615"',  # noqa E501
+                'NAME="/dev/sdd" MODEL="SanDisk 3.2Gen1" SERIAL="01016c73a2fabf22051f78f30546627c9fb5e8083fbc14d477f85d6a269220dd15710000000000000000000060a25f5700804f0091558107a3ab6843" SIZE="57.3G" TRAN="usb" VENDOR=" USB    " HCTL="6:0:0:0" TYPE="disk" FSTYPE="" LABEL="" UUID=""',  # noqa E501
                 "",
             ]
         ]
@@ -1703,6 +1699,22 @@ class OSITests(unittest.TestCase):
                     root=False,
                     partitions={"/dev/sdb4": "btrfs"},
                 ),
+                Disk(
+                    name="/dev/sdd",
+                    model="SanDisk 3.2Gen1",
+                    serial="01016c73a2fabf22051f78f30546627c9fb5e8083fbc14d477f85d6a269220dd15710000000000000000000060a25f5700804f0091558107a3ab6843",
+                    size=60083404,
+                    transport="usb",
+                    vendor="USB",
+                    hctl="6:0:0:0",
+                    type="disk",
+                    fstype=None,
+                    label=None,
+                    uuid=None,
+                    parted=False,
+                    root=False,
+                    partitions={}
+                )
             ]
         ]
         # As all serials are available via the lsblk we can avoid mocking
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ rockstor-tasks-huey.db*
 # Poetry error logs
 poetry-installer-error*
 /build-output.txt
+
+# Pyenv version
+.python-version
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -326,19 +326,20 @@ def scan_disks(min_size: int, test_mode: bool = False) -> list[Disk]:
         if line == "" or re.match("NAME", line) is None:
             continue
         # lsblk example line (incomplete): 'NAME="/dev/sdb" MODEL="TOSHIBA MK1652GS" VENDOR="ATA     " LABEL="" UUID=""'
-        # line.strip().split('" ') = ['NAME="/dev/sdb', 'MODEL="TOSHIBA MK1652GS', 'VENDOR="ATA     ', 'LABEL="', 'UUID=""']   # noqa E501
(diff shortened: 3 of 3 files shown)


cellisten:
Realised I hadn't added the updated test case in the right place. Now I have (hence the latest force push)

phillxnet:
@cellisten Hello again, and thanks for this submission. Great to have more developer input.

It does look like what we have previously handled, and submissions to this area of the code must come with accompanying tests to reproduce an original issue, and thus prove the proposed fix. Take a look at the PR & linked issue I put against this PR's issue (I'll add in this PR's text shortly). I did leave links in @Hooverdan96 issue for this PR. There you should see how to run all the unit tests, it should be relatively simple to add your reproducer lsblk output as you presented on the forum to an existing test or to a new dedicated one. This code area is so low, it's always a worry to make changes. 

I'm looking forward to having a closer look at what you have presented here however. And there is for-sure more robustness to be had. But I'd also like to try and keep stuff readable. Take a look at the issue linked PR which last fixed something similar and presented a new test. That should help to see who one runs/extends tests in this area.

Thanks again for this submission: as always much appreciated. And my apologies for not being able to take a proper look right-away.

phillxnet:
@cellisten Opps we've had a comment overlap. Re:

> Realised I hadn't added the updated test case in the right place. Now I have (hence the latest force push)

That modifies an existing test. That would potentially loose the regression that tested against. Have another look at the test structure, it's possible to add an entire new set (potentially constrained to smallest possible) to an existing test. That way we preserve the prior tests proven work. And add another set of data that may show a slightly different behaviour that it looks like you have approached (leading space). My prior comment hints as that being outstanding: so this is a nice addition.

See if you can add a new test data set, rather than modify an existing one. To preserve the original test. Which was for something a little different. Each test ideally tests for a specific failure.

Again apologies I can't focus just yet on this PR, but again this effort is much appreciated.

phillxnet:
@cellisten Those comments are a little unclear addimittely. It failed before my changes with the indicated data. And passed there-after. To prove those particular changes. What we need is another set of data on that same test to initially fail with 5.0.14, and then pass post your changes. All other tests must also pass.

> However, the test case had no values that began with a space. If it had, the error would've been caught earlier

And my comments acknowledge this very short-comming. We are defintely getting there on this one :) . I addressed a different failure detailed in the associated issue and that test proved what I was fixing to be fixed. We need another test to prove a similar fix presented here. I like what you've done by the way, from only another quick look.

Changing existing tests means we risk loosing what they were designed to fix. We recently had to completely re-vamp who we do this part of the code. So many things were changed. And we have a number of tests to cover this whole area. We need to ensure we have some continuity from how we worked before. The test, as may of our are, are intended to be extended with specific minimum reproducers. But often it's just real life data, which you have. Plug in your reproducer data and see it fail with another loop over the new set of data. And then we gain two instances of specific failure. Regression testing. Sometimes a test does not quite test what it was intended to test agreed. But that test failed then passed with the changes made to address (not) leading space - but trailing. We still failed on leading space: as indicated in my comment quotes in the associated issue to help focus the issue itself.

phillxnet:
https://github.com/rockstor/rockstor-core/blob/658291b11b28e9c9e23c4538983f1e6d059d4ca7/src/rockstor/system/tests/test_osi.py#L1653-L1657

cellisten:
No problem at all, I agree that existing regressions need to be tested. Was a bit naive in thinking that testing both in the same would be ok. I have now added an extra line in the existing test case, hope that is more suitable.

src/rockstor/system/tests/test_osi.py
```diff
@@ -1665,6 +1665,7 @@ def test_scan_disks_lsblk_parse_fail(self):
                 'NAME="/dev/sdb3" MODEL="" SERIAL="" SIZE="2G" TRAN="" VENDOR="" HCTL="" TYPE="part" FSTYPE="swap" LABEL="SWAP" UUID="ee6de8d5-eb1c-4d19-943c-0d24ecee92a7"',  # noqa E501
                 'NAME="/dev/sdb4" MODEL="" SERIAL="" SIZE="29.9G" TRAN="" VENDOR="" HCTL="" TYPE="part" FSTYPE="btrfs" LABEL="ROOT" UUID="46befa9e-b399-480d-b10d-338a4c875615"',  # noqa E501
                 "",
+                'NAME="/dev/sdd" MODEL="SanDisk 3.2Gen1" SERIAL="01016c73a2fabf22051f78f30546627c9fb5e8083fbc14d477f85d6a269220dd15710000000000000000000060a25f5700804f0091558107a3ab6843" SIZE="57.3G" TRAN="usb" VENDOR=" USB    " HCTL="6:0:0:0" TYPE="disk" FSTYPE="" LABEL="" UUID=""',  # noqa E501
```
phillxnet, line 1668: @cellisten Just a note about this additional line. All good except that we purposefully try to mimic the output of our command rapper: which notably has an empty line as it's last line of output. Hence the "" entry there. So your additional line should belong before, not after that "" entry. Also note that the order of output can be important. Hence us predominantly using real-world output where applicable. But I don't think that is important in this case as there is no partition info for this extra drive: these normally come after the parent block device (whole device).

Thanks for continuing to chip away at this one. Again much appreciated.

Have you run all of the tests with your changes in-play? That may be informative.

Also, what long serial numbers some of those SanDisk drive have. And of course this new quirk re spaces all over the place around  in VENDOR value (up-front even now!). Sometimes I think they just make this stuff up :).

cellisten: Honestly, lsblk might do well to sanitize some of the more non-machine readable output :)
I have run the full barrage of tests without any issues with the change. I will move the entry. I was wondering about that final empty one and first thought it was a test case in of itself but now I understand it.

phillxnet: Agreed, some of these commands have a json output option. But we are also still releasing rpms for older Leap versions. So that can mean that the newer commands that have these more modern option modes are not yet available.

But all in good time hopefully. Plus there are ongoing improvements in some Python libraries native to btrfs that should help lift us out of some of this more messy human readable command output fragility. But in a way I also like that we run 'real world' commands that folks can also run themselves to test output outside of our parsing. Swings and roundabouts.

cellisten:
Thank you very much for the fast feedback loop and nice welcome! As a new contributor it is heart warming to get this much contact so fast. Looking forward to any comments when do you do get time for a proper review!

phillxnet (approved):
@cellisten Thank for seeing to this outstanding failure, much appreciated.
My apologies for taking so long to get back to this.

I've now, finally, gotten around to doing a test rpm build and this in turn runs all our tests, and all looks good on that front. From the forum discussion re running tests etc we have the following doc entry: https://rockstor.com/docs/contribute/contribute.html#code-test
and:
```
```
on an rpm based install, looks happy, ending in :
```
Ran 285 tests in 36.991s

OK
```

Thank for the quick responses re earlier feedback by the way, and for the test work. My intention with expanding that test was to add an entire other system data set, with its own expected output, to the existing test loop, expressing the leading space reproducer: but the added drive entry approach is similar. We should still via failure log be able to see which parsing failed. I just like whole system data sets as they can end up being expressive of something that was not noticed before - and the more whole-system datasets the better. Hence us having loads of test data seemingly redundant: but never-the-less the end target is whole systems after all. This is also the reason for ending the mock output of lsblk as it actually ends, with the empty line. The existing data we had on this front was super useful for the last refactoring. And hopefully we can in-turn refactor later once all our OS bases can give us json to hand-off this tedious parsing to the underlying command itself. And then bring in via json to dict.

On a further note re tests/unit-tests nomenclature, as per a Brian Okken podcast comment/advice, I think we should now just call these 'tests': i.e. what even is a 'unit' these days. But we do still try to keep them small. We hope in-time to have what I regard as more meaningful end-to-end tests, using openQA: but again all down to developer resources. We once, years ago now, has a selenium test set. But that has long since faded into an unmaintained state and I don't want to depend on Java for any part of our infrastructure.

I'm chuffed to have this fix in, as it bothered me that I'd not found a solution when last there - and this does look to be a relatively elegant approach. And thanks for the in-code comments: much appreciated. This code recently received a major re-vamp so some new bugs were inevitable. But as-way it had become almost unworkable obscure/convoluted. I look forward to our underlying tools being able to present us with JSON, but all in good time I guess.

Thanks again for all your efforts here, and do keep an eye out on the forum for more bugs in this area as there are few folks who dabble this low in the disk management, and the more the better really.

This patch will end-up being released in our pending 5.0.15-0 rpm.

@Hooverdan96 Thanks for helping with the creation of the original issue here.

PR #2910, opened 2024-09-30
Use pyproject.toml's 'version' in build.sh #2895

### LINKED ISSUE #2895: Use pyproject.toml's 'version' in build.sh
It is proposed that we normalise/centralise on the `version` defined in pyproject.toml within build.sh when setting the `jslibs_version` bash variable. This should ease further the procedures required to release rpms: where we synchronise GitHub releases to these version numbers.

phillxnet:
Use pyproject.toml's (version = "5.0.14") via `poetry version` output: e.g. "rockstor 5.0.14" to populate jslibs_version variable within build.sh. Intended to ease project versioning/release strategies by having a single point of reference for release versions. Typically rockstor-jslibs has a GitHub released in parallel with rockstor-core.

Includes a minor build.sh logging fix re .env file source directory.

Fixes #2895

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -51,9 +51,13 @@ poetry self show plugins >> poetry-install.txt
 poetry install -vvv --no-interaction --no-ansi >> poetry-install.txt 2>&1
 echo
 
+# Source package version from pyproject.toml's (version = "5.0.14") via `poetry version` output:
+# e.g. "rockstor 5.0.14"
+ROCKSTOR_VERSION=$(poetry version | sed 's/rockstor //')
+
 # Add js libs. See: https://github.com/rockstor/rockstor-jslibs
 # Set jslibs_version of GitHub release:
-jslibs_version=5.0.14
+jslibs_version=$ROCKSTOR_VERSION
 jslibs_url=https://github.com/rockstor/rockstor-jslibs/archive/refs/tags/"${jslibs_version}".tar.gz
 
 #  Check for rpm embedded, or previously downloaded jslibs.
@@ -84,7 +88,7 @@ fi
 /usr/bin/gpg --quick-generate-key --batch --passphrase '' rockstor@localhost || true
 # Init 'pass' in .env defined PASSWORD_STORE_DIR using above GPG key, and generate Django SECRET_KEY
 set -o allexport
-echo "Sourcing ${pwd}.env"
+echo "Sourcing $(pwd).env"
 source .env  # also read by rockstor-build.service
 set +o allexport
 /usr/bin/pass init rockstor@localhost


phillxnet:
# Testing

The `build.sh` script is run by the `rockstor-build` systemd service.
To test the modifications here the following was enacted:
(An existing 5.0.14-* rpm install was in place; but with the proposed build.sh modifications)
```
systemctl stop rockstor*
rm /opt/rockstor/rockstor-jslibs.tar.gz
rm /opt/rockstor/rockstor-jslibs.tar.gz.sha256sum 
rm -rf /opt/rockstor/.venv/
```
And a full service restart was initiated via:
```
Sep 30 12:16:24 rleap15-6 build.sh[14455]: Getting rockstor-jslibs version 5.0.14
Sep 30 12:16:24 rleap15-6 build.sh[14521]: --2024-09-30 12:16:24--  https://github.com/rockstor/rockstor-jslibs/archive/refs/tags/5.0.14.tar.gz
...
Sep 30 12:16:26 rleap15-6 build.sh[14455]: Sourcing /opt/rockstor.env
Sep 30 12:16:26 rleap15-6 build.sh[14530]: Password store initialized for rockstor@localhost
```
N.B. we have an outstanding typo re path'/'.env file to be addressed when we are next modifying this file.

cellisten:
I hit this yesterday when testing but couldn't figure it out :)

phillxnet:
@cellisten Yes, we have had these two 'version' references and it's a pain when doing releases. So I was just seeing to a little simplification hopefully. Currently there is little movement in jslibs but our next testing phase intends to focus more on updates/reductions there. So I wanted to get this in place ready. I've just got to make sure this doesnt' break our test build back-end and then I'll get to merging.

cellisten:
As a workaround I managed to build by running it directly with ./build.sh rather than the recommended sh build.sh from the docs

phillxnet:
An rpm was build with this branch and looks to have a successful:

> Start rockstor-bootstrap service

step.
See: https://github.com/rockstor/rockstor-rpmbuild/tree/testing for rpm spec file and procedure.
We build our distribution rpm, via poetry and two source elements, so that we can distribute the appropriate jslibs with the versioned rpm. But we maintain the flexibility for build.sh to enact it's own GitHub release download of rockstor-jslibs.

phillxnet:
@cellisten re:

> As a workaround I managed to build by running it directly with ./build.sh rather than the recommended sh build.sh from the docs

Yes, there are always improvements to be had in the docs. Do feel free to submit any corrections/enhancements that take your fancy. New folks to our build process often have valuable insite as to what is not clear.

Note however that build.sh, although handy to run by-hand, must always be ameanable to being run by it's systemd service which we mention in the docs:

>  rockstor-bootstrap service

As that sets up the expected environment etc. And note the sensitivity re .venv: our locally installed/build python virtual environment. We have changed a few details and added this very service in this on-going testing phase. We used to do the build via rpm's %posttrans scriptlet but that ended up being way too much overload for what is a very restrictive environment and far too much to do during an install (or stricktly the latter section post transitional install). Moving the .venv build to a systemd service has helped massively with making changes there more easily testable and eases diagnosis re failed builds of the .venv re internet access etc. As they show up as a failed service: far more approachable than a posttrans scriptlet fail.

cellisten:
Is the rockstor-bootstrap service still available the first time after the original install is removed and the rsync has been done? I might redo it all to verify the steps and then I will try to see if I can think of any improvements

Issue #2913, opened 2024-10-04
OpenVPN rock-on install wizard fails to show pre-install summary 


phillxnet:
While developing against "Rockon pre & post-install summaries are inconsistent" #2904 an OpenVPN install wizard failed to show the pre-install config summary page: a second Confimation on the existing dialog continued the installation which succeeded successfully.

With debug enabled the following log resulted when installing this Rock-on: the only Rock-on remaining that uses the "custom_config" key: passed from front-end to backend in the following as cc_map
```
[04/Oct/2024 18:16:54] DEBUG [storageadmin.views.rockon_id:108] install request with share_map={}, port_map={'1194': 1194}, dev_map={}, cc_map={'servername': 'fqdn.fake.biz'
}, env_map={}
```

Browser console error:
> Uncaught TypeError: can't convert undefined to object

with a reference to our storageadmin.js (all-in-one js lib) re:
```
nextPage: function() {
        var _this = this;
        var promise = !_.isNull(this.currentPage) ?
            this.currentPage.save() :
            $.Deferred().resolve();
        promise.done(function(result, status, jqXHR) {
            _this.incrementPage();
        });
        promise.fail(function(jqXHR, status, error) {
            console.log(error);
        });
    },
```

---

Post install the entered details appear as expected (post proposed changes associated with #2904) in the post install config summary:

Resource type | Internal Reference | External Reference / Configured value
-- | -- | --
Port | 1194 | 1194
Custom | servername | fqdn.fake.biz

Hooverdan96:
May be it's also time to retire this docker image and switch to the offical one that's available (don't know when it was first posted) from openvpn, and possibly retire the final custom_config usage?

https://hub.docker.com/r/openvpn/openvpn-as

Has to be tested, but it seems that it's pretty self-contained and also contains an ui component.

phillxnet:
@Hooverdan96 Agreed on all accounts. I'd not looked into the availability of  an official image, and so if there is one available then it definitely our current offering - with the custom config and lots of command-line requirements. The Official one can't be much worse :) .

> ... and also contains an ui component.

OK, so that's a major win. Nice find. Yes, lets drop this Rock-on entirely and go in that direction.

Hooverdan96:
I opened [#392](https://github.com/rockstor/rockon-registry/issues/392) on the rockon repo for this.

Hooverdan96:
[...]

PR #2915, opened 2024-10-07
Rockon pre & post-install summaries are inconsistent #2904

### LINKED ISSUE #2904: Rockon pre & post-install summaries are inconsistent
As a continuation of the original issue #1588 (and associated PR #2887) there is possibly another instance where the shares are swapped Name of share within the docker container vs. the Rockstor share mapped to it. For example, when installing the WGEasy Rockon (also see recent PR [#389](https://github.com/rockstor/rockon-registry/pull/389)) the share is swapped in the summary screen before the installation is triggered:

image

`/etc/wireguard` should be the name, and the blurred out Rockstor share would be the `Mapped Representation`

phillxnet:
Clarify header and mouse-over text in both associated tables. Normalise on column order (as per header) as per existing pre-install summary table.

Includes:
- Bug fix re some internal/external header-to-contents columns swapped.
- Incidental bugfix re empty internal (to Rock-on) device cell entries.

Fixes #2904

---

## Caveat

We have, post this proposal, one significant anomaly remaining.

In the first of the following summary dialog proofs (Pre install) we list for example `container-id:22` against the share name in the `External Reference / Configured value` column. It would be more appropriate to indicate this container reference against the corresponding `Internal Reference` column. And to of course resolve the id to the actual container name. However; this change is far deeper than is appropriate at this stage of our late testing phase development. And is to be addressed in due-course as we progress through itterative testing/stable phases. The main focus of this issue/PR set was to, with trivial template changes only, address confusing column switches and miss-representations only. In the same manner it would be nice to also surface container (by name) references in the final installed summaries. This, similarly, will be considered once we have addressed more pressing matters such as an update to our backbone dependencies.

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/settings_summary_table.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/settings_summary_table.jst
@@ -2,24 +2,24 @@
 <table class="table table-condensed table-bordered table-hover table-striped tablesorter">
 <thead>
   <tr>
-     <th>Resource type&nbsp;&nbsp<i class="fa fa-info-circle fa-lg" title="Type of the system resource on Rockstor. Eg: Share, port numbers etc.. Custom type is for arbitrary variables needed by the Rock-on that may not map to a system resource." rel="tooltip"></i></th>
-    <th>Name&nbsp;&nbsp<i class="fa fa-info-circle fa-lg" title="Name of the resource according to Rockstor. Eg: Share names, port numbers etc.." rel="tooltip"></i></th>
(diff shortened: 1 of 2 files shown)


phillxnet:
# Testing

An rpm was build and the handbrake rockon installed as it has share, env, dev, and facilitates the addition of docker entworks (rocknets) and labels (docker labels) to provide example screenshots of pre & post config submission summary screens thus:

## Pre install

hb-pre-install-config-summary


## Post install

hb-post-install-config-summary

---

## Pre rocknet
Rocknets are a docker networks wrapper within the Web-UI.

hb-pre-rocknet-add-summary

## Post rocknet

hb-post-rocknet-add-summary

---

## Pre container label

https://rockstor.com/docs/interface/overview.html#add-labels
Rockstor's Web-UI wrapper around docker labels.

hb-pre-label-add-summary

## Post container label

hb-post-label-add-summary

phillxnet:
We have only one Rock-on that uses the custom config option, which is the openVPN rockon. There is an existing issue:

"OpenVPN rock-on install wizard fails to show pre-install summary" #2913

Which prevents the creation of a pre install summay of this option variant. But the following is a post install summary demonstrating we have intended representation of the custom config option used:

openVPN-post-install-custom-option-summary

Issue #2937, opened 2024-12-17
Milestones for default & optional environment & volume-mapping fields


kanecko:
I thought of making a parent issue so we can track the progress being made.

- [ ] Add support for default values in required environment fields. #2012
- [ ] Add support for default values in required **volume-mapping** fields. 
- [ ] Add support for **optional** values in **environment** fields.  #2094
- [ ] Add support for optional values in **volume-mapping** fields. #2280
- [ ] Update the following Rockons with default and optional fields:
   - [ ] pi-hole
   - [ ] jellyfin
   - [ ] netalertx
   - [ ] omada-controller
   - [ ] nginx-proxy-manager
   - [ ] radarr
   - [ ] sonarr
   - [ ] transmission
   - [ ] **(please suggest others as well!)**
- [ ] Add new Rockon "Unpackerr"

phillxnet:
@FroggyFlox & @Hooverdan96 Should we make this into a Milestone proper? I.e. we have now it seems reached some-what of a critical mass on our Rock-ons v1 lets say: arbitrarily. Maybe we have sufficient interest, re @kanecko renewed presence, to establish a Rock-ons V2 milestone. To be addressed as interest/resources allow, in parallel with our front end updates. Partial conflict but that is unavoidable with the large changes we have already made in the past: and are still planning for re UI dependencies re-vamp.

Super important in all these Rock-ons changes is to not break at least our last Stable, currently 4.6.1-0, and ideally last but one 4.1.0-0. But with our next Stable just around the corner we might only really have to concern ourselves re backward compatibility to 4.6.1-0 in the very near future; as we will then have a 5.1.0-0. And thus far changes such as convenience defaults honours this (at least I think/hope it does).

We need not list all affected Rock-ons as that is redundant and dynamic: we need only collect issues that detail planned Rock-on subsystem improvements. Especially as we have to honour all existing installs (last one or two stables). Alternatively we just use issue labels: but they are less obvious progress wise. And if we have a renewed interest in upping the Rock-on sub-system it would be good to establish this up-front via an Issue Milestone.

@kanecko Great to have renewed interest in our Rock-ons subsystem. However we are unlikely to be able to cram any of these deeper (model change) improvements into the next Stable (current testing before Stable spin-off) but that does give us way more leeway to make bolder changes in the testing channel there-after (soon). So all good: especially as we are all after a much shorter testing phase the next go around.

kanecko:
[...]

Issue #2012, opened 2019-02-04
add env default option for rockons


magicalyak:
This is related to https://github.com/rockstor/rockon-registry/pull/174 where I realized we don't have a default option in the json that is respected for environment options in rockons. 
1) adding default has not effect
2) I would like to be able to add default like the volume or port options for environment items

FroggyFlox:
Adding a new feedback from the forum supporting the integration of this feature: 
[https://forum.rockstor.com/t/add-a-default-value-to-the-environment-object-of-rockon-conf-file/8633](https://forum.rockstor.com/t/add-a-default-value-to-the-environment-object-of-rockon-conf-file/8633?u=flox)

That post should be updated upon resolution of this issue.

FroggyFlox:
Adding another forum thread requesting this feature:
[https://forum.rockstor.com/t/optional-environment-variables-in-rock-ons/8756](https://forum.rockstor.com/t/optional-environment-variables-in-rock-ons/8756?u=flox)

PR #2925, opened 2024-11-11
NFS tests in need of maintenance #2923

phillxnet:
NFS test coverage/quality improvements.

Fixes #2923

## Includes
- Fixture details & creation instructions via code comments.
- Addition of test_nfs.json fixture from above instructions.
- Code comment docs on how to invoke this test set.
- Completes previous partial re-enablement of fixtures.
- Addresses all outstanding TODOs.
- Improves input validation coverage for both POST & PUT.

PR #2938, opened 2024-12-17
Adds support for default values in required environment fields

kanecko:
Implements #2012
Parent #2937

### DIFF
--- a/src/rockstor/storageadmin/models/rockon.py
+++ b/src/rockstor/storageadmin/models/rockon.py
@@ -227,7 +227,7 @@ class DContainerEnv(models.Model):
     val = models.CharField(max_length=1024, null=True)
     description = models.CharField(max_length=2048, null=True)
     label = models.CharField(max_length=64, null=True)
-
+    default_val = models.CharField(max_length=1024, null=True)
 
 class DContainerDevice(models.Model):
     container = models.ForeignKey(DContainer, on_delete=models.CASCADE)
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/cc_form.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/cc_form.jst
@@ -16,7 +16,21 @@
             {{else}}
                  type="text"
             {{/if}}
-            id="{{this.id}}" name="{{this.id}}" value="{{#unless null}}{{this.val}}{{/unless}}">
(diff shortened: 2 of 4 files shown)


FroggyFlox:
Hi @kanecko ,

Thanks a lot for getting this going, and thank you especially for mentioning tests. I was actually going to add that to your milestone list as I believe we need to first establish some test coverage before we can add all these new features.

I'll try to get back with more details soon (unless someone near me to it, of course) as I'm a bit short on time at the moment but here are a few pointers real quick in case you haven't seen it already:
First, this would be for last when no more adjustments are needed anymore (if any):
https://rockstor.com/docs/contribute/contribute.html#database-migrations

Examples of testing and how to run them: I would have a look at the most recent work of that type by @phillxnet: https://github.com/rockstor/rockstor-core/pull/2925#issue-2649411385

I'll get back with more shortly, hopefully, and thanks again. That's all exciting!

phillxnet:
@kanecko Nice start. Could you move this to draft status, and to a named branch against the primary issue. This helps greatly with avoiding common mistakes when working on several things simultaneously. Which it looks like you may be doing shortly.

I.e. this PR is from your testing branch: better to have an issue specific branch. This will likely mean a re-submission, but not worries as you can reference this PR in the next one.

See our contributor guidelines on this here: https://rockstor.com/docs/contribute/contribute.html#developers

Named branches also help all-around to keep track of what a branch is supposed to be concerned with. 

Great to have more eyes on the Rock-on code by the way.

kanecko:
Thanks for the info.

@FroggyFlox I didn't explicitly mention or add tests there because I want to include them before moving on to the next milestone anyway.

@phillxnet I will resubmit the PR as instructed, thanks for the heads up.

> This helps greatly with avoiding common mistakes when working on several things simultaneously. Which it looks like you may be doing shortly.

Actually, each milestone is a standalone, releasable improvement. I imagined my workflow would be to first complete (review+commit to rockstor/rockstore-core:testing) a task before moving on. 
I'm somewhat new to git and github, so I don't know what the usual workflows are here. I don't mind creating new branches for each task, but I would prefer to complete a current task before moving on to the next one. I also prefer to do many small commits rather than one big one, because I can get early feedback.
Another reason why I prefer to commit directly to your testing branch is that I don't know if I'll be able to finish all the milestones by the end of my vacation. If I have to resume the work in a year, it would be great if I didn't have to merge changes that have happened in the meantime.

PR #2962, opened 2025-02-26
Normalise menu capitalisation #2960

### LINKED ISSUE #2960: Normalise menu capitalisation
Our existing "SUPPORT" top level menu in the Web-UI has the following entry re support email: "Team email":

Image

It is proposed, for consistency, that this be change to read "Support Email", as per the email references within our Website "Contact Us" subsection:

Image

We might also like to tend to this menu's other entries regarding their capitalisation; i.e. see the "SYSTEM" menu items use of capitalisation in both words of double word menu items.

Hooverdan96:
Fixes #2960 

- Adjust capitalization in Top Navigation Bar
- make indentation consistent (spaces instead of tab on some tags).

With new built (and collected statics):

image 
image

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/common/navbar.jst
@@ -13,7 +13,7 @@
         <li><a href="#replication"><i class="fa fa-long-arrow-right"></i> Send</a></li>
         <li><a href="#replication-receive"><i class="fa fa-long-arrow-left"></i> Receive</a></li>
       </ul>
-      <li><span><i class="glyphicon glyphicon-share"></i> File sharing</span></li>
+      <li><span><i class="glyphicon glyphicon-share"></i> File Sharing</span></li>
       <ul style="list-style-type: none">
         <li><a href="#nfs-exports">NFS</a></li>
         <li><a href="#samba-exports">Samba</a></li>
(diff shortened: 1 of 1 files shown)


phillxnet (approved):
@Hooverdan96 Thanks for seeing to this: looks much better now. And for extending and renaming the issue appropriate to similar instances. Plus the spaces instead of tabs fix, we tend/aim towards sets of 2 spaces in HTML and 4 in our Python.

As per your tests: I concur via an rpm build/install:

support-menu
FileSharing-menu-entry

It is I think important to maintain as much consistency with such things as we are able given it can be almost sub-consciously jarring otherwise. Plus with some established norms we can get and keep this right as we work towards re-doing our front end technologies.

phillxnet:
@Hooverdan96 I've removed the brackets in the title as I'm unsure of our changelog parser on this front. But it looks better with them. I have considered this format previously but didn't want to throw 2 characters at it :). Lets try this format when we being our next testing phase - where we can more easily handle a failure on that front.

Hooverdan96:
> we tend/aim towards sets of 2 spaces in HTML and 4 in our Python.

Yes, I found that the tabs were there instead of 4 spaces on some of the indented 2nd and for some 3rd levels had a combo of one tab and two spaces. I assume, that was because of subsequent PRs and possibly editors that by default would use tab instead of spaces for indentation (notepad++ does that for certain languages). While probably widely accepted to use 2 and 4 spaces respectively, it probably makes sense to add that to the contribution section explicitly as a minimal style guide (and maybe also the capitalization of menu entries, etc.). All without going overboard on formatting requirements of course 😄.

PR #2964, opened 2025-02-28
Re-establish Django admin interface #2891

### LINKED ISSUE #2891: Re-establish Django admin interface
By way of preparation for our next Stable rpm release, it is proposed that we do a little house-keeping regarding some old legacy files, such as those dating back to pre rockstor-jslibs times; and pre GitHub publishing times also! An example in this case would be the long defunct Django admin override template files found in:

- src/rockstor/templates/admin

- [x] Remove to avoid their potential interference with the now way-newer .venv sourced files from our current Django:

- .venv/lib/python3.11/site-packages/django/contrib/admin/static

- [x] Re-instate a fresh Django Admin site config/capability.
- [x] Add Django admin site Docs feature.

phillxnet:
Intended primarily to improve the developer experience.

Fixes #2891 

- Remove prior ancient/legacy admin template overrides.
- Add previously missing URL entry for admin site.
- Enable admin docs site, along with matching URL.
- Remove redundant admin.autodiscover() use.
- Update Django admin config to reflect more the current project state.
- Add basic admin site config for:
- All storageadmin app models: Share, Snapshot, Disk, Pool, SambaShare, SambaCustomConfig, SFTP, NFSExportGroup, NFSExport, IscsiTarget, PosixACLs, DVolume, DContainer, DCustomConfig, ContainerOption, DContainerLink, RockOn, DPort, DContainerArgs, DContainerLabel, DContainerEnv, DContainerDevice, DContainerNetwork, User, Group, DImage, PoolScrub, PoolBalance, BridgeConnection, BondConnection, TeamConnection, EthernetConnection, NetworkDevice, NetworkConnection, APIKeys, Appliance, UpdateSubscription, AdvancedNFSExport, SupportCase, DashboardConfig, Setup, Plugin, InstalledPlugin, OauthApp, TLSCertificate, SMARTInfo, SMARTCapability, SMARTAttribute, SMARTErrorLog, SMARTTestLog, SMARTTestLogDetail, SMARTIdentity, ConfigBackup, EmailClient, Pincard.
- All smart_manager app models: CPUMetric, DiskStat, LoadAvg, MemInfo, NetStat, NFSDCallDistribution, NFSDClientDistribution, NFSDShareDistribution, NFSDUidGidDistribution, VmStat, SProbe, PoolUsage, Service, ServiceStatus, ShareUsage, Replica, ReplicaTrail, ReplicaShare, ReceiveTrail, TaskDefinition, Task.


FroggyFlox (approved):
That looks amazing, @phillxnet, thanks a lot for taking the time to not only re-enable the Admin, but include all models and convenience info in many list views!

Just tested in Leap 15.6 and all worked well!

phillxnet:
@FroggyFlox Thanks, it was a bit of leg-work but at least we have an easier on-boading now for folks already familiar with Django's gem of the admin site. I'd like to link this in to your DDT work pending: but that should wait until the next testing phase given it's a little deeper than what was done here. I.e. it would be good to have the admin site similarly enabled, even if it's just the urls, akin to what you are lining up re DDT. 

We have some model comment formatting changes to be done so the admindoc site can show them in every model entry. But I didn't want to push so many (any ideally)  model file changes: even if they were comment change only. So we can deal with that in another pr when we are safely into the next testing phase also.

Thanks for the review, as always much appreciated. I'll  get this merged shortly ready for the next rpm release also.

I'll try to get a doc issue/pr together soon to introduce the `createuser` requirement and the slightly non-standard (on purpose) url. This should make for a nice addition to our developers docs hopefully.

PR #2966, opened 2025-03-04
Remove defunct dead-code associated SystemTap artifacts #2965

### LINKED ISSUE #2965: Remove defunct dead-code associated SystemTap artifacts
We have some compiled binaries associated with our long-dead (unused) SystemTap NFSD moditoring subsystem. It is proposed that we remove these by way of avoiding possible confustion. Only the binary artifacts need be removed as the existing code can remain to guide this issues parent issue's efforts in transiting this same capability to a bpftrace / eBPF mechanism.

---

Parent issue: Revise/review SystemTap NFSD monitoring #2963

phillxnet:
Remove the subject binary artifacts to avoiding possible confusion. All existing SystemTap SProbe code (dead) remains to server as a guide; assuming this entire subsystem is not later superseded/deleted.

Fixes #2965


FroggyFlox (approved):
Thanks a lot for digging into that and clearly laying it out.
I completely agree that this should be cleaned up to help with our overall code base simplification and removing unnecessary binary blobs.
Hopefully we'll be able to implement a nice and modern alternative at some point in the future.

phillxnet:
@FroggyFlox Thanks for the review, as always much appreciated.

I've since done a basic NFS test: likely redundant but no harm done apparently :).

Merging to move on with our current milestone.

PR #2987, opened 2025-05-26
Update public GPG key #2986

### LINKED ISSUE #2986: Update public GPG key
Our ongoing (just) public PGP key, used to sign all RPM & SRPM packages, as well as all repositories we distribute them via, has undergone an expiration extension.

Update the local to each install copy held in the `conf` sub-directory as the file named `ROCKSTOR-GPG-KEY`

Install full path for this file is:

- /opt/rockstor/conf/ROCKSTOR-GPG-KEY

---

This file is read and used as an rpm --import on every visit to SYSTEM -> Softwoare Updates. Likely and area we can save the odd 13 ms.

A future enhancement would be to provide a built-in way to ensure this files contents is kept in sync, in a non resource intensive way, with the now canonical URL source: https://rockstor.com/ROCKSTOR-GPG-KEY but it would be overkill to resource this canonical location on every page visit and would also not be a welcome mechanism for the nature of appliance we represent.

phillxnet:
Used in `rpm --import` as all 'rockstor' RPM & SRPM packages and the repos they are distributed via are signed by this key's private counterpart.

Fixes #2986 

Previous key's expiration date: May 29/05/2025

Details:

Email: support@rockstor.com
Name: The Rockstor Project (Rockstor Development)
Valid from: 30/05/2020
Valid Until: 24/05/2030
Key-ID: 1C42 62F2 5F04 3187


phillxnet:
As a simple update to the in-code public GPG key this is to be merged directly.
There is an outstanding issue where our existing `rpm --import ...` of this expiration extended key silently fails while the original key is still known to rpm (at least while it is still valid), currently requiring the originals removal via:
```
rpm --erase gpg-pubkey-5f043187-5ed2a099
```
But this prior pre-expiration-extended key's removal is to be addressed in a follow-up dedicated issue/pull-request set.

PR #2996, opened 2025-06-02
2995 nfs exports of snapshots fail to be deleted

delboy711:
My first Rockstor commit.  Hope I have done it correctly.
This works for me on my test VM.
When an  NFS share export is deleted it will iterate through all the current exports looking for exports of snapshots for that share, and delete them before deleting the share export itself.
Whether it helps with any of the other issues with NFS I cannot say


FroggyFlox:
Hi @delboy711 ,

Thanks a lot for your contribution! Yes, please close this PR and open a new one against the Testing branch.
Just in case you haven't seen this yet, you will find some relevant documentation at https://rockstor.com/docs/contribute/contribute.html

If you can, please also include brief instructions on how we can best test your PR; in your case I believe it could be as simple as "create NFS export, do x and y on them, try to delete them, etc...". That greatly helps in speeding up the review process so that we know exactly what to test on our end.

Thank you again!

PR #2998, opened 2025-06-04
Delete snapshot NFS exports prior to deleting share exports #2995

### LINKED ISSUE #2995: Delete snapshot NFS exports prior to deleting share exports
Using Rockstore 5.0.15
If a share is created and then exported via NFS, and then a snapshot is created, then the snapshot is automatically exported via NFS with its own entry in /etc/exports and mounted on /export/share_name_.snapshot_name

If the NFS export of the original share is then deleted, the NFS export of the snapshot remains as well as the mount point.

It is then not possible to recreate the export of the share, nor is it possible to delete the snapshot.  Manual intervention is required to delete the snapshot export and its (immutable) mount point.

I have seen this behaviour on my live system and a proxmox test VM

Forum message thread https://forum.rockstor.com/t/nfs-and-snapshots-not-working-well-together/10584/

I am a new Rockstor user, but am attempting to research a fix myself.

```
NFS Issue 
---------------------

1/ Create Rockstor instance create user with  uid=1000, gid=1000  (NFS needs to keep client and server uid's consistent)
Create BTRFS pool . I used RAID 1.  Create a share called 'share1' with uid/gid=1000

2/ Export share1 with NFS to a client on the same LAN.  Check that files can be uploaded from the client to the share.
Observe the mount point /export/share1 has been created.

3/  Create a snapshot of share1 called snap1 , writable and visible.  Observe that /export/share1/.snap1 has been created and files have correct ownership.
Note also /etc/exports contains a line for the share1 export, and another line for the snap1 snapshot. 

3/ Now delete the NFS export for share1 using the UI.  Observe that the UI shows share1 is no longer exported, and the line for share1 has been removed from /etc/exports
BUT there is still a line for .snap1 in /etc/exports, and /export/share1/.snap1 is still being exported but it contains no files.
(confirmed by exportfs -v)  THIS SEEMS to be a BUG!!

4/ Now recreate the NFS export.  This FAILS with the error message
"Houston, we've had a problem.
An export already exists for the host string: (192.168.1.33).

            NoneType: None"
 No export for /share1 is created, but the pre existing export for /export/share1/.snap1 is still present.
 
 5/ CONCLUSION  The failure to tear down the .snap1 export in step 3 is the root of this problem.  NFS exports 
 and all their children should be totally torn down when a share export is deleted.
```

delboy711:
First contribution by a new user.

Fixes #2995 

The problem solved (hopefully) by this patch is this  -
Once a share is exported by NFS the exported share is mounted at
```
/export/sharename/
```
Any [EDIT:phillxnet "Visible to user? "] see: (https://github.com/rockstor/rockstor-core/pull/2998#issuecomment-3223552099)  snapshots taken after the creation of the NFS export will be automatically exported and mounted at
```
/export/sharename/.snapshotname/
```
If the original NFS share export is deleted, the [EDIT:phillxnet "Visible to user?"] see: https://github.com/rockstor/rockstor-core/pull/2998#issuecomment-3223600106 snapshot export remains and cannot be deleted through the UI.
Any attempt to recreate the NFS export of the share will fail with messages to Houston.
The only way to recover is to go into the terminal and run

```
umount /export/sharename/.snapshotname
chattr -i  /export/sharename/.snapshotname
rmdir /export/sharename/.snapshotname
chattr -i  /export/sharename
rmdir /export/sharename
nano /etc/exports
edit out lines for /sharename/.snapshotname
exportfs -arv
```
This  patch on function delete of storageadmin/views/nfs_exports.py prior to
 deleting a share NFS export, it will iterate through exported snapshots
 of that share,  and  delete them first.

To test it follow this sequence.

- Create a share  'share1'  
- Export share1 with NFS - Observe it is created at /export/share1, the command 'mount' shows it is mounted,  and a line has appeared in /etc/exports.
- Create a snapshot 'snap1' with the 'visible' option checked - Observe a mount point is created at /export/share1/.snap1 and a second line has appeared in /etc/exports.
- Delete the NFS export of share1 - Observe  'mount' shows the mounts at  /export/share1  and /export/share1/.snap1 are gone, and all lines relating to share1 in /etc/exports are gone.  'exportfs -v'  will confirm the NFS exports are deleted.
- The above can be repeated with any combination of multiple shares, and snapshots.  Deleting any share NFS export will  also delete its corresponding snapshot exports.

Limitations -

- Snapshots  created before the share is exported by NFS do not  automatically get exported themselves.  This is the original Rockstor behaviour.
- There are other bugs relating to NFS and snapshots. If NFS shares client lists are edited, or the share ownership is edited.  These bugs  will be addressed in a future PR.  Baby steps...

### DIFF
--- a/src/rockstor/storageadmin/views/nfs_exports.py
+++ b/src/rockstor/storageadmin/views/nfs_exports.py
@@ -208,6 +208,12 @@ class NFSExportGroupDetailView(NFSExportMixin, rfc.GenericView):
             cur_exports = list(NFSExport.objects.all())
             for e in NFSExport.objects.filter(export_group=eg):
                 export_pt = "%s%s" % (settings.NFS_EXPORT_ROOT, e.share.name)
+                for snaps in NFSExport.objects.all():    # #2995 Iterate through current mounts to find snaps and delete them before deleting base share
+                    if snaps.mount.count(export_pt+"/."):
+                        logger.info("Deleting snapshot export {} ".format(snaps.mount))
+                        nfs4_mount_teardown("%s" % snaps.mount)
+                        cur_exports.remove(snaps)
+                        snaps.delete()
                 if e.export_group.nohide:
                     snap_name = e.mount.split(e.share.name + "_")[-1]
                     export_pt = "%s/%s" % (export_pt, snap_name)


Hooverdan96:
This probably needs to be considered in the context of PR #3009, correct? If the suppression is successfully implemented, this piece would only be needed for "legacy" users that created NFS shares before that fix goes in. I don't think that's necessarily a bad thing, just wanted to connect the two. @delboy711, do you agree?

delboy711:
I 'think' (but am not certain)  that exports are created dynamically from saved shares when Rockstor boots. So if a 'legacy' system boots up after implementing #3009, then nfs exports of snapshots will not be created.
If I am correct in this understanding then #2998  is unnecessary since the code would never execute.
I will do some testing to confirm.

delboy711:
After more testing I have confirmed my understanding described in the previous post was not correct.  NFS exports are not dynamically recreated after a reboot.  Any snapshot exports created before an upgrade to #3009  will still be present.
If #2998 is also implemented then it will be possible for the user to delete those snapshots and so tidy up the system.

However it is still possible for the user to run into difficulties.
#2998 operates by deleting snapshot exports of a share when the share is deleted.  If a share had already been deleted before implementing #2998 and #3009 then 'zombie' snapshot exports will already be present with no owning share, requiring the manual procedure listed at the top of this PR to recover.

phillxnet:
@delboy711 Just having  a look at this now. And thanks for the excellent run-down on the intent here. Most helpful.

I'll take this bit-by-bit over a few comments so we have each to reference independently.

Re:

> Any snapshots taken after the creation of the NFS export will be automatically exported and mounted at
> 
> ```
> /export/sharename/.snapshotname/
> ```

This behaviour pertains only to visible snapshots, i.e. on a 5.1.0-0 install.

With a Share (no snapshots) having a default NFS export we get the expected:

```
cat /etc/exports
/export/NFS-share *(rw,async,insecure) 
```

Two snapshots were then created as follows:

  | Name | Created on | Visibility to user | Writable | Total Size | Exclusive Size
-- | -- | -- | -- | -- | -- | --
  | NFS-share-snap2-visible | August 26th 2025, 10:58:46 am | Visible | No | 0 bytes | 0 bytes
  | NFS-share-snap1-default | August 26th 2025, 10:58:05 am | Hidden | No | 16.00 KB | 16.00 KB

With no change in `cat /etc/exports` after the creation of `NFS-share-snap1-default` (invisible by default) ; where-as after the creation of `NFS-share-snap2-visible` we now see:

```
cat /etc/exports
/export/NFS-share *(rw,async,insecure) 
/export/NFS-share/.NFS-share-snap2-visible *(rw,async,insecure,nohide) 
```

As you have detailed. I just wanted to provide the contextual creation of these exports, i.e. at least functionally they are applied only to 'visible' snapshots, i.e. those created with the `Visible to user?` option ticked. From memory this was the intended behaviour. However this does not negate your findings - I just wanted to provide some context in case we are find out we are moving in the wrong direction here :) .

Further-more; deleting the above visible (and auto NFS exported) snapshot also removes it's export:
```
cat /etc/exports
/export/NFS-share *(rw,async,insecure)
```

I'll now edit you original text referencing this comment for context. Then proceed with the review.

N.B. later in original text:

> * Create a snapshot 'snap1' with the 'visible' option checked ...

phillxnet:
> If the original NFS share export is deleted, the snapshot export remains and cannot be deleted through the UI.

Start state, as per mid: https://github.com/rockstor/rockstor-core/pull/2998#issuecomment-3223552099
```
cat /etc/exports
/export/NFS-share *(rw,async,insecure) 
/export/NFS-share/.NFS-share-snap2-visible *(rw,async,insecure,nohide) 
```

Delete NFS Share export, with both a default and visible snapshot created post NFS export in play; and we have our issue reproducer:

```
cat /etc/exports
/export/NFS-share/.NFS-share-snap2-visible *(rw,async,insecure,nohide) 
```

**A rougue/orphaned NFS export of the visible snapshot remains.** The default snapshot (not visible to the use is of no relevance as it was never exported). I'll similarly edit in your original PR text, if nothing else to make what we are changing here easier to re-visit.


N.B. later in original text:

> * Create a snapshot 'snap1' with the 'visible' option checked ...

phillxnet:
> Any attempt to recreate the NFS export of the share will fail with messages to Houston.

Specifically:

> Houston, we've had a problem.
> An export already exists for the host string: (*).
> 
>             NoneType: None

Assumed to be due to the orphaned visible snapshot NFS export indicated in https://github.com/rockstor/rockstor-core/pull/2998#issuecomment-3223600106

phillxnet:
@FroggyFlox @Hooverdan96 I'm inclined to now merge this as per my review. It does what @delboy711 indicated was the aim. And as also indicated there are other fish to fry, and there is a potential interplay here with a later PR from @delboy711 in the same area. But as per @delboy711 https://github.com/rockstor/rockstor-core/pull/2998#issuecomment-3146460429 in response to @Hooverdan96 https://github.com/rockstor/rockstor-core/pull/2998#issuecomment-3140761761 we have:

> If #2998 is also implemented then it will be possible for the user to delete those snapshots and so tidy up the system.

Ergo we can avoid folks tripping over what looks to have been addressed here via a merge and then consider the partially related snapshot mount creation side addressed in PR #3009 .

---

Note there is some confusing references re deleting snapshots when the references intended is to the NFS export of said snapshots. No snapshots were apparently harmed/modified in the application of this code :) .

Hooverdan96:
I vote for merge!

FroggyFlox:
Agreed on merging. 

Thank you all!

PR #3006, opened 2025-07-15
Fix rockstor-bootstrap timeout fail on slower hardware #3005

### LINKED ISSUE #3005: Timeouts too strict on slower hardware
I've recently upgraded my NAS from openSUSE 15.5 to 15.6, including Rockstor. After startup, I noticed the `rockstor-bootstrap.service` not succeeding. 

It was printing the log:
```
Exception occurred while bootstrapping. This could be because rockstor.service is still starting up. will wait 2 seconds and try again. Exception: Exception while setting access_token for url(http://127.0.0.1:8000): HTTPConnectionPool(host='127.0.0.1', port=8000): Read timed out. (read timeout=2). content: None
```
multiple times and giving up. After I changed the timeout to 5 in `api_wrapper.py` it managed to run successfully.

Similarly, I also get logs about `zypper --xmlout list-updates` not completing in 15 seconds. It needs a little more time on my machine:

```
nas:~ # time zypper --xmlout list-updates
<?xml version='1.0'?>
<stream>
<message type="info">Loading repository data...</message>
<message type="info">Reading installed packages...</message>
<update-status version="0.6">
<update-list>
</update-list>
</update-status>
</stream>

real	0m16,391s
user	0m14,376s
sys	0m1,012s
```

I'm running Rockstor on an AMD E-450 with 4GB RAM and an SSD for the OS. It is not the fastest unit but serves me well. Do you want a PR for this?

hberntsen:
Fixes https://github.com/rockstor/rockstor-core/issues/3005

In the above I described that the timeout for the oauth token is too strict on slower hardware. This extends the timeout to 4 seconds read, which is enough for my AMD E-450 system.

---

Minor formatting/wording edit by @phillxnet.

### DIFF
--- a/src/rockstor/cli/api_wrapper.py
+++ b/src/rockstor/cli/api_wrapper.py
@@ -67,7 +67,7 @@ class APIWrapper(object):
                 data=token_request_data,
                 headers=auth_headers,
                 verify=False,
-                timeout=2,
+                timeout=(1, 4),
             )
             content = json.loads(response.content.decode("utf-8"))
             self.access_token = content["access_token"]


phillxnet:
@hberntsen This is nice thanks. And thanks also for exploring the newer requests timeouts.

Our master branch is for stable release maintenance only. Could you please re-present this pull request (close and create a new one), which looks just dandy, but against the testing branch instead. We may end-up cherry picking that commit into master once it's received more field testing; which is the purpose of our testing channel. But ultimately everything in testing ends up in at least the next major stable release anyway - assuming it wasn't reverted.

Sorry to have to ask this, however it should be fairly easy as the two branches are almost identical at this point; given we have just embarked on the next testing phase. I'm hoping we can get your fix out in our first testing release of this new phase.

hberntsen:
No worries! I've changed the base from master to testing. There is no need to create a new PR for that: press edit in the top right corner in GitHub, now you can edit the title of the PR and just below that change the base. I earlier already rebased against testing but forgot to change this one.

FroggyFlox:
Thanks a lot @hberntsen!
Sorry for not having taken a look at your PR earlier. Now that we have our Stable rpm just released, we can start our next Testing phase.

@phillxnet, at first this looks like a safe change, allowing for more flexibility for the hardware that needs it without impairing others. You have a lot more experience than me with this timeout setting, however, so I'll defer to your judgment on this.

phillxnet (approved):
@hberntsen Thanks again for your efforts on this fix. I'm always keen to keep compatibility with slower hardware; and it can also be informative on where exactly we are dragging our heels as it were.

I was hoping to do a real-world test of these changes on a likely reproducer Core2Duo with 2 GB RAM, however with only a USB 2.0 interface for booting on that device currently it would be a too much of a time consuming effort. If I do get around to this test I'll report it in this pull request at a later date - post its merge and closure most likely.

Functionally this change moves us from 2 seconds all-in timeout, to 1 second for initial response (connect timeout), and a total of 4 seconds to complete the request there-after. So we have added fidelity here thanks to our newer requests use from the last testing phase. From your generously reported experiments (in associated issue) this looks to have done the fixing deed and got us out from a slow corner-case exposed by the bootstrapping process and your reproducer hardware. So I'm going with that as good-enough to merge ready to kick-off a new testing phase. Along with:

An rpm was successfully build and installed on a Leap 15.6 OS instance from this branch and looked to behave as expected, including a booting up test and a number of our normal activities: i.e. pool import scub Rock-on service setup / start / stop and subscribing to testing/stable. All of which already worked on the instance used so we have at least no obvious regressions. But once this is out as an early testing release we can get way more field testing underway.

phillxnet:
@FroggyFlox & @Hooverdan96 As this is a simple extension / added fidelity to a requests call timeout settings I'll get this merge as it really needs more field testing (on slower hardware) to know for sure the potential ramifications. But it does look to have fixed an obviously failure which is good. We should also use this newer timeout fidelity in all our requests calls as we come across them in need of maintenance.

hberntsen:
Awesome, thank you!

PR #3009, opened 2025-07-28
Suppress creation of visible snapshot NFS exports #2170

### LINKED ISSUE #1879: Error when editing exisiting NFS share
Whenever editing an NFS share it comes up with the error: "An export already exists for the host string xxxx" where xxxx is the IP for the NFS export.

### LINKED ISSUE #1960: Cant remove share due to stale NFS mount options
I am attempting to remove a share via the UI but it is giving me the following error:

```
Share (Usenet) cannot be deleted as it is exported via NFS. Delete NFS exports and try again.

  File "/opt/rockstor/eggs/gunicorn-19.7.1-py2.7.egg/gunicorn/workers/sync.py", line 68, in run_for_one
    self.accept(listener)
  File "/opt/rockstor/eggs/gunicorn-19.7.1-py2.7.egg/gunicorn/workers/sync.py", line 27, in accept
    client, addr = listener.accept()

```

Reviewing NFS config via Storage > NFS > Advanced Edit I see that the snapshot folders are still mounted in the automatically added entries.  Note there is no NFS export created for the share when viewing the current NFS exports in the UI.  I did 'modify' the previously configured NFS export to use a different share but then ultimately deleted it after I received this error.

Now it just seems to be stuck in limbo and I cannot remove the share until it is removed from the NFS service but I cannot modify the NFS exports in the UI because that share isn't actually being exported anymore (at least not in the UI).

Is it possible to modify the NFS settings in the CLI somehow? I tried modifying the /etc/exports and restart the NFS service but to no avail.  Sort of stuck here now.

### LINKED ISSUE #2170: NFS improperly mounts shares with visible snapshots
As reported in the forum thread [Can’t access NFS shares that have snapshots](https://forum.rockstor.com/t/cant-access-nfs-shares-that-have-snapshots/6766), NFS under rockstor does not currently allow one to effectively use shares that have "visible" read-only snapshots.

Such shares result in snapshots being mounted under `/exports/.<snapshot_path>` and the end user being presented _only_ with snapshots instead of the top-level directories of the share.

In order to use such a share with NFS, one must manually unmount the snapshots mounted under `/exports` before removing them and creating snapshots without the "visible" option selected.

An update to the rockstor snapshot creation page (and scheduled task page) adding a tooltip has been proposed as an interim solution to warn users about visible snapshots under NFS. I am currently exploring the workings of github to try these changes myself and will reference a pull request once I succeed.

### LINKED ISSUE #2742: (t) Error when trying to delete NFS export via WebUI
@FroggyFlox, @phillxnet 

I tried to delete a NFS export that I had previously created using the last stable (4.5.8). In the meantime I activated the test channel and went to 5.0.5 (not sure whether that matters).

when attempting to delete the NFS export, it created an error message:

```
  File "/opt/rockstor/src/rockstor/storageadmin/views/nfs_exports.py", line 147, in refresh_wrapper
    refresh_nfs_exports(exports)
  File "/opt/rockstor/src/rockstor/system/osi.py", line 780, in refresh_nfs_exports
    return run_command([EXPORTFS, "-ra"])
  File "/opt/rockstor/src/rockstor/system/osi.py", line 263, in run_command
system.exceptions.CommandException: Error running a command. cmd = /usr/sbin/exportfs -ra. rc = 1. stdout = ['']. stderr = ['exportfs: /etc/exports.d/rockstor.exports [1]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [2]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/ccc".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [3]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/ddd".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [4]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb/eee".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [5]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb/fff".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [6]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb/ggg".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [7]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb/ggg/evenlongerdirectorynameshere".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [8]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb/ggg//wemakelongdirectorynames".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [9]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb/ggg/wemakelongdirectorynames/anotherdirectoryoverhere".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: /etc/exports.d/rockstor.exports [10]: Neither \'subtree_check\' or \'no_subtree_check\' specified for export "*:/export/AAA/bbb/ggg/wemakelongdirectorynames/anotherdirectoryoverhere/andanotherlongname".', "  Assuming default behaviour ('no_subtree_check').", '  NOTE: this default has changed since nfs-utils version 1.0.x', '', 'exportfs: Failed to stat /export/AAA/bbb/ggg/wemakelongdirectorynames/anotherdirectoryoverhere/andanotherlongname: No such file or directory', 'exportfs: Failed to stat /export/AAA/bbb/ggg/wemakelongdirectorynames/anotherdirectoryoverhere: No such file or directory', 'exportfs: Failed to stat /export/AAA/bbb/ggg//wemakelongdirectorynames: No such file or directory', 'exportfs: Failed to stat /export/AAA/bbb/ggg/evenlongerdirectorynameshere: No such file or directory', 'exportfs: Failed to stat /export/AAA/bbb/ggg: No such file or directory', 'exportfs: Failed to stat /export/AAA/bbb/fff: No such file or directory', 'exportfs: Failed to stat /export/AAA/bbb/eee: No such file or directory', 'exportfs: Failed to stat /export/AAA/ddd: No such file or directory', 'exportfs: Failed to stat /export/AAA/ccc: No such file or directory', 'exportfs: Failed to stat /export/AAA/bbb: No such file or directory', '']

```

The current nfs-utils version seems to be 2.1.1

### LINKED ISSUE #2995: Delete snapshot NFS exports prior to deleting share exports

### LINKED ISSUE #2997: Editing NFS share  clients does not change clients of snapshots
If an NFS share is edited to change the Clients or Admin client, the associated snapshot exports are not altered.
Related to #1879
```
more /etc/exports
/export/share1 192.168.1.*(rw,sync,insecure)  192.168.1.29(rw,no_root_squash)
/export/share1/.snap1 192.168.1.33(rw,async,insecure,nohide)
```

delboy711:
### Background

When first evaluating Rockstor as my potential NAS I did as many people do and just started trying out all the features. When I got as far as trying out NFS shares I quickly ran into problems because I had also created some snapshots, and some of them were marked 'visible'.  It soon became clear that NFS shares and visible snapshots do not play nice together. Forum posts and github issues about problems with NFS went back years although many people had not made the connection with visible snapshots.

Fixes #2170

[Update]
Fixes #2997 
Exports of snapshots are no longer created.

Fixes #2742 
Since #2742 has not had a second reproducer and the symptom is certainly similar to this issue it should probably be closed.

Fixes #1960 

Fixes #1879 
[\Update]

On investigating it became clear that the root of most of these problems is the way exports work with NFS shares that is different to how exports work with samba or sftp shares.
The structure of the file system is exactly the same for nfs as samba, or sftp.
For example
```
/export
└── share1
    ├── .snap-share1-1
    │   └── screen02.png
    ├── .snap-share1-2
    │   ├── .snap-share1-1
    │   ├── Autumn.bmp
    │   └── screen02.png
    ├── Autumn.bmp
    └── screen02.png
```
Visible snapshots are mounted as hidden folders on the parent folder. Successive snapshots contain the contents of older snapshots at the time the snapshot was made.  A user mounting the base folder is therefore able to read the contents of any snapshot if it is marked visible and to write into snapshots if they are marked rw permissions.
What is unique about an NFS share is that there is also an NFS export of every individual snapshot. so in this example
```
exportfs

/export/share1   *.local            #where *.local is the client range (Any host with avahi on the local network)
/export/share1  192.168.1.22    #where 192.168.1.22 is the admin host if present
/export/share1/.snap-share1-1 *.local
/export/share1/.snap-share1-1 192.168.1.22
/export/share1/.snap-share1-2  *.local
/export/share1/.snap-share1-2 192.168.1.22
/export/share1/.snap-share1-2/.snap-share1-1  *.local
/export/share1/.snap-share1-2/.snap-share1-1 192.168.1.22
```
Only the first two entries in this list are useful. All the others are just subsets of the same file system.

Problems arise when the user tries to change things with the share. For example:
#2995 If an NFS share is deleted the exports of the snapshots remain.
#2997 Editing the owner of a share does not change the owner of its snapshots
#1879 Editing the client string of a share does not change the client string of a snapshot
Note:  I have already submitted a PR to fix #2995  That PR is made redundant by this PR which offers a more elegant and comprehensive solution.

### Solution Offered
The fix found for all these issues turned out to be astonishingly simple. It is simply to **NOT**  create NFS exports for snapshots.   Then there is then no need to worry about creating/deleting NFS exports, or maintaining ownerships and permissions among dozens of exports.  Rockstor already does a very good job of mounting, and managing snapshots in the file system since that is what it already does for SFTP and Samba shares.
The attached patch simply removes the code to create snapshot  NFS exports in the file snapshots.py.

The code to delete NFS snapshot exports remains as 'cruft' because I did not want to break anything by removing too much.  It can be removed at a future date.

### Testing
Testing of this PR is quite simple

1. Create a share share1 with ownership uid1000  so a client with uid1000 will be able to write into the share. Create an NFS export giving the IP address of your client machine as the client string.  Mount the share on your client and confirm you can write/read into the share.
2. Once you are confident you have a working NFS share create a visible snapshot. Observe that a hidden folder with the snapshot name has appeared in your share on your client machine. Add more files into the base share.
3. Create a second snapshot. Observe the snapshot has appeared in the base share.  Check the file structure on Rockstor with 'tree -a'.  Check the NFS exports with 'exportfs'.
4. Delete your second snapshot.  It should delete from the base share.
5. Delete the first snapshot.
6. Now add the snapshots again, then edit the share to change the ownership.  Confirm the snapshots ownership has also changed.
7. Delete the share without first deleting the snapshots. Observe the snapshots get deleted and the file system is as expected.
8.  Go mad and try to break it!  Anything that works on an SFTP or Samba share should also work with NFS.
Note: while it is OK to export an NFS share multiple times with different client lists. If you edit a share to give it a client which is included in another export of the same share you will see an error message. This is correct functionality.

### Impact on existing users
If there are any users who already use NFS exports with visible snapshots (which would surprise me since it is so buggy), then migrating to this new operation will not affect existing NFS shares so long as only the base share is mounted. If a snapshot is mounted directly then that mount will no longer work. 

Managing existing NFS shares will operate as it always has with the same buggy operation. Once shares and snapshots have been deleted, then new ones can be created without the buggy operation.

### DIFF
--- a/src/rockstor/storageadmin/views/snapshot.py
+++ b/src/rockstor/storageadmin/views/snapshot.py
@@ -87,19 +87,12 @@ class SnapshotView(NFSExportMixin, rfc.GenericView):
         export_pt = snap_mnt_pt.replace(settings.MNT_PT, settings.NFS_EXPORT_ROOT)
         if on:
             mount_snap(share, snap_name, snap_qgroup)
-
-            if NFSExport.objects.filter(share=share).exists():
-                se = NFSExport.objects.filter(share=share)[0]
-                export_group = NFSExportGroup(
-                    host_str=se.export_group.host_str, nohide=True
-                )
-                export_group.save()
-                export = NFSExport(
-                    share=share, export_group=export_group, mount=export_pt
-                )
-                export.full_clean()
-                export.save()
-                cur_exports.append(export)
+            ########################################################
+            # #2007 #1879   For many years the operation of NFS shares has created an NFS export for every snapshot
+            # This has led to numerous bugs with NFS
+            # To simplfy NFS operation this facility has been removed and now NFS behave like SFTP and Samba shares
+            # with an export for only the top level share with snapshots appearing as folders in the share.
+            #########################################################
         else:
             for mnt in (snap_mnt_pt, export_pt):
                 try:


phillxnet:
@delboy711 I'l like soon to move onto the review of this PR. Could you edit the original text to include a:

> Fixes #

on it's own line, with the number of the issue it addresses. If it addresses multiple existing issues then do multiple `Fixes #` lines. If one does not exist then please create one. We need these  issue/pull-request pairings as our changelog uses them for attribution and to collect what the problem/enhancement was understood to be; matched to what was proposed as the fix/instantiation of the enhancement. In retrospect these are sometimes not what was understood at the time and so it servers us well to leave a trail of intend & action as it were.

[UPDATE] Note: I've now added a `Fixes #2170` as per:
- @FroggyFlox https://github.com/rockstor/rockstor-core/issues/2995#issuecomment-2927195585
- @delboy711 response https://github.com/rockstor/rockstor-core/issues/2995#issuecomment-2927570185
Which petained to the removal of these NFS exports of associated snapshots - but this PR looks to be the actual fix: not creating them in the first place as @mattyvau looks to have identified as failure cause in their #2170 issue from a while ago.

---

Note we also title our commits ending in the "#issue-num they relate to, and similarly with the pull request title container them.  The Web-UI changelog display then turns these into links to the issue/PR.

phillxnet:
@delboy711 Re:

> Note: I have already submitted a PR to fix #2995 That PR is made redundant by this PR which offers a more elegant and comprehensive solution.

Just wanted to clarify this point a little. This PR avoids the auto NFS export creation of visible snapshots, but your last PR (now merged) improves our behaviour re tidying these up if they already exist, which as you indicated here leads to some of the problems encountered by them being created in the first place.

These auto-created NFS exports of visible snapshots made of the parent Share post export have been quite the riddle to unravel, well done on trying to link, as requested, to all affected issues. I've chosen #2170 as our main issue here and have modified the title accordingly. Our Web-UI changelog parsing can currently only handle a single issue, but it is now associated with the others via the GitHub links.

---

Linking to the following code highlight that surfaced from @mattyvau forum thread:

- https://forum.rockstor.com/t/cant-access-nfs-shares-that-have-snapshots/6766/8

which is addressed via this PR.

phillxnet:
> Only the first two entries in this list are useful. All the others are just subsets of the same file system.

A quick note on this one, all btrfs subvols (Shares, Snapshots, Clones in Rockstor speak) of the same parent Volume (Pool in Rockstor parlence) are of the same filesystem in a btrfs volume. And also in btrfs a subvol is presented as a directory entry. All a bit confusing to not only us but apps like df etc.

phillxnet:
@delboy711 Re:

> What is unique about an NFS ~~share~~[EDIT: export within Rockstor] is that there is also an NFS export of every individual [EDIT: writable] snapshot.

This looks to be the nub of our issues here. Especially since it was not fully implemented, i.e. yours and others findings regarding such things as changing those auto-created mounts in-sync with the parent for example.

phillxnet:
Re:

> The code to delete NFS [EDIT: writable] snapshot exports remains as 'cruft' because I did not want to break anything by removing too much. It can be removed at a future date.

We do have to consider here the state inherited by legacy systems. So we will have to maintain a backward 'address' here. This is partly why I have merged your prior PR (#2998) as it did just that.

But I entirely agree that we should try to approach even large issues in defined steps. Not always easy when we are changing major framework elements such as we are in the middle of currently. But this issue does change our behaviour - which is why I propose we leave, at least for now,  your comment referencing changed behaviour re the removed if clause.

delboy711:
> Re:
> 
> > The code to delete NFS [EDIT: writable] snapshot exports remains as 'cruft' because I did not want to break anything by removing too much. It can be removed at a future date.
> 
> We do have to consider here the state inherited by legacy systems. So we will have to maintain a backward 'address' here. This is partly why I have merged your prior PR (#2998) as it did just that.
> 
> But I entirely agree that we should try to approach even large issues in defined steps. Not always easy when we are changing major framework elements such as we are in the middle of currently. But this issue does change our behaviour - which is why I propose we leave, at least for now, your comment referencing changed behaviour re the removed if clause.

The cruft I am referring to here is for example in function  ```refresh_nfs_exports()``` in file   system/osi.py  where there is code referring to exports of NFS snapshots which becomes redundant.

phillxnet:
@delboy711 Re:

> The cruft I am referring to here is for example in function `refresh_nfs_exports()` in file system/osi.py where there is code referring to exports of NFS snapshots which becomes redundant.

Are OK. Good point. Could you create an issue, referencing this one: indicating that we now have orphaned functionality to be addressed at a later date, along with the removal of the comments left by this PR; assuming we have not reverts in our future. Normally, when we catch such things earlier, we try to note them via Python TODO comments: https://www.jetbrains.com/help/pycharm/using-todo.html

No worries here thought as I'm in the last stages of testing what we have to-date, and it is good to have as few a changes as possible - especially as we are modifying behaviour. All-in I like this change and I don't think we will have need to revert - but we can see how this pans out in the wild as it were.

phillxnet:
# Testing

An rpm was build using this branch against current testing branch (i.e. including #2998) on a Leap 15.6 host which subsequently installed successfully.

## No visible snapshot NFS export created:

An existing no-snapshots Share was created and exported via NFS.

```
cat /etc/exports
/export/NFS-share2 192.168.2.173(rw,async,insecure) 
```

1. A non visible snapshot was created on this Share - as before there were not NFS ramifications.
2. A visible snapshot was created - and unlike prior to these changes (see: https://github.com/rockstor/rockstor-core/pull/2998#issuecomment-3223552099) we have no `visible` snaphost NFS export auto-created.

```
cat /etc/exports
/export/NFS-share2 192.168.2.173(rw,async,insecure) 
```

But we do have an additional mount created for this snapshot.

```
mount | grep export
/dev/vda on /export/NFS-share2 type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=877,subvol=/NFS-share2)
/dev/vda on /export/NFS-share2/.NFS-share2-snap1-visible type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=879,subvol=/.snapshots/NFS-share2/NFS-share2-snap1-visible)
```

@delboy711 So we are still executing a mount of the snapshot here, but only the visible one. However we are no longer additionally exporting this visitble snapshot: the stated goal of this PR.

phillxnet (approved):
@delboy711 Nice find, and much appreciated.

I like that we are now treating the NFS export more similarly to how we do SMB. And it does look like our prior visible snapshot NFS exports were miss-guided; likely due to considering subvols as filesystems in their own right - but they share metadata with their ultimate parent volume.

It is as a rule-of-thumb not our practice to quote non-existent code or behaviour. But as you have indicated - we have some remaining artefacts - so lets leave the quote so we can get this fix into the wild. And I see you have already created: #3015 to address this and further tidy-ups. Thanks.

# Testing
See my https://github.com/rockstor/rockstor-core/pull/3009#issuecomment-3234011650

phillxnet:
@FroggyFlox & @Hooverdan96 I'm think this is now game to go as it were? A change of behaviour, especially long standing, may well come back to haunt us - but this does look to do-away with a whole family of issues! And was never fully implemented anyway i.e. no subsequent maintenance of said auto exports.

FroggyFlox:
Agreed for merging. This has resulted in a few issues reported in the forum over the years so I'm happy by the prospect of having that behind us.

phillxnet:
@delboy711 I've closed #2170 as per our usual messaging. Do you fancy doing the same messaging on all the issues linked that you think this has fixed. Or indicate your prior NFS fix if that is more fitting. I can then close the ones not created by you against the appropriate PR. Thanks. You can of course close the ones created by yourself that you deem appropriate.

Thanks gain for all this NFS work, it's been a long time since it has received such attention unfortunately.

PR #3051, opened 2025-10-25
SMB Admin user 3 char minimum validator #3041

### LINKED ISSUE #3041: SMB Admin user 3 char minimum validator
Desc:
While entering a username in any grant-access input field on Rocstor WebUI ie, Samba Export & Admin(s) field, then there is a validator which says "Please enter 3 or more characters" to find or enter a user.

Issue:
If we have a valid user with only 2 characters username like the "ha" (ie, HomeAssistant) then this UI validator in Rockstor UI didn't allow to add that user to the grant access list.

Solution:
Change validator requirements to allow to find users with minimum 2 characters in the username instead of minimum 3 characters.

Image

tsx4k:
Fixes #3041

Changed the Select2 validator limit from minumum 3 characters long to minumum 2 characters long for finding/entering a user to the Samba Export & Admins input field.

This allows to search and enter a short named users like "ha".

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/views/add_samba_export.js
+++ b/src/rockstor/storageadmin/static/storageadmin/js/views/add_samba_export.js
@@ -167,7 +167,7 @@ AddSambaExportView = RockstorLayoutView.extend({
             return null;
         }
         this.$('#admin_users').select2({
-            minimumInputLength: 3,
+            minimumInputLength: 2,
             allowClear: true,
             placeholder: '',
             matcher: matchCustom


phillxnet (approved):
@tsx4k Thanks for taking the time to see to this improvement/bug-fix. It does seem like the original minimum of 3 characters in the username field, in an attempt to alleviate a performance issue re chromium (with very large user counts) is now dated and had little benefit in the end. Also this was at the cost of fencing off the 2 character username option (bug) you have addressed here.

# Testing
An rpm was build and installed on a Leap 15.6 host using this branch. The result looks to be as intended; an updated tooltip, and enabling the entry of a 2 character minimum user name:

smb-admin-user-char-min-tool_tip

smb-admin-2-char-user

phillxnet:
@tsx4k Thanks again for jumping through a few hoops to get this in. Cheers.

PR #3055, opened 2025-11-14
Py3.11 crypt library will be deprecated #2572

### LINKED ISSUE #2572: Py3.11 crypt library will be deprecated
This is actually not applicable to Py3.6 yet but will be in higher versions - 
according to [PEP-0594](https://peps.python.org/pep-0594/#rationale) the crypt library will be deprecated (starting in 3.11). Just want to put a placeholder here that we need to consider replacing the use of the `crypt` library, looking at replacements like an implementation of something like [argon2](https://argon2-cffi.readthedocs.io/en/stable/index.html) or some other module that specializes in password salting/hashing (with emphasis on password).

The `crypt` modules is currently in use here:

https://github.com/rockstor/rockstor-core/blob/e8c98b6af5490d09f4b93b947fd025d94d759781/src/rockstor/system/users.py#L163-L177

Interestingly, as part of the 3.6 upgrade work in progress, in the `poetry.lock` file `argon2` (and `bcrypt`) show up as an `package.extras` already:
https://github.com/rockstor/rockstor-core/blob/1d4b9d5acd7236fc74d5c86f2cbfbb2c98326490/poetry.lock#L44-L47
or even in the current testing channel

https://github.com/rockstor/rockstor-core/blob/e8c98b6af5490d09f4b93b947fd025d94d759781/poetry.lock#L58-L60

as part of the `django` package.

phillxnet:
Replace the Python `crypt` module for password crypted hash generation, with a subprocess call to the `mkpasswd` command: which is similarly a features-rich front end to the password encryption OS function of crypt. Ensure the password, and its encrypted hash, are handled only via pipes (stdin & stdout). Avoiding exposure via command options.

Includes a timeout addition to the generic run_command() subprocess.Popen() wrapper. Plus a minor log normalisation re smbpasswd.

Fixes #2572 

---

[EDIT] An alternative approach to using `mkpasswd` from the whois package, would be to directly call openssl: 
```
openssl passwd -6
Password: 
Verifying - Password: 
$6$mIJ2527tmNnh1o2v$9AM7RdGZn0w573SCDMF8K0oZHbXmHNFeKJmVnQv696ofyrg2FhCqO42lNxL83ajsYUJ.gxQnzdVP0BKJVMV0M.
```
However the resulting hash generation has not been tested. Leaving this to a future possible enhancement.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -249,6 +249,7 @@ def run_command(
     log: bool = False,
     pinput: AnyStr | None = None,
     raw: bool = False,
+    timeout: int | None = None,
 ) -> (list[str] | str, list[str], int):
     try:
         # We force run_command to always use en_US
@@ -269,7 +270,12 @@ def run_command(
             env=fake_env,
             universal_newlines=True,  # 3.7 adds text parameter universal_newlines alias
         )
-        out, err = p.communicate(input=pinput)
+        try:
+            out, err = p.communicate(input=pinput, timeout=timeout)
+        except subprocess.TimeoutExpired:
+            logger.error(f"Timeout running command ({cmd}).")
+            p.kill()  # Because timeout does not kill the child process.
+            out, err = p.communicate()
         # raw=True allows parsing of a JSON output directly, for instance
         if not raw:
             out = out.split("\n")
--- a/src/rockstor/system/users.py
+++ b/src/rockstor/system/users.py
@@ -14,7 +14,6 @@ General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
-import crypt
 import logging
 import os
 import pwd
@@ -39,6 +38,8 @@ GROUPADD = "/usr/sbin/groupadd"
 USERDEL = "/usr/sbin/userdel"
 GROUPDEL = "/usr/sbin/groupdel"
 USERMOD = "/usr/sbin/usermod"
+MKPASSWD = "/usr/bin/mkpasswd"
+XARGS = "/usr/bin/xargs"
 SMBPASSWD = "/usr/bin/smbpasswd"
 CHOWN = "/usr/bin/chown"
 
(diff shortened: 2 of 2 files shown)


phillxnet:
An important part of this replacement of the Python crypt module with an OS package is that they are both interfaces to the system 'crypt' element of glib (libcrypt.so.1) - or a stand-in/replacement such as libxcrypt / OpenSSL, or what-ever.

But in some cases - where password hash generation is time based - failing through to possible python only hash generation risks taking an impractical amount of time in comparison to the compiled language alternatives found on the OS itself.

I.e. we should strive to keep as close the OS as possible on this front given some Python fall-through would be entirely impractical. I.e.: from https://passlib.readthedocs.io/en/stable/lib/passlib.hash.scrypt.html#scrypt-backends in the case of Scrypt hashing within the apparently orphaned passlib we have:

> Scrypt Backends
> This class will use the first available of two [EDIT 'three'] possible backends:
> 
> 1. Python stdlib’s hashlib.scrypt() method (only present for Python 3.6+ and OpenSSL 1.1+)
> 2. The C-accelerated scrypt package, if installed.
> 3. A pure-python implementation of SCrypt, built into Passlib.
>
> Warning If hashlib.scrypt() is not present on your system, it is strongly recommended to install the external scrypt package. The pure-python backend is intended as a reference and last-resort implementation only; it is 10-100x too slow to be usable in production at a secure rounds cost.
> Changed in version 1.7.2: Added support for using stdlib’s hashlib.scrypt()

I.e. a 10 to 100x slow-down potential if we find ourselves falling back, via a python wrapper, to a pure-python implementation given password hashing current practice.

phillxnet:
# Testing
On a Leap 15.6 X86_64 system we have the following result from a default CLI useradd:

```
useradd default
passwd default
cat /etc/shadow | grep default

default:$6$1K/yo2ZH8qfgiW5S$TDvuM8TmVUyQO3DEl4gp48oWP1rXb.4MY2M4yQK.kkCaQfT75MsKur4S85aKK/kLQ4CMFLoHNukFL0tEowzHc0:20407:0:99999:7:::
```
Web-UI user addition `pru3055`, or password change with an rpm build from this, and the paired rockstor-rpmbuild pr branch gives:
```
cat /etc/shadow | grep "default\|pru3055"

default:$6$1K/yo2ZH8qfgiW5S$TDvuM8TmVUyQO3DEl4gp48oWP1rXb.4MY2M4yQK.kkCaQfT75MsKur4S85aKK/kLQ4CMFLoHNukFL0tEowzHc0:20407:0:99999:7:::
pru3055:$6$yW8HaQujUU4pgOwD$rQPaVE7E88m7/puHc5XUuiMHskXFEruqujhlYHCgj4TddaXyHxx6CLGlLc6.FY2qc8eUtnFPn9tPjxnxcFb4s/:20407:0:99999:7:::
```

- The same as default `$6$` beginning denoting SHA512-Crypt
- A Salt between the above ID header and the next $
- - `'default' - 1K/yo2ZH8qfgiW5S`
- - `'pru3055' - yW8HaQujUU4pgOwD`
- The salted SHA512-Crypt hash of each before the next ":":
- -  `'default' - TDvuM8TmVUyQO3DEl4gp48oWP1rXb.4MY2M4yQK.kkCaQfT75MsKur4S85aKK/kLQ4CMFLoHNukFL0tEowzHc0`
- -  `'pru3055' - rQPaVE7E88m7/puHc5XUuiMHskXFEruqujhlYHCgj4TddaXyHxx6CLGlLc6.FY2qc8eUtnFPn9tPjxnxcFb4s/`

And the Web-UI generated, or password-changed user, can login successfully using the Web-UI set password.
```
rleap15-6 login: pru3055
Password: 
No mail.
Last login: Sat Nov 15 14:25:19 on ttyS0
Have a lot of fun...
pru3055@rleap15-6:~>
```

phillxnet:
N.B. opensuse is one of a few distributions that has supplanted the glibc libcrypt, with libxcrypt:
```
rleap15-6:~ # zypper info libcrypt1
Loading repository data...
Reading installed packages...


Information for package libcrypt1:
----------------------------------
Repository     : Main Repository
Name           : libcrypt1
Version        : 4.4.15-150300.4.7.1
Arch           : x86_64
Vendor         : SUSE LLC <https://www.suse.com/>
Installed Size : 259.0 KiB
Installed      : Yes (automatically)
Status         : up-to-date
Source package : libxcrypt-4.4.15-150300.4.7.1.src
Upstream URL   : https://github.com/besser82/libxcrypt
Summary        : Extended crypt library for DES, MD5, Blowfish and others
Description    : 
    libxcrypt is a modern library for one-way hashing of passwords.  It
    supports DES, MD5, SHA-2-256, SHA-2-512, and bcrypt-based password
    hashes, and provides the traditional Unix 'crypt' and 'crypt_r'
    interfaces, as well as a set of extended interfaces pioneered by
    Openwall Linux, 'crypt_rn', 'crypt_ra', 'crypt_gensalt',
    'crypt_gensalt_rn', and 'crypt_gensalt_ra'.
```

PR #3056, opened 2025-11-15
Python version 3.13 does not have module distutils.util #3054

### LINKED ISSUE #3054: Python version 3.13 does not have module distutils.util
In `src/rockstor/system/osi.py` we use strtobool via:

> from distutils.util import strtobool

We only use `strtobool` in the following code:

https://github.com/rockstor/rockstor-core/blob/5032bddf980e6420cffdfbfcd096ad34290bec00/src/rockstor/system/osi.py#L2321-L2335

Which in-turn has 2 uses (bar import) in src/rockstor/system/network.py

phillxnet:
We only use this module for the trivial strtobool() function. Open code equivalent function in prior user: to_boolean(). Adds parameter and return type hints to to_boolean()

Incidental removal of now dated subprocess import TODO.

Fixes #3054

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -23,13 +23,12 @@ import re
 import shutil
 import signal
 import stat
-import subprocess  # TODO: consider drop in replacement of subprocess32 module
+import subprocess
 import time
 import uuid
 from socket import inet_ntoa
 from struct import pack
 from tempfile import mkstemp
-from distutils.util import strtobool
 from typing import AnyStr, IO
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


phillxnet:
# Testing

```
...
Ran 292 tests in 54.606s

OK
```

If we do suffer a modified behaviour here it will originate from only 2 places in src/rockstor/system/network.py:

> to_boolean(dtmap["Options"]["com.docker.network.bridge.enable_icc"])

> to_boolean(dtmap["Options"]["com.docker.network.bridge.ip_masquerade"]) 

E.g. respectively:

https://github.com/rockstor/rockstor-core/blob/5032bddf980e6420cffdfbfcd096ad34290bec00/src/rockstor/system/network.py#L252-L255

https://github.com/rockstor/rockstor-core/blob/5032bddf980e6420cffdfbfcd096ad34290bec00/src/rockstor/system/network.py#L257-L264

PR #3062, opened 2025-12-31
Multiple gpg-agent & scdaemon instances lingering #3026

phillxnet:
When we timeout the zypper process looking for installable updates, in pkg_updates_info(), under subprocess.run the main zypper process is killed; there-after we receive the TimeoutExpired exception.

Ensure after the above time out events that we kill all zypper associated gpg-agent processes, they auto-regenerate when required, and tidy-up zypper's `var/tmp/zypp.*` directory spamming artifacts.

The scdaemon's are auto terminated as children of the gpg-agents.

Fixes #3026 

---

Thanks to @FroggyFlox for some pointers on this fix.

PR #3063, opened 2025-12-31
Insufficient timeout on zypper list-updates call #3032

### LINKED ISSUE #3032: Insufficient timeout on zypper list-updates call
On version 5.0.15-0 I frequently observe the below timeout error message in the `rockstor.log`:

```
ERROR [system.pkg_mgmt:575] Consider applying updates to reduce backlog: Command '['zypper', '--xmlout', 'list-updates']' timed out after 15 seconds
```
The default timeout for `zypper` itself is 180 seconds.
from `/etc/zypp/zypp.conf`
```
##
## Maximum time in seconds that you allow a transfer operation to take.
##
## This is useful for preventing your batch jobs from hanging for hours due
## to slow networks or links going down. Limiting operations to less than a
## few minutes risk aborting perfectly normal operations.
##
## Valid values:  [0,3600]
## Default value: 180
##
# download.transfer_timeout = 180
```
This also seems to happen when there are not really that many updates to process. Roughly timing within an interactive session, zypper spends most of the time in refreshing and preparing the repositories vs pulling the actual updates, which is probably highly dependent on internet connection/package server performance. While this is not a critical error (should this really be a warning rather than an error) the suggestion is to increase the timeout. Now, should it be 180 seconds, I don't know.
If it's meant to highlight a possible bottleneck in package management (especially due to a large backlog of unapplied updates), maybe moving it up to half the default timeout (i.e. 90 seconds) might be a better choice.

Open to suggestions or other opinions, especially if I misunderstood the intent behind the 15 second timeout.

phillxnet:
Increase the max_wait timeout on `zypper --xmlout list-updates` from 15 to 40 seconds due to excessive timeouts observed in wider testing.

Fixes #3032

---

Thanks to @Hooverdan96 for development assistance on this issue.

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -560,7 +560,7 @@ def zypper_killed_cleanup():
     )
 
 
-def pkg_updates_info(max_wait: int = 15) -> typing.List[dict[str:str]]:
+def pkg_updates_info(max_wait: int = 40) -> typing.List[dict[str:str]]:
     """
     Fetch info on installable updates across all repos via zypper xml output call.
     Resolves as per 'zypper up' excluding packages with dependency problems.


phillxnet:
Testing an rpm build and installed on a Leap 15.6 host using this branch does look to have greatly improved our behaviour here. As came up in the issue discussion we are some-what doing things wrongly here. I.e. we should schedule this installable updates rather than invoking it on every refresh - but delaying by a second to enable the subscription page to get zypper access. But this can be worked on later once we have moved to a more Django native scheduler and can more easily give user feedback on zypper availability in contention scenarious.

Bit by bit. Moving to merge this for now as a better default that what we had re time-out to collect the installable updates. Plus we now have our zypper-killed-cleanup() procedure in place re the recently merged:

- https://github.com/rockstor/rockstor-core/pull/3062

So we are gradually improving our behaviour through-out in this regard.

PR #3066, opened 2026-01-03
Add gid element to Rockons definition #3029

### LINKED ISSUE #3029: Add `gid` element to Rockons definition
Further to https://github.com/rockstor/rockon-registry/issues/445#issuecomment-3052711491 it is proposed that we add this complementary element to our existing `uid` counterpart.

I.e. extend the DContainer model to also include a `gid` identical to the existing "uid = models.IntegerField(null=True)" but pertaining to the group to pass via the docker `--user` directive.

This would essentially extend our existing capability to docker command line option pass:

> --user uid

to

> --user uid:gid

---

Related rockon-registry README.md discussion from which this proposal emerged: https://github.com/rockstor/rockon-registry/issues/445

Model link: https://github.com/rockstor/rockstor-core/blob/626354939b183084ef7ab5e71f320e742122b931/src/rockstor/storageadmin/models/rockon.py#L98-L109

See: src/rockstor/storageadmin/views/rockon_helpers.py

https://github.com/rockstor/rockstor-core/blob/626354939b183084ef7ab5e71f320e742122b931/src/rockstor/storageadmin/views/rockon_helpers.py#L376-L382

[EDIT] Note also:

https://github.com/rockstor/rockstor-core/blob/f9919c8bdf53d6fc5e664e3531de70d5f3d8cc32/src/rockstor/storageadmin/views/rockon.py#L237-L247

phillxnet:
Mirror existing behaviour of uid, including first volume mount point group id when value is -1 for gid. Plus an additional -2 special case for gid only that substitutes the system's "docker" gid during Rock-on generic_install().

If no uid is specified, gid is ignored as there can be no `--user` :group component without first specifying a user id. This way we maintain our explicit use of uid or uid:gid when the first is specified.

This patch adds int only type hinting to the treatment of uid, and the new gid, of the first volume (-1 special case); and no longer returns None but root (docker default) if no volume is mapped.

N.B. A DContainer gid model field addition migration is included.

Incidental changes
- String.format to fstrings on modified code.
- Previously missed nfsexportgroup ordering Meta option migration.

Fixes #3029


FroggyFlox:
@phillxnet, just noting here before I forget about it; my apologies for asking without checking myself first. 

Do we need to also adjust our Django admin accordingly to have the new `gid` attribute displayed? I'm wondering about the following list:
https://github.com/rockstor/rockstor-core/blob/f9919c8bdf53d6fc5e664e3531de70d5f3d8cc32/src/rockstor/storageadmin/admin.py#L806

phillxnet:
@FroggyFlox Cheers. I'd forgotten about updating the admin-site. I'll push the changes to this draft.

phillxnet:
We currently have very few users of the uid element:

UID-and-GID-in-admin-site

phillxnet:
When using the currently draft PR modifications to the Forgejo-runner Rock-on under:

- https://github.com/rockstor/rockon-registry/pull/497

We have, post a Rock-on update button use, the following entries in the DB re DContainers:

Forgejo-runner-gid-use

phillxnet:
# Testing

Using the linked draft exmple implementation presented in:

- https://github.com/rockstor/rockon-registry/pull/497

The indicated Rock-on, post the modifications here and in the above rock-on definition,  installed and successfully registered with codeberg.org as a self hosted runner. We also have the following indication of group attribution via the dynamically assigned (via `gid: -2` json element) docker group:

```
grep "docker" /etc/group
docker:x:477:
dockertest:x:444:

docker inspect forgejo-runner | grep "\"User\":"
            "User": "1000:477",

docker exec -it forgejo-runner sh
~ $ id
uid=1000(runner) gid=477 groups=477
```
Where-as the default is 1000:1000 as inherited from the FROM image used by the sited wrapper Rock-on image.

Note also that we see, via debug rockstor logging, thefollowing during install:

```
[13/Jan/2026 16:37:44] DEBUG [system.osi:261] Running command: /usr/bin/docker pull codeberg.org/phillxnet/forgejo-runner-wrapper:latest
[13/Jan/2026 16:37:45] DEBUG [system.osi:261] Running command: /usr/bin/docker run -d --restart=unless-stopped --name forgejo-runner -v /mnt2/runner-data:/data -v /etc/localtime:/etc/localtime:ro --user 1000:477 -v /var/run/docker.sock:/var/run/docker.sock ...many-more-unrelated-options
```
I.e. we see the dynamically assigned `--user 1000:477` via our uid: 1000 and gid: -2 interpretation.

## Before
Prior to the changes here and in the linked rockstor-core repo we get the following results (using a different install):
```
docker:x:478:

docker inspect forgejo-runner | grep "\"User\":"
            "User": "0",

docker exec -it forgejo-runner sh
~ # id
uid=0(root) gid=0(root) groups=0(root),0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)

~ # id runner
uid=1000(runner) gid=478(docker) groups=478(docker),478(docker)
```

PR #3076, opened 2026-02-16
Ignore spurious error when setting APM #3075

### LINKED ISSUE #3075: APM settings: consider ignoring specific spurious errors
Thanks to forum member [redshift](https://forum.rockstor.com/u/redshift) for investigating and raising this issue seemingly affecting certain Western Digital and Seagate drives.

Proposal is to ignore specific errors during setting of APM spindown parameters, where the WebUI will prevent setting the parameter, while when running the underlying instruction directly at the command line, it will succeed. If the current response from the `hdparm` command (`SG_IO: bad/missing sense data`) can be considered spurious (i.e. not impacting the actual expected functionality) and not a "real" error, this could be possibly incorporated into the WebUI logic.

For more details, see the forum discussion:

https://forum.rockstor.com/t/apm-not-persisted-due-to-spurious-drive-error/10857

tjkirch:
Fixes #3075

See friendly forum thread: https://forum.rockstor.com/t/apm-not-persisted-due-to-spurious-drive-error/10857

In short, when running hdparm commands on certain drives (at least a couple of Seagate and Western Digital USB drives) it will return 0 but give a spurious error when setting some options, including the APM setting `-B` needed for spindown to work.  The operation does work regardless of the message in stderr.

Before this change, I had to use a systemd service override to force in the `-B 127` setting.  Trying to set it in Rockstor's spindown interface did nothing because of the conditional being modified in this PR, and the error would show up in Rockstor logs.

This change checks for the specific spurious error message and continues if that's all it sees and the command returned 0.  (The new function could be used in other places if the same spurious error is preventing other actions, but I haven't run into any others.)

**Testing:**

I tested the new function on its own against a bunch of good and bad arguments to be sure it only excused the specific error under discussion.  Then I tested on my Rockstor server and confirmed that the systemd service has the requested `-B` option without an override.

### DIFF
--- a/src/rockstor/system/osi.py
+++ b/src/rockstor/system/osi.py
@@ -1567,6 +1567,19 @@ def get_disk_APM_level(dev_byid):
             return level
     return 0
 
+def hdparm_stderr_ok(stderr: list[str]) -> bool:
+    """
+    For some hardware, hdparm prints error messages to stderr even when it's
+    able to successfully perform the request.  This function takes the stderr
+    output from hdparm and returns True if we believe the command ran OK, i.e.
+    the error text was spurious.  Returns False if it seems to be a real error.
+    :param stderr: The stderr output from hdparm, split on newlines into a list.
+    :return: True if the command seemed to run OK, meaning any stderr error
+    message was spurious, and False if there seemed to be a real error.
+    """
+    spurious = r"SG_IO: bad/missing sense data, sb\[\]: ( [0-9A-Fa-f]{2}){32}"
+    return len(stderr) == 1 or (len(stderr) == 2 and
+                                re.fullmatch(spurious, stderr[0]) is not None)
 
 def set_disk_spindown(
     dev_byid, spindown_time, apm_value, spindown_message="no comment"
@@ -1620,7 +1633,7 @@ def set_disk_spindown(
         # Try running this -B only hdparm to see if it will run without
         # error or non zero return code.
         out, err, rc = run_command(hdparm_command, throw=False)
-        if rc == 0 and len(err) == 1:
+        if rc == 0 and hdparm_stderr_ok(err):
             # if execution of the -B switch ran OK then add to switch list
             switch_list += apm_switch_list
         else:


phillxnet (changes requested):
@tjkirch Thanks for such as considered and focused pull request.

My apologies that it has taken us so long to review. We are about to start a new testing phase and, given you diligence here, I think it would be safe, and great, to get this into that early testing phase for field testing.

This type of change is a tricky one to actually test - bar what you have done already. And the hardware 'quirk' nature we see re hdparm's response does look like a safe thing to side-step in our current blanket refusal if any errors are returned. At least from now on we have a mechanism of sorts that may have to incorporate future false positives re incompatible hdparm -B settings.

Thanks also for all the explanation and the in-code comments/explanations, plus type hints on this one. And on the type hints note alone I have a small request:

> hdparm_stderr_ok(stderr: str) -> bool:

Expects a string, however "err" as returned by run_command() and as interpreted (though ambiguously for string first char or list first element) by your:

> stderr[0]

is actually intended to be typed as a list[str] type:

https://github.com/rockstor/rockstor-core/blob/efa86620fe6b6963f6647b274f50bf08704e69c6/src/rockstor/system/osi.py#L252

Though I appreciated we need to refine the type-hinting there also as it look to be defaulting to Any!! Hence no useful IDE feedback on this.

Image

Could you correct your str type-hint to be list[str] and we can then merge this ready for the next rpm released in testing. This way we help to avoid tripping ourselves up in the future by inadvertently expecting a string when looking, in isolation, at expanding this new hdparm_stderr_ok() function. Please also git squash and rebase this change on latest testing and re-present to help keep our git history tidier. We are only talking about a type-hint change here after-all so we need not incorporate that in our history with code that is as-yet un-merged. Plus this is all in the context of work-needed on the run_command() return type. But that is for another issue & PR I think.

Thanks again for all your efforts here, and I do fully appreciate that these are only hints - and in the context of a dysfunctional run_command() return type hinting. But best we try to hang on to what type rigour or suggestion we currently have :) .

tjkirch:
No problem, and thanks for your review!  I've corrected the type hint and the `:param` comment to match, then rebased on testing.  (I had originally developed the function to take a string before finding out about the `run_command` return, as you suspected, and I forgot to change the type hint.  Sorry about that.)

phillxnet (approved):
@tjkirch Thanks for the rapid follow-up on this one.

Re: my first review: https://github.com/rockstor/rockstor-core/pull/3076#pullrequestreview-3948914853

> Could you correct your str type-hint to be list[str] and we can then merge this ready for the next rpm released

Thanks for noticing and update the param entry also, much appreciated.

Moving to merge; I also re-tested an rpm build and install on a Leap 15.6 and all looked OK, as per the last review.

Issue #3072, opened 2026-01-20
Forget Pool feature


phillxnet:
An issue abstracted from the conversation had previously regarding the previously auto-imported system (root) subvolume: https://github.com/rockstor/rockstor-core/issues/2974#issuecomment-2708250136

> It would be a great Web-UI facility that need-be no-more than a button initiating a dialog explaining all of this Pools associated Share/Snapshot/Clone/Export/Task/Rock-on Web-UI elements will be removed. With a proviso that the Pool, and its data, can be re-imported there-after. This would make for a nice usability improvement. The back-end of this should be a simple DB operation: assuming we have all our model cascade deletes lined up correctly. Tasks and some other non Foreignkey related things will require some cascade opperations of our own: i.e. removing associated cron jobs etc; presumably via the appropriate subsystems delete functions. But we must tread carefully here as some delete functions are associated with data removal !!! This function needs to be entirely Web-UI/DB constrained only with regard to actual Pool content. But we don't want to leave a mes behind regarding system configurations. We may even want to consider some kind of ro status for the affected pool, prior to enacting this function. That way we can block inadvertent buggy overreach when instructing as-yet unprepared subsystem related to undoing their configuration.

It is proposed that we add an un-import button, or a Forget Pool button that effectively reverses the import process: i.e. one could then unimport a Pool and all associated shares exports etc.

The facility here pertains to removing Rockstors managmenet of a Pool so that it could be re-imported or left unmanaged as the user requires.

phillxnet:
Promoting this feature to our next Milestone, once we have our current test cycle restarted milestone of https://github.com/rockstor/rockstor-core/milestone/30 done and dusted.

Issue #3058, opened 2025-11-21
Move Share chown chmod operations to the scheduler


phillxnet:
Greatly informed by the expositions in both: https://github.com/rockstor/rockstor-core/issues/2976 and the following forum thread:

- https://forum.rockstor.com/t/unknown-internal-error-doing-a-post-to-api-shares-7-acl/10797

We have a show-stopper for large file counts (and/or snapshots) re chown & chmod recursive operations. From the former https://github.com/rockstor/rockstor-core/issues/2976#issuecomment-2727312818 we have @slowhand93  stating for example:

> As far as I remember it took some time. Definetly not 1-2 minutes. Maybe 10-20.

And from for examlpe: https://forum.rockstor.com/t/unknown-internal-error-doing-a-post-to-api-shares-7-acl/10797/5

we have:

```
time chown -R media:users /mnt2/Media

real    3m45,956s
user    0m0,312s
sys     2m58,541s
```

which is attempted via our API in the foreground/Web-UI process. With obvious failures/response issues resulting.

Ergo we, at least until we move to replace our current Huey scheduler with for example Django-Q2, move to a Huey task; ensuring we make no object  references within that schedule: just the raw commands themselves. There are threading concerns with accessing Django objects with our current Huey scheduler: hence the proposed move to a more Django native scheduling system.

Hooverdan96:
Out of curiosity and I don't know the answer re: my comment about task conflicts while these operations are executed (irrespective of whether they are in fore or background): have you seen issues reported elsewhere where a problem arises if a new/changed file is written while the `chown` or `chmod` is going on? Or is that overthinking it since it's a "file by file" type of operation (after all it's not like a database table where I would want to ensure that _all_ records are locked before executing some mass update/change)?

phillxnet:
@Hooverdan96 Re:

> ... have you seen issues reported elsewhere where a problem arises if a new/changed file is written while the `chown` or `chmod` is going on? Or is that overthinking it since it's a "file by file" type of operation

This is outside our remit really. I.e. this question pertains to all filesystem users, of which we are a member. I.e. it is not our place to delegate app-to-filesystem interactions, but down to the kernel and filesystems themselves. So kind of out of our hands.


> it's not like a database table where I would want to ensure that _all_ records are locked before executing some mass update/change)?

[...]

PR #3079, opened 2026-02-20
Pool import not reflecting owner:group, perms, and compression #3073

### LINKED ISSUE #3073: Pool import not reflecting owner group permissions & compression
A partially completed design implementation: the 'Source of Truth' re owner:group and permissions within the Web-UI has resulted in a long standing anomaly: essentially, prior Pool imports not reflecting the disk rights - which we simply don't yet observe on import. The rights are still active but now out-of-sync re disk state in Web-UI.

This also pertains equally to CLI share owner:group or permission changes not being reflected in the Web-UI on next refresh.

See also: https://forum.rockstor.com/t/file-permissions/8813/13 where the resulting user created issue of #2554 unexpectedly drew together at least two overlapping failures. This issue pertains to the remaining short-fall indicated above.

The main issue here is to reflect on-disk info: for some system elements the DB is the SOT (Source of Truth) and our place is then to enforce what is in the DB to the system. But it has always been the intention re the Pool for it to alwatys be the SOT: hence we import pools to inform the DB and only change pool state on Web-UI interaction: but reflect any effect found on pool (ours or others) in the Web-UI.

Ergo fix disk owner:group permissions to be strictly bottom-up. Web-UI interactions/requests may temporarily alter DB to reflect ongoing/initiated intention: but ultimately they should revert to the SOT of on-disk. I.e. during a system re-install we import a pool and then apply a config restore. The import needs to update the DB fully re Share owner:group permissions.

phillxnet:
Establish source of truth as on-disk re Share; owner, group, permissions, & compression.

Avoid using owner:group, permissions, & compression_algo Share model defaults with user-created (http: POST) Shares; where possible. Resourcing os.stat and subvol property on newly created subvols instead. There-by tracking OS defaults.

During Pool/Share DB import/update:

- Use os.stat for user:group & permissions.
- Use get_property() to read subvol compression properties.

## Incidental changes:

- Use a pool.save() for foreign_key update before newly discovered share.save() in import_shares().
- Comment updates/improvements and more string.format to fstring changes.
- TODO: and comment additions re pending Huey scheduling of chown & chmod.
- Python datime change re timezone awareness.
- Initial testing re subvol as source-of-truth re defaults for owner:group, perms, and compression.
- On ShareACLView post (submit), call os.stat to establish subvol defaults for; owner:group, perms when not specified. Adds Share DB model
  update re compression while we are handling a Share instance.
- Employ update_fields on minor Share updates during ACL changes.

Fixes #3073

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -752,7 +752,7 @@ def mount_root(pool):
         mnt_options = pool.mnt_options
     if pool.compression is not None:
         if re.search("compress", mnt_options) is None:
-            mnt_options = "{},compress={}".format(mnt_options, pool.compression)
+            mnt_options = f"{mnt_options},compress={pool.compression}"
     if pool.role == "root" and root_pool_mnt != "/":  # boot-to-snap - See pool model
         mnt_options = "{},subvol=/@".format(mnt_options)
     # Prior to a mount by label attempt we call btrfs device scan on all
@@ -2321,10 +2321,11 @@ def set_property(mnt_pt, name, val, mount=True, force=False):
         return run_command(cmd)
 
 
-def get_property(mnt_pt, prop_name=None):
+def get_property(mnt_pt: str, prop_name: str | None=None):
     """
     Convenience wrapper around 'btrfs property get prop_name mnt_pt'.
     :param mnt_pt: Vol(pool)/subvol(share/snap) mount point.
+    :param prop_name: Property name or None for all.
     :return: if called with no prop_name specified then a dict of available
     properties. But if called with a single property then the value and type
     appropriate for that property ie:
@@ -2335,7 +2336,6 @@ def get_property(mnt_pt, prop_name=None):
     N.B. compression property for subvol only, vol/pool uses mount option.
     """
     KNOWN_PROPERTIES = ["ro", "compression", "label"]
-    # TODO: Consider using -t as a form of typesetting on our mnt_pt:
     cmd = [BTRFS, "property", "get", mnt_pt]
     if prop_name is not None:
         cmd.append(prop_name)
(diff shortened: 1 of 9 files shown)


phillxnet:
# Testing

An rpm was build and installed on a Leap 15.6 host from this branch. A prior Rockstor native Pool, with a representative range of; owner:group, permissions, and subvol compression settings was successfully imported: with all prior owner:group, permissions, and subvol compression settings reflected within the Rockstor Web-UI: i.e. Share detail view, default 'Usage' tab for compression, and 'Access control' tab for Owner:Group, and Permissions.

N.B. as per CLI, when importing a pool with unknown users or groups the on-disk UID or GID is displayed instead.

## Unit test run
We have some test extensions to accommodate some, but not all, of the additions here. Notably we have no existing test for our ACL POST. Or our Pool import; where we intend to use instead an openQA approach to confirm end-to-end import capability from a prior real on-disk Pool. However this test setup will be considerably eased once the following capability is added later in the pending testing cycle.

- https://github.com/rockstor/rockstor-core/issues/3072

Test results:
```
...
Ran 292 tests in 59.442s

OK
```

phillxnet:
Moving to merge without review as I'm about to continue the work in this area anyway by returning to:

- https://github.com/rockstor/rockstor-core/issues/3058

[EDIT] It is acknowledged that we may want to abstract the partial repeated code here: but there are large performance hits likely in that abstraction if we are not very carefull. I.e. twice here we update all, and almost all of the Share object and it's DB backing. Once we touch only a few changed elements. In the former two instances we may have to do this a lot of times - ergo in-lining as we have here is subject to less overheads on Python procedural calls. I'm thinking of when we import large subvol counts here. Or when existing imported Pools are processed by this code during an update.

PR #3080, opened 2026-02-27
move share chown chmod operations to the scheduler #3058

### LINKED ISSUE #3058: Move Share chown chmod operations to the scheduler

phillxnet:
Establish a Huey management task to serially execute each
of chown and chmod, in their own tasks, when a Share access
rights change is requested. Stamping the relevant Share with
the management task's unique task id. On management task
completion, invoke via Huey signals a follow-up task to remove
this stamp.

Incidental changes:
- Mostly related string.format to fstring changes.
- Use of system/constants.py for chown & chmod binary locations.

Fixes #3058


phillxnet:
Rebasing to pick-up newer Huey, and squashing to continue development under likely another draft PR. But definitely getting there :) .

phillxnet:
Development now moved to:
- https://github.com/rockstor/rockstor-core/pull/3084

PR #3090, opened 2026-03-18
Reference emerging Rock-on norms re Share (owner:group) etc #3074

### LINKED ISSUE #3074: Reference emerging Rock-on norms re Share (owner:group)
We have emerging and progressively implemented 'norms' for our Rock-ons as developed under:

- https://github.com/rockstor/rockon-registry/issues/465

It is proposed that we update our existing dialog user facing text/guide re Share selection in the following template:

- https://github.com/rockstor/rockstor-core/blob/testing/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/install_choice.jst

Currently reads:

> Shares provide storage to the Rock-on. Read mousever 'i' tooltips for specific information before making a selection. We strongly recommend creating dedicated Share assignments. If a Config or Data Share is assigned to more than one Rock-on, it could cause strange behavior.

to honour/explain Rock-ons Share description 'norms' such as "(10001:10001)" indicating the required (if proceeded by "must be") uid:gid.

N.B. docker has no name mapping from container to the host: only via uid:gid. I.e. user "test" ( uid 1111) in the container is mapped to the host via uid only. If a user exists at all for that user. Similarly for the gid.

phillxnet:
Improve and update our user-facing instructional / guiding text on Rock-on subsystem setup and Rock-on install.

Predominantly focused on now established Rock-on norms re hints and Share owner:group settings. Note that emphasis on dedicated per Rock-on Share has been reduced as we enforce this in code currently. And some self-evident prior instructions have been striped. Helping to avoid a 'wall-of-text'.

Formatting of these texts elements has also been changed from warning to info - more readable and more appropriate. Plus there is now more use of HTML "strong" and italics for emphasis.

Explicit indication of the relevance of Share owner:group has also been added in both the first and last wizard elements, with the latter indicating a need to end the install wizard if there is doubt.

A doc reference was added to the rockon-subsystem OFF landing page. Intended to ease initial Rock-on subsystem configuration.

N.B. there are no functional changes within this patch. Only user facing text changes and their associated formatting. A borderline case of this is the removal of a non-functional (dialog remained) 'Create Share' link.

Fixes #3074

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/rockons.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/rockons.jst
@@ -54,7 +54,7 @@
 </div>
 {{else}}
 <div class="alert alert-info">
-  <strong>Note:</strong> The Rock-on service is not running. Turn it ON to enable this OCI image Docker/Podman based plugin system.
+  <strong>Note:</strong> The Rock-on service is not running. Turn it ON to enable this OCI image plugin system based on Docker.
   <p><i>See our online <a href="https://rockstor.com/docs/interface/overview.html" target="_blank">Rock-ons (Docker Plugins)"</a> guide for full setup instructions.</i></p>
 </div>
 {{/if}}
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/rockons.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/rockons.jst
@@ -53,7 +53,8 @@
   </div>
 </div>
 {{else}}
-<div class="alert alert-warning">
-  <strong>Warning!</strong> Rock-on service is not running. Please turn it on to use features on this screen.
+<div class="alert alert-info">
+  <strong>Note:</strong> The Rock-on service is not running. Turn it ON to enable this OCI image Docker/Podman based plugin system.
+  <p><i>See our online <a href="https://rockstor.com/docs/interface/overview.html" target="_blank">Rock-ons (Docker Plugins)"</a> guide for full setup instructions.</i></p>
 </div>
 {{/if}}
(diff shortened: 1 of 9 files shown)


src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/rockons.jst
```diff
@@ -53,7 +53,8 @@
   </div>
 </div>
 {{else}}
-<div class="alert alert-warning">
-  <strong>Warning!</strong> Rock-on service is not running. Please turn it on to use features on this screen.
+<div class="alert alert-info">
+  <strong>Note:</strong> The Rock-on service is not running. Turn it ON to enable this OCI image Docker/Podman based plugin system.
```
FroggyFlox: Should we keep references to Podman for when we do have support for it?
Maybe something like the following with a more verbose style, yes, but maybe more friendly too:

```suggestion
  <strong>Note:</strong> The Rock-on service is not running. Turn it ON to enable this OCI image plugin system based on Docker.
```

FroggyFlox (commented):
Thanks for improving this all around, @phillxnet!

I just have a minor suggestion for you to consider. I just find it a bit unusual to reference Podman while it's not an option yet. Unless maybe you are trying to mention Docker/Podman there as an example of where OCI images are used? Sorry of that was indeed the intent. 

I'm good either way, honestly 👍.

src/rockstor/storageadmin/static/storageadmin/js/templates/rockons/rockons.jst
phillxnet: @FroggyFlox Thanks for the review.

On your suggestion, I was in two minds about this when I added it, and was just thinking ahead to when we supplant our Docker dependency with a Podman one (initially used in compatibility mode) - and having this user facing text relevant in both situations. But on reflection I agree - when we perform that transition, we can similarly change our Web-UI text references accordingly. And move to mentioning both in the docs to pertain to both existing and newly updated systems post the transition to Podman.

phillxnet:
Last rpmbuild and install test (Leap 15.6 host) with pics before merging.

# Resulting user-facing text:

1_Rockons-before-setup

2_Rockons-Share-setup

3_Rockon-install-Share-selection

4_Rockoin-install-Ports-config

5_Rockon-install-additional-config

6_Rockon-install-verify-before-install

Issue #1624, opened 2017-01-23
[Enhancement] Increase number of btrfs quotas rescan


MFlyer:
While checking over shares usage found we perform quota rescan only while creating snapshots, so if users don't have scheduled snapshots (obviously we hope users have them!) or massively delete them, we can have wrong reporting over shares (quota limits in the future: 2015/* qgroups easily reaching quota limit )

Option A:
add quota rescans when deleting snapshots too

Option B:
make it a supervisored task every x (5-10? mins)

@schakrava & @phillxnet ?
While testing shares usage and deleting some snapshots (no scheduled snapshots) my 2015/* was nicely going up while not expected (when deleting snaps with exclusive sizes, 2015/* being their "father" takes those orphans and seems at least 2 rescan required to come back to real values)

phillxnet:
@MFlyer Another nice find on your part.

My concern with solutions akin to Option B: (periodic re-scans) is that it will break drive power down features. If it turns out we can't avoid this in order to maintain recent info then maybe we could use the existing drive power state to guide these updates; ie akin to what smartmontools does where it can not query a drive if it is in standby but will only 'be nice' for a set number of attempts. There after it will go ahead and wake the drive in the interests of ensuring a recent read of the drives status.

Also I don't see that these options are mutually exclusive and Option A seems like a good idea anyway.

Do we have to account for quota re-scans potentially taking a long time (ie when there are many snapshots).

Apologies if I have missed the point there.

MFlyer:
@phillxnet from btrfs changelog https://btrfs.wiki.kernel.org/index.php/Changelog#btrfs-progs_v4.9.1_.28Jan_2017.29

btrfs-progs v4.9.1 (Jan 2017)

check:

- use correct inode number for lost+found files
- lowmem mode: fix false alert on dropped leaf

size reports: negative numbers might appear in size reports during device deletes (previously in EiB units)
mkfs: print device being trimmed
defrag: v1 ioctl support dropped
quota: print message before starting to wait for rescan
qgroup show: new option to sync before printing the stats
other:

- corrupt-block enhancements
- backtrace and co. cleanups
- doc fixes

MFlyer:
Ref to @schakrava too, hands up for update to latest btrfs (pls remember my tests with 4.9 working fine)

Mirko

phillxnet:
[...]

PR #3099, opened 2026-04-12
"Quotas Disabled" loop - unhandled "rescan is running" #3098

### LINKED ISSUE #3098: "Quotas Disabled" loop - unhandled "rescan is running"
# "Quotas Disabled" loop caused by unhandled "rescan is running" warning in `are_quotas_enabled`

## Summary

On modern kernels (4.12+), `btrfs quota enable` automatically initiates a quota rescan.
During the rescan, `btrfs qgroup show` outputs the following on stderr:

```
WARNING: rescan is running, qgroup data may be incorrect
```

However, `are_quotas_enabled()` in [`fs/btrfs.py`](https://github.com/rockstor/rockstor-core/blob/769c9f9005bdafc008f3abc0bc656116c3ac9e67/src/rockstor/fs/btrfs.py#L1323) only accepts two stderr patterns as "quotas enabled":

- `""` (empty)
- `"WARNING: qgroup data inconsistent, rescan recommended"`

Any other message causes the function to return `False`, incorrectly reporting quotas as disabled.

## Steps to Reproduce

1. Have a pool with quotas enabled on a modern kernel (tested on `6.4.0-150600.23.92-default`)
2. Restart the Rockstor service
3. On startup, `enable_quota()` is called during pool import (`disk.py`), triggering an automatic rescan
4. A concurrent check calls `are_quotas_enabled()` while the rescan is in progress
5. The function sees `"WARNING: rescan is running, qgroup data may be incorrect"` and returns `False`
6. Rockstor tries to re-enable quotas → triggers another rescan → loop

## Observed Behavior

- WUI shows **"Quotas Disabled"** for shares on affected pools
- `btrfs qgroup show` output shows accumulating higher-level qgroups (`2015/1`, `2015/2`, `2015/3`, ...) as Rockstor repeatedly calls `enable_quota()`
- The loop continues indefinitely as long as the service is running

## Expected Behavior

- `are_quotas_enabled()` should return `True` when a rescan is in progress, as quotas **are** enabled — they are simply being rescanned
- No "Quotas Disabled" warning should appear in the WUI during a rescan

## Root Cause

`are_quotas_enabled()` in `fs/btrfs.py` does not handle the `"WARNING: rescan is running, qgroup data may be incorrect"` stderr message, which is emitted by newer versions of `btrfs-progs` during an active rescan.

The comment in `pool.py` already acknowledges that modern kernels auto-start a rescan on `enable`:

```python
# The second above enable_quota() call currently initiates a
# rescan so the following is redundant
```

But `are_quotas_enabled()` was never updated to handle this state.

This can be confirmed by manually triggering a rescan:

```
# btrfs quota rescan /mnt2/<pool>
quota rescan started
# btrfs qgroup show -f --raw /mnt2/<pool> > /dev/null
WARNING: rescan is running, qgroup data may be incorrect
```

## Fix

In `fs/btrfs.py`, add the "rescan is running" warning to the accepted patterns:

```python
# Before
if rc == 0 and (
    e[0] == "" or e[0] == "WARNING: qgroup data inconsistent, rescan recommended"
):
    return True

# After
if rc == 0 and (
    e[0] == ""
    or e[0] == "WARNING: qgroup data inconsistent, rescan recommended"
    or e[0] == "WARNING: rescan is running, qgroup data may be incorrect"
):
    return True
```

## Environment

- Rockstor: 5.5.0-0 (also reproduced on 5.1.0-0)
- Kernel: `6.4.0-150600.23.92-default`
- btrfs-progs: v6.5.1

## Related

- Community forum discussion: https://forum.rockstor.com/t/cant-enable-quotas/10729

mzyy94:
Fixes #3098

## Changes

`fs/btrfs.py`: add `"WARNING: rescan is running, qgroup data may be incorrect"` to the accepted stderr patterns in `are_quotas_enabled()`.

A rescan in progress means quotas **are** enabled — the data is simply being recalculated. Treating this warning as a failure caused Rockstor to re-enable quotas in a loop on startup, accumulating stale `2015/N` qgroups and showing "Quotas Disabled" in the WUI.

## Testing

- Rockstor 5.5.0-0 (also reproduced on 5.1.0-0), kernel `6.4.0-150600.23.92-default`, openSUSE Leap 15.6
- After applying the fix, restarting the Rockstor service no longer causes the re-enable loop
- WUI correctly shows quotas as enabled while a rescan is in progress
- No accumulation of stale `2015/N` qgroups observed after restart

## Related

- Community forum discussion: https://forum.rockstor.com/t/cant-enable-quotas/10729
- #1624

### DIFF
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -1320,7 +1320,9 @@ def are_quotas_enabled(mnt_pt):
     """
     o, e, rc = run_command([BTRFS, "qgroup", "show", "-f", "--raw", mnt_pt])
     if rc == 0 and (
-        e[0] == "" or e[0] == "WARNING: qgroup data inconsistent, rescan recommended"
+        e[0] == ""
+        or e[0] == "WARNING: qgroup data inconsistent, rescan recommended"
+        or e[0] == "WARNING: rescan is running, qgroup data may be incorrect"
     ):
         return True
     # Note on above e[0] clauses:


phillxnet:
@mzyy94 Thanks for seeing to this more recent anomaly re our treatment of btrfs quota status. It has been a rocky-ride all in on that front - as you've seen re our double tap in the past on enabling etc. But bit by bit and this looks like a nice adaptation which keeps things clear.

In time it would be good to introduce the new 'simple' quota system introduced in more recent btrfs. But we are somewhat starved of predominantly human resources for such endeavours.

I've successfully built, installed, and run an rpm from this branch on a Leap 15.6 host and we look to be good.

As is, the test build of what we have in this branch was able to successfully import a test prior Rockstor created Pool in both quotas enabled status (as normal) and the stated fix (reproducer) of importing a quotas disabled Pool. This this is a win :). However as per my https://github.com/rockstor/rockstor-core/issues/3098#issuecomment-4289326218 on your associated issue - we have some improvements re testing - pre package release - on non-trivial pool. All in good time hopefully.

Also note details such as:

"After applying the fix, restarting the Rockstor service no longer causes the re-enable loop"

We have a number of systemd services now, with 4 main ones indicated in our docs here:
code build: https://rockstor.com/docs/contribute/contribute.html#code-build
With the 'top' service as bootstrap - which does the mounts, however, under a variety of circumstances a mount is requested if actions find the underlying pool/share is not mounted. Post some recent changes we have in total 7 main rockstor services as indicated when installing/removing a modern package:

```
Failed to stop rockstor-bootstrap.service: Unit rockstor-bootstrap.service not loaded.
Failed to stop rockstor-scheduling.service: Unit rockstor-scheduling.service not loaded.
Failed to stop rockstor-replication.service: Unit rockstor-replication.service not loaded.
Failed to stop rockstor-collector.service: Unit rockstor-collector.service not loaded.
Failed to stop rockstor.service: Unit rockstor.service not loaded.
Failed to stop rockstor-pre.service: Unit rockstor-pre.service not loaded.
Failed to stop rockstor-build.service: Unit rockstor-build.service not loaded.
```

So all-in thanks for the fix, much appreciated - as always. I'd noticed this failure with non-trivial pools and been struggling to find the time to tend to it - so always good to have more folks eyes our our community effort here.

Thanks also to forum user simon-77 for helping to highlight this issue with some initial reporting.

PR #3096, opened 2026-03-27
Add 'edge' repo tier #3094

phillxnet:
Revises existing update repo tiers of "Testing" & "Stable" by adding "Edge".

Includes:
- Enhancements on all repo tier explanations.
- Remove an existing & redundant 'double-calling' of switch_repo() made by rockstor_pkg_update_check().
- Removed an optional but unused use of switch_repo() by update_run().
- Improved overall robustness and feedback on repo tier switching.
- Fails any repo tier switch if any prior tier is not first removed. Includes user feedback to re-try in this scenario.
- Additional Python type hints on modified code.

Fixes #3094

PR #3103, opened 2026-04-18
Adapt to our fiscal host's legacy tier removal #3101

### LINKED ISSUE #3101: Adapt to our fiscal host's legacy tier removal
[The Rockstor Project](https://opencollective.com/the-rockstor-project) is fiscally hosted by [Open Source Europe](https://opencollective.com/europe), where both entities exist on/within the [Open Collective platform](https://opencollective.com/).

Yesterday (as of this issue's creation) on the 16th April **Open Source Europe** posted the following:
 
- https://opencollective.com/europe/updates/update-to-fundraising-features-on-open-source-europe
 
Indicating that from 24th April (7 days from issue creation)  "... all legacy tiers will be transformed into donation tiers, ...".

Seeking clarification yesterday on the unknown to me  "legacy tiers" reference I received swiftly clarification from the always helpful and supportive Open Source Europe (previously Open Collective Europe) folks to the effect that our existing sustainability endeavour, namely our:

- Stable Updates subscription

Is considered legacy: predominantly as: "... a contribution is tied to a defined benefit, access, or deliverable, ..." which was further explained as potentially creating hosting risks for OSE themselves. We and they obviously do not want this; understandably.

Given the above we must adapt, and rapidly, to a donations only model: which is actually where we began and failed, like the vast majority of Open Source endeavours before and after us. However, as I see it, we have only one path forward here. OSE is our sole fiscal host and we have existing supporters/members to honour. Ergo we move-with-the-times, fingers crossed that they are now more enlightened, and try again at sustainability via a donations only model.

Proposal:
1. We remove the requirement on a "Stable Updates subscription" to access our [Stable Channel](https://rockstor.com/docs/interface/system/update_channels.html#stable-channel) package repo via email arranged activation code.
2. And instead change the "Active Stable updates" [dialog](https://rockstor.com/docs/_images/activate_stable_channel.png) Web-UI element to  show a reminder that our; sustainability, continued development, and distribution is entirely dependent on fiscal contributions via non-obligatory tiers within our Open Collective.
3. After consultation with OSE on the possible ramifications regarding changing either only the existing Stable Updates subscription description - or also it's title. As we do not want to affect any existing subscriptions. We also need their sign-off on our adaptation to their recent changes.

Noting here also that every stable repo  rpm package, during our OCE/OSE time, was first published in the testing repo; at the end of the preceding testing phase that lead to the Stable repo upload. Thus our software package offering was never exclusive, only the curation associated with the Stable updates repo. However this is a 'benefit' linked to an OSE subscription and thus now out-of-bounds!

---

N.B. the current clarification from OSE standing at "... a contribution is tied to a defined benefit, access, or deliverable, ..."  also clearly precludes our long standing plan to offer [Paid support](https://rockstor.com/paid_support.html) via an as yet non-existent "**Incident-based Support subscription**" within our Open Collective: folks will, having paid for a support plan, understandably expect some form of 'deliverable' i.e. support! Or to in some other way hope at least that a 'benefit' might be derived from having contributed towards such support arrangements financially! As such, as I interpret this, any paid support for Rockstor or any open source project hosted by OSE, would, somewhat ironically, have to entail an independent, presumably for-profit entity! We are where we find ourselves so lets be pragmatic on this one and concentrate on the code and its development, distribution, and ongoing maintenance. And hope that enough folks will indulge in our to-be-redefined contributory offerings via our Open Collective.

phillxnet:
Predominantly this removes the pre-requisite "Stable Updates subscription" re Stable Updates activation: as far as rockstor-core is concerned.

As this stands we have a full Web-UI adaptation re user-facing text, but only the barest minimum re code changes. Leaning on the nature of http auth access; where if it is not required, it is generally not provided, or not a problem if it is. This minimalist approach is motivated by the tight externally defined deadline, and the desire to get this into production, and hopefully the Stable release also, as soon as possible. We can, at a later date, remove all auth associated code elements concered with the Stable repo. But for now they do no harm in-place.

N.B. Requires auth removal from the Stable repo prior to proper function.

Minor existing typo also addressed.

Fixes #3101

### DIFF
--- a/src/rockstor/storageadmin/static/storageadmin/js/templates/update/version_info.jst
+++ b/src/rockstor/storageadmin/static/storageadmin/js/templates/update/version_info.jst
@@ -51,7 +51,7 @@
         Rockstor package
         <a href="https://rockstor.com/docs/update-channels/update_channels.html" target="_blank"> update channels. </a>
         Activate appropriately for your needs and/or intentions.
-        <br>- <strong>Stable updates</strong>: Intended for production use (<i>non-profit</i> subscription).
+        <br>- <strong>Stable updates</strong>: Intended for production use (free - <i>donation requested</i>).
         <br>- <strong>Testing updates</strong>: Field-testing towards future Stable releases (free).
         <br>- <strong>Edge updates</strong>: Untested releases - developers only (free).
         <br>See our <a href="https://forum.rockstor.com/" target="_blank">friendly forum</a> for advise.
@@ -62,6 +62,10 @@
         are a <i>non-profit</i> community endeavour fiscally hosted by
         <a href="https://opensourceeurope.org/" target="_blank"> Open Source Europe (OSE).</a>
       </strong>
+      <br>
+      <i>See our
+        <a href="https://opencollective.com/the-rockstor-project" target="_blank">OPEN COLLECTIVE</a>
+        for donation options.</i>
     </div>
     {{/no_sub_active}}
 
@@ -226,6 +230,10 @@
         are activated - install is eligible for
         <a href="https://rockstor.com/paid_support.html" target="_blank"><i>Paid Support.</i></a>
       </strong>
+      <br><br>
+      <i>See our
+        <a href="https://opencollective.com/the-rockstor-project" target="_blank">OPEN COLLECTIVE</a>
+        for donation options.</i>
     <hr>
     The following development orientated channels are available:<br>
     <dl>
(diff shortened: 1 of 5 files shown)


phillxnet:
# Testing re modified user dialogs:

## Subsequent Dashboard visits popup

When no update channel has been selected:

Subsequent-dashboard-pop-up

## More info / Initial channel selection

update_channel_options

## Stable info popup

activate_stable_channel

### Stable activated

stable_activated

---

See PR:  https://github.com/rockstor/rockstor-core/pull/3096 for our as-yet unpackage Edge repo addition for the nature of these Web-UI elements, as rendered, prior to the changes proposed here.

phillxnet:
# Testing re functionality

This one is tricky as we must first remove auth from our stable repo - but for that we must first have all our ducks in a row. So as per the approach taken in https://github.com/rockstor/rockstor-core/issues/3101#issuecomment-4269704452 to evaluate possible side-effects for existing installations (subscribed to Stable) post auth removal, the following was done for Web-UI end-to-end testing.

A clean system (no .initrock), with all services stopped had the stable url defined within settings.py modified to be testing instead: stable will be as testing is now post auth removal. The rockstor-bootstrap services was then started to enable DB population and the system was subscribed to stable (without activation code). The result a Rockstor-Stable (but using testing repo) but with all the associated auth elements attached - functioned normally: i.e. all additional auth associated elements were instantiated but without effect/side-effect. We would need more field testing to evaluate properly, but this looks to be good to go re edge/testing publishing but ideally we first remove the auth on the stable repo to avoid folks running this version and then having insufficient in-Web-UI info to subscribe to Stable.

N.B. during the above testing approach it was found that we do not update our db reflection of the urls defined within settings.py unless no existing subscription DB elements are in play - we can address that anomaly at a later date in a dedicated issue/pr set. Hence invoking a DB wipe via the .initrock removal and services bounce.

src/rockstor/storageadmin/static/storageadmin/js/templates/update/version_info.jst
```diff
@@ -271,26 +279,52 @@
...
-            <a href="https://opencollective.com/the-rockstor-project/contribute" target="_blank">Open Collective.</a>
-          </li>
-          <li>Enter / Edit / Check your <span style="color:green">Current Appliance ID</span> in your
-            <a href="https://appman.rockstor.com/" target="_blank">Appliance ID manager</a> <strong>Appman</strong>.
-          </li>
-          <li>Enter your Activation code below, emailed by <strong>Appman</strong> immediately after step 2.</li>
-        </ol>
-        Thank you for helping to support Rockstor's non-profit development.
+        <br>
+        Help to keep our digital sovereignty endeavour from falling to a for-profit for its maintenance,
```
FroggyFlox, line 300: Unsure about this one, but would the following be more proper (or just sound better)?
```suggestion
        Help keeping our digital sovereignty endeavour from falling to a for-profit for its maintenance,
```

src/rockstor/storageadmin/static/storageadmin/js/templates/update/version_info.jst
```diff
@@ -271,26 +279,52 @@
...
+        continued development, distribution, etc.
+        Our code has always been
+        <a href="https://rockstor.com/docs/interface/system/update_channels.html#licensed-fsf-free-libre-osi-approved" target="_blank">FSF Free/Libre & OSI approved</a>,
+        but a project is more than its code!
+        Real folks from around the world contribute their time.
+        And distribution, infrastructure, and online services / servers are not free.
+        Backend support systems, webpages, docs , guides, how-tos etc all also require folks time, attention, and resources.
```
FroggyFlox:
```suggestion
        Backend support systems, webpages, docs, guides, how-tos, etc... all also require folks time, attention, and resources.
```

src/rockstor/system/pkg_mgmt.py
```diff
@@ -442,8 +442,14 @@ def switch_repo(subscription: UpdateSubscription, enable_repo: bool = True):
 
 
 def repo_status(subscription: UpdateSubscription):
-    if subscription.password is None:
+    # We need not check authentication credentials when they are not requried.
```
FroggyFlox:
```suggestion
    # We need not check authentication credentials when they are not required.
```

FroggyFlox (approved):
@phillxnet,

Thank you for writing all of this down, I agree with all of it and I do like the new acknowledgement requirement.

I only see a typo or two and maybe a couple suggestions for improvement; all up to you.

I could verify:
- builds on Leap 15.6 and Slowroll, both x86_64
- behaves as expected on Slowroll, x86_64

src/rockstor/storageadmin/static/storageadmin/js/templates/update/version_info.jst
phillxnet: @FroggyFlox Thanks for looking over all this. I have now presented a counterpart doc entry so some of this may well have to await the next round of changes - given the deadline here.

Re this specific change - I think we are better off with the original - we are starting a-fresh on this sentence and just saying to folks 'Help to keep us from ... falling'. So 'Help keeping us from failure' doesn't read as well I think.

phillxnet:
@FroggyFlox Thanks for the review, much appreciated. And for doing a multi target build test. Nice.

I've applied all but one of your suggestions. Thanks. I'll go ahead and merge this ready to line everything up for our next testing release, to be synced with the requisite repo auth changes; also lined up.

Only the pending related doc PR to review and merge now then.

PR #3108, opened 2026-04-23
build.sh errors - standalone use #3107

### LINKED ISSUE #3107: build.sh errors - standalone use
I am having issues with building the vm using the web instructions, though ...
after fulfilling all the pre-requisites and then executing `build.sh` the following error message appears at the end:

```
gpg: A key for "rockstor@localhost" already exists
Sourcing /opt/rockstor.env
build.sh: line 92: source: .env: file not found
```

I executed the build file after having changed into the `/opt/rockstor` directory (as per instructions).
The `.env` file is present in the /opt/rockstor/ directory ...

I know there was some issue with the gpg key stuff, but it did not seem completely related to what's happening here ...

If I put an explicit path on the .env file (i.e. `source /opt/rockstor/.env`, it went beyond that but more error messages were coming subsequently (don't worry about the SECRET_KEY below, since this is a test VM, it will be destroyed shortly anyway):

```
gpg: A key for "rockstor@localhost" already exists
Sourcing /opt/rockstor.env
Password store initialized for rockstor@localhost
python-keyring/rockstor/CLIENT_SECRET: reencrypting to 79D16E7385EAFEB5
gpg: public key decryption failed: No secret key
gpg: decryption failed: No secret key
The generated password for python-keyring/rockstor/SECRET_KEY is:
wNB6kbtgyrNT4tfaiVW8IzqgBYbE5QJloDTmazxePX8q87mRqxTGmofnWMsMWv49h3kvMRYRA3ZknH65Pa1zVbkmw3DUaLC6K7ll
gpg: public key decryption failed: No secret key
gpg: decryption failed: No secret key
  File "/opt/rockstor/.venv/bin/django-admin", line 8, in <module>
    sys.exit(execute_from_command_line())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 382, in execute
    settings.INSTALLED_APPS
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/django/conf/__init__.py", line 81, in __getattr__
    self._setup(name)
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/django/conf/__init__.py", line 68, in _setup
    self._wrapped = Settings(settings_module)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/django/conf/__init__.py", line 166, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/opt/rockstor/src/rockstor/settings.py", line 135, in <module>
    CLIENT_SECRET = keyring.get_password("rockstor", "CLIENT_SECRET")
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/keyring/core.py", line 65, in get_password
    return get_keyring().get_password(service_name, username)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/keyring_pass/__init__.py", line 183, in get_password
    ret = command(
          ^^^^^^^^
  File "/opt/rockstor/.venv/lib/python3.11/site-packages/keyring_pass/__init__.py", line 26, in command
    return subprocess.check_output(cmd, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['pass', 'show', 'python-keyring/rockstor/CLIENT_SECRET']' returned non-zero exit status 2.
```

On the last error message, I executed the command manually after the failure:
`pass show python-keyring/rockstor/CLIENT_SECRET`
which returned this:
```
gpg: public key decryption failed: No secret key
gpg: decryption failed: No secret key
```

Looking at `gpg --list-keys` and `gpg --list-secret-keys` in both cases a valid (not expired) entry was presented. I subsequently deleted both the secret and non-secret key (and reloaded the `keyboxd`) to run the build program once more. It did not provide the message that a key already existed but still ran into the subsequent errors.

So, not sure what I am missing, or what might have changed from the last time (a while ago) I built a VM ... this is on a Leap 15.6 VM where I remove Rockstor, re-installed it using zypper, and removed it once more ...

_Originally posted by @Hooverdan96 in https://github.com/rockstor/rockstor-core/issues/3104#issuecomment-4285296610_

phillxnet:
Fix a path fault in an `echo` notification, and the associated `source` of our .env file that failed with consequences when not run under systemd rockstor-build.servcie via `EnvironmentFile=/opt/rockstor/.env` where the required contents were already established.

Fixes #3107 

---

There are likely associated developer doc updates that we need to tend to also.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -88,10 +88,11 @@ fi
 /usr/bin/gpg --quick-generate-key --batch --passphrase '' rockstor@localhost || true
 # Init 'pass' in .env defined PASSWORD_STORE_DIR using above GPG key, and generate Django SECRET_KEY
 set -o allexport
-echo "Sourcing $(pwd).env"
-source .env  # also read by rockstor-build.service
+echo "Sourcing $(pwd)/.env"
+source $(pwd)/.env  # also read by rockstor-build.service via "EnvironmentFile=/opt/rockstor/.env"
 set +o allexport
-/usr/bin/pass init rockstor@localhost
+# Ensure password-store is initialized:
+/usr/bin/pass init rockstor@localhost || true
 /usr/bin/pass generate --no-symbols --force python-keyring/rockstor/SECRET_KEY 100
 
 # Collect all static files in the STATIC_ROOT subdirectory. See settings.py.
--- a/build.sh
+++ b/build.sh
@@ -108,6 +108,6 @@ echo "If installing from source, from scratch, for development; i.e. NOT via RPM
 echo "Note GnuPG & password-store ExecStartPre steps in /opt/rockstor/conf/rockstor-pre.service"
 echo "1. Run 'systemctl start postgresql'."
 echo "2. Run 'cd /opt/rockstor'."
-echo "3. Run './build.sh'."
+echo "3. Run 'sh ./build.sh'."
 echo "4. Run 'poetry run initrock' as root (equivalent to rockstor-pre.service ExecStart)."
 echo "5. Run 'systemctl enable --now rockstor-bootstrap'."


phillxnet:
@Hooverdan96 Could to see if these changes, along with the procedure indicated (rather verbosely) in the associated issue https://github.com/rockstor/rockstor-core/issues/3107#issuecomment-4305923667 works to address the issue you were seeing.

As indicated we likely also need to refresh our developer docs also. But the proposed changes here do look to be required to ease developer experience.

phillxnet:
The included  "|| true" change here is likely not actually required - but we can leave them in place regardless. It may be there is a change in behaviour going forward regarding existing setup and not returning true.

Hooverdan96:
Testing the prep procedure you described in the associated issue, first with the "old" `build.sh` program to see whether the error is still thrown. As expected, it is. Using your updated version completes successfully, it does show in the two error messages:

after displaying the password
```
Error: python-keyring/rockstor/SECRET_KEY_FALLBACK is not in the password store.
No SECRET_KEY_FALLBACK - rotated on reboot / rockstor services restart.
Error: python-keyring/rockstor/CLIENT_SECRET is not in the password store.
```
continues on with the copying steps and subsequently completes the process with the expected information on how to proceed:

```
518 static files copied to '/opt/rockstor/static', 520 post-processed.

ROCKSTOR BUILD SCRIPT COMPLETED

If installing from source, from scratch, for development; i.e. NOT via RPM:
Note GnuPG & password-store ExecStartPre steps in /opt/rockstor/conf/rockstor-pre.service
1. Run 'systemctl start postgresql'.
2. Run 'cd /opt/rockstor'.
3. Run './build.sh'.
4. Run 'poetry run initrock' as root (equivalent to rockstor-pre.service ExecStart).
5. Run 'systemctl enable --now rockstor-bootstrap'.
```

Step 1 and 2 execute as expected
Step 3 throws the error `-bash: ./build.sh: Permission denied`
which makes sense, since the file attribute is not set to `executable`. If this step is still necessary, then maybe the instruction step just needs to be changed to `3. Run 'sh ./build.sh'`
Step 4 completes as expected
Step 5 also completes as expected.

Rockstor initial screen can now be loaded:
image

and proceeding with activating the edge channel:
image

A reboot after the initial correctly starts Rockstor, i.e. no `rockstor-pre` service failure. So, all looks in order at this time ...



> I think as part of the initial prep instructions you've listed in the issue comment, for the updated community contributions, we might want to also include the (optional) removal of a prior created admin user (in case the previous install - rpm or dev build) with `userdel <prior admin user name>` or similar, so the same one can be recreated during the initial install - likely more important if a config backup is used to quickly restore any additional setup that involved having the admin user owning/running it.

build.sh
phillxnet: The suggested patch looks like it deletes a line and replaces it with  the echo command suggestion. Do you fancy having another go. Totally agree with the suggestion indicentall.

phillxnet: Or maybe we have a bug in the presentation here as there is no clear line and it does state line 111 is to be wiped!!

phillxnet:
@Hooverdan96 Thanks for the review. Much appreciated. Now we just need to work out what's going on with the funny looking change suggestion.

build.sh
Hooverdan96, line 112:
```suggestion
echo "Note GnuPG & password-store ExecStartPre steps in /opt/rockstor/conf/rockstor-pre.service"
echo "1. Run 'systemctl start postgresql'."
echo "2. Run 'cd /opt/rockstor'."
echo "3. Run 'sh ./build.sh'."
echo "4. Run 'poetry run initrock' as root (equivalent to rockstor-pre.service ExecStart)."
```
explicitly call bash for the `build.sh` batch file

Hooverdan96: still not quite right, but might be better. May be it's not doing it right, because it's not within any of your other changes?
Otherwise, just do it from your side, I have enough attribution in the conversation 😄

phillxnet: @Hooverdan96 Re:

> because it's not within any of your other changes?

I think you are right.

phillxnet:
@Hooverdan96 I've added your suggestion, re-based on latest testing, squashed, and re-presented under:
- https://github.com/rockstor/rockstor-core/pull/3110

phillxnet:
@Hooverdan96 note that our instructions as printed appear in zypper logs, they are intended as the whole instructions for source build - not a line-by-line, but nearly: i.e. they allude the the systemd key setup stuff for example. Source build is a dev endeavour, but agreed we have to keep it clean: but we need not dumb it down too much as it is assumed folks may be in various stages of buid i.e. re-build need not reset everything.

This is in the context that if one has already just run build.sh, there is no need to re-run it as per the from-scratch instructions presented. We need to work on our docs in this regard I appreciate.

PR #3114, opened 2026-04-24
Fix remaining utc treatment (#3111)

### LINKED ISSUE #3111: Additional missed requirement re utc usage change
In reference to #3102 

There is one more instance of the utc usage that will be removed at some point in the future that was not immediately apparent when investigating the above referenced issue.

According to https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

`utcfromtimestamp` will be removed in the future as well (deprecated in 3.12, removal timing TBD),for the same reason as the other utc related datetime functions were (naive vs. timezone aware behavior).

Fortunately, there seems to be only one occurrence in the rockstor-core code base:

https://github.com/rockstor/rockstor-core/blob/cd749898b7175ca0e716fe80ea46463d64ae0ed4/src/rockstor/smart_manager/agents/nfsd_calls.py#L29-L31

The recommendation is to replace this line with something like this:
from:
`datetime.datetime.utcfromtimestamp(float(ts)).replace(tzinfo=timezone.utc)`
to:
`datetime.datetime.fromtimestamp(float(ts), tz=datetime.UTC)`

and adjust the import from:
https://github.com/rockstor/rockstor-core/blob/cd749898b7175ca0e716fe80ea46463d64ae0ed4/src/rockstor/smart_manager/agents/nfsd_calls.py#L18

to:
```
from datetime import datetime, UTC
```

Admittedly, low-ish priority, but wanted to record this regardless.

Hooverdan96:
Fixes #3111.

Coding updated as per described in the above issue to ensure future compatibility with higher versions of python (namely 3.13 and up).

Executing all tests:

```
Ran 292 tests in 73.206s

OK
```
Also realized that I think there are no nfs specific tests currently in the library.

In any case, created an nfs export, and connected from a Ubuntu instance to it. Ability to see and open files worked successfully. Again, not entirely sure whether this particular function was called during any of it. However, no syntax errors, etc. have been observed.


FroggyFlox:
Thanks a lot, @Hooverdan96 !

I think the target branch was inadvertently set to `master`, unfortunately. Would you be able to correct it to `testing`, please?

Hooverdan96:
Of course, this was a bad choice to put the PR against the Master branch. I will close and create a new one against Testing.

Sorry about that.

Hooverdan96:
superseded by new PR #3115 against `testing` branch.

Issue #2963, opened 2025-02-27
Revise/review SystemTap NFSD monitoring


phillxnet:
We have, somewhat latent, advanced NFSD kernel module monitoring - no longer surfaced within the Web-UI; based on SystemTap kernel module technologies:

- https://documentation.suse.com/sles/15-SP6/html/SLES-all/cha-tuning-systemtap.html

See: https://github.com/rockstor/rockstor-core/tree/master/src/rockstor/smart_manager/taplib/nfsd

for our existing SystemTap scripts, in a language akin to AWK and C.

It is proposed that we consider reviving this non-trivial effort to re-instate the advanced NFSD monitoring that it can provide; however there are now better options to retrieve the same data from this module, i.e. [eBPF](https://ebpf.io/what-is-ebpf/).

See: https://ebpf.io/what-is-ebpf/#development-toolchains

- "BCC is a framework that enables users to write python programs with eBPF programs embedded inside them."
- bpftrace: "... makes use of BCC for interacting with the Linux eBPF subsystem ... The bpftrace language is inspired by awk, C, and predecessor 
  tracers such as DTrace and SystemTap."

## How are eBPF programs written
From the same page: https://ebpf.io/what-is-ebpf/#how-are-ebpf-programs-written

- BCC: https://ebpf.io/applications/#bcc Referencing in-turn GitHub: https://github.com/iovisor/bcc
- bpftrace: https://ebpf.io/applications/#bpftrace Referencing in-turn GitHub: https://github.com/bpftrace/bpftrace Project Website: https://bpftrace.org/

From the above notes, bpftrace looks like a good fit: based on our prior and presumed working SystemTap scripts, and the fact that bpftrace appears to be a superset/wrapper of BCC.

---

Linking to last modifications associated with this code area:
- update smart probe kernel modules in 2.0 #348
- Fixes #348 #355

phillxnet:
The associated API endpoints within the host Django app of smart_manager are as follows:
```
 ^api/sm/sprobes$ [name='probe-view']
^api/sm/sprobes/ ^metadata$ [name='probe-view']
^api/sm/sprobes/ ^metadata/(?P<pid>[0-9]+)$ [name='probe-view']
^api/sm/sprobes/ ^diskstat/$ [name='diskstat-view']
^api/sm/sprobes/ ^meminfo/$ [name='meminfo-view']
^api/sm/sprobes/ ^netstat/$ [name='netstat-view']
[...]

PR #3115, opened 2026-04-24
Fix remaining utc treatment #3111

### LINKED ISSUE #3111: Additional missed requirement re utc usage change

Hooverdan96:
Fixes #3111 .

Coding updated as per described in the above issue to ensure future compatibility with higher versions of python (namely 3.13 and up).

Executing all tests:

Ran 292 tests in 73.206s

OK
Also realized that I think there are no nfs specific tests currently in the library.

In any case, created an nfs export, and connected from a Ubuntu instance to it. Ability to see and open files worked successfully. Again, not entirely sure whether this particular function was called during any of it. However, no syntax errors, etc. have been observed.

### DIFF
--- a/src/rockstor/smart_manager/agents/nfsd_calls.py
+++ b/src/rockstor/smart_manager/agents/nfsd_calls.py
@@ -15,7 +15,7 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
-from datetime import datetime, timezone
+from datetime import datetime, UTC
 from smart_manager.models import (
     NFSDCallDistribution,
     NFSDClientDistribution,
@@ -27,7 +27,7 @@ from smart_manager.models import (
 
 
 def get_datetime(ts):
-    return datetime.datetime.utcfromtimestamp(float(ts)).replace(tzinfo=timezone.utc)
+    return datetime.datetime.fromtimestamp(float(ts), tz=UTC)
 
 
 def process_nfsd_calls(output, rid, l):


phillxnet:
@Hooverdan96 Thanks for these changes, this is all a help towards not tripping ourselves up as we do Python updates for sure. Much appreciated. I've done a test rpm build and it completed as expected.

Re:
> not entirely sure whether this particular function was called during any of it.

This is dead code, as-in not called any-longer, and I think some of it is still in Python 2.7!. It relates to defining a Python agent/interface that talks to a currently unmaintained custom kernel module using SystemTap for advanced NFSD  monitor that we use to 'carry' and surface within the Web-UI. Take a look at the following issue for some context:

- https://github.com/rockstor/rockstor-core/issues/2963

I'd really like to restore this functionality, but transition it to something way more modern such as eBPF with bpftrace. That would likely mean we could use much of this existing work and restore what we once had re advanced NFS live statistics within the Web-UI: retrieved via eBPF in the kernel ultimately.

phillxnet:
@Hooverdan96 I've edited the PR title as it may trip-up our Web-UI Changelog parsing or the like. We need to make all that stuff more robust in time.

PR #3124, opened 2026-05-08
Drop support for NIS #2947

### LINKED ISSUE #2947: Possibly drop support for NIS?
Reading up more on NIS, NIS+ vs ldap and other domain authentication tools, it seems to me that since NIS is both fairly insecure and likely not in wide use anymore, it could make sense to retire the related code and functionality? On the same token, if nothing is broken and users are not complaining about it not working, it could stick around until OpenSUSE or the package maintainer drops it.

Thoughts?

phillxnet:
Remove now dated, & long unmaintained upstream, NIS service. This service is also no longer available in many linux distros, including Leap 16.0.

Fixes #2947


phillxnet:
# Testing

An rpm was build, and subsequent installed successfully on a Leap 15.6 host. All unite tests were then run successfully on the resulting install.

The services page worked as expected and indicated not NIS listing.
N.B. we have as yet no automated update or config restore mechanisms in place. So we should invite testing feedback on this front after the next rpm containing these changes is published to the testing channel repo.

phillxnet:
An Edge repo rpm, without changelog updates, is available now as version 5.5.2-3124 with the proposed changes in this pull request.

phillxnet:
Note: We still have fixtures (test DB contents) that contain service info related to NIS, however this is not considered significant and can be addressed once we have need to re-create the relevant fixtures in the future.

PR #3139, opened 2026-06-10
gpg: Note: database_open ... waiting for lock #3117

### LINKED ISSUE #3117: gpg: Note: database_open ... waiting for lock
~~In development only scenarios (to date) have lead to~~ A locked gpg database during build.sh's execution [EDIT: and in other now observed scenarios]:

```
gpg: Note: database_open 134217901 waiting for lock (held by 18139) ...
gpg: Note: database_open 134217901 waiting for lock (held by 18139) ...
gpg: Note: database_open 134217901 waiting for lock (held by 18139) ...
```
with consequent timeouts on password-store management.

The indicated PID in the above instance was:

>keyboxd --homedir /root/.gnupg --daemon

## Possible work-around: 
```
gpgconf --kill gpg-agent
gpgconf --kill keyboxd
rm -f ~/.gnupg/public-keys.d/pubring.db.lock
gpgconf --reload gpg-agent
```

Issue opened to help collaboratively explore what is cause this GPG db lock to be held detrimentally.
It may be that we need a preventative measure within build.sh to avoid the possibility that in-production updates are affected similarly.

phillxnet:
After successfully killing `gpg-agent --homedir /var/tmp/zypp`, ensure the gnupg pubring.db.lock file is removed.

Adds todo for future move to Django tasks, to avoid forced zypper time-outs and consequent clean-up.

Closes #3117 

---

We have, as yet, no reliable reproducer for the referenced issue. As such this 'fix' has undetermined confidence.

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -564,6 +564,8 @@ def zypper_killed_cleanup():
     """
     # See: https://www.gnupg.org/documentation/manuals/gnupg/Agent-Signals.html
     # pkill (procps package) defaults to SIGTERM
+    # TODO: Now we are on Django 6.0 - remove all zypper forced time-outs and establish
+    #  Django task based management of, for example, updates available.
     out, err, rc = run_command(
         ["pkill -f 'gpg-agent --homedir /var/tmp/zypp'"], shell=True, throw=False
     )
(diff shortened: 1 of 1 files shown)


phillxnet:
# Testing

All OS & arch target RPMs were successfully built and installed on their respective hosts - with all unit tests then passing.

Merging ready for field testing of this approach - at least until we move to Django tasks now we are using Django 6.0.

Issue #2773, opened 2023-12-26
Shellinabox service is the only one with custom config instantiation


FroggyFlox:
For all services listed in *System* > *Services*, we use a default config of `null` at the exception of the Shellinabox service, which has a pre-set default config. This is defined at:
https://github.com/rockstor/rockstor-core/blob/b870fa2f0264a8696440cccbc2526c50617ef773/src/rockstor/scripts/prep_db.py#L45-L51

As mentioned in that snippet, we should consider moving this default config to `null`, in line with all other services. This would in  turn greatly help with our handling of system services as we currently find ourselves needing to set an exception for the shellinabox service every time we work on services configurations.

Incidentally, since @phillxnet's rework of our database instantiation mechanisms (#2729/#2733), Bug 2 detailed in https://github.com/rockstor/rockstor-core/issues/2723#issue-1959426926 now applies to the Shellinabox service, which results in an empty config for that service as well after initial database setup.

```
systemctl stop rockstor rockstor-bootstrap
rm .initrock # so that existing data is wiped upon first run of initrock
poetry run initrock
```
The Shellinabox service now has no default config defined:
```
buildvm155:/opt/rockstor # poetry run psql -U rocky -d smartdb -c "SELECT * FROM smart_manager_service WHERE name = 'shellinaboxd';"
Password for user rocky: 
 id |     name     |  display_name  | config 
----+--------------+----------------+--------
 15 | shellinaboxd | Shell In A Box | 
(1 row)
```

In such a situation, there does **not** appear to be any negative consequence as the Shellinabox service:
- can still be configured as normal
- can still be started as normal

This thus seems to support the fact that we can simply remove our mechanisms that enforce a default config for that service but I would still need to verify that further as I'm not sure why we have such mechanisms in the first place.

FroggyFlox:
> This thus seems to support the fact that we can simply remove our mechanisms that enforce a default config for that service but I would still need to verify that further as I'm not sure why we have such mechanisms in the first place.

This seems to come from:
https://github.com/rockstor/rockstor-core/pull/1388#discussion_r70688733

Indeed, with a `null` config, clicking on "System" > "System shell" leads to an endless spinning icon with dev tools console showing:
```
Uncaught TypeError: right-hand side of 'in' should be an object, got null
```

FroggyFlox:
To go back to @schakrava's comment linked above (https://github.com/rockstor/rockstor-core/pull/1388#discussion_r70688733):

[...]

Issue #2522, opened 2023-03-25
4.5.8-0 System Shell not working in mobile browsers


kensee:
Tested with all browsers like Internet, Chrome, firefox, Safari.

No enter key in keyboard to enter for userid/password

For firefox, on top of the above issue, multiple characters appear when trying to login.

Same issue already exists since 4.1

phillxnet:
@kensee Thanks for the report. We use shellinabox for this embedded Web-UI terminal.

As a  workaround you should be able to use any ssh client app available for your phone OS and use that for terminal access. Currently our root user is permitted ssh access on our Leap based installers.

kensee:
@phillxnet Thanks for the reply. It works fine with desktop browsers. Will there be a fix for it in mobile browsers? With that I do not need to open ssh port to the Internet.

phillxnet:
@kensee Hello again.
Re:
> Will there be a fix for it in mobile browsers?

Not from our end as we basically use it as it's delivered form upstream openSUSE repos on OBS. We do host it on our own OBS but it's basically unchanged from how it is delivered upstream but we needed at the time an ARM target that was missing so we hosted the exact same code and added a ARM target to allow us to install it in our aarch64 installers.

You could report the issue upstream maybe.
Our build of shellinabox is build on OBS here:
https://build.opensuse.org/package/show/home:rockstor/shellinabox
but is just a copy (unchantged at the time)  from the upstream within OBS here:
https://build.opensuse.org/package/show/shells/shellinabox
Where the last change noted there was one we submitted to the project via:
https://build.opensuse.org/request/show/781298

So that would rather indicate that it's not actively maintained. However our suggested change detailed in the above was received by the OBS packages for shellinabox pretty quick and accepted as soon as I got my changelog ducks-in-a-row.

As you see our prior change was only to the included systemd file as we required these changes in our move from our old/prior upstream of CentOS to our new "Built on openSUSE" setup.

I wonder if a work-around such as a specialist keyboard overlay type app would help your situation.

I'll leave this issue open for a bit but to see if we get any interest, but you might like to try asking on our forum:
https://forum.rockstor.com/
to see i others have found other workarounds or have some knowledge of how this can be fixed upstream. But really I think it's basically working as the original authors intended and that you just need a keyboard app that work son your phone OS and with this terminal in a webpage shellinabox that we basically use unaltered. We don't embed it in our app, we just start and stop it via it's systemd apps and point it's output to a page of ours.

[...]

PR #3153, opened 2026-07-05
Remove Shellinabox #3151

### LINKED ISSUE #3151: Remove Shellinabox
Currently unavailable in Leap 16.0 due to OBS build issues associated with an unmaintained and also removed in Leap 16.0 script dependency that has now been updated; in part as a result of @FroggyFlox's inquiries to openSUSE folks to try and resolve this cascade dependency & maintenance failure.

Proposal here is to remove our dependency on Shellinabox:
- OBS Rockstor (home_rockstor repo): https://build.opensuse.org/project/show/home:rockstor
- Our package (carried to provide aarch64 builds): https://build.opensuse.org/package/show/home:rockstor/shellinabox
- OBS upstream: https://build.opensuse.org/package/show/shells/shellinabox

In the above @FroggyFlox & myself are the only contributors to this OBS project for the last 7 years, and then only in fixing packaging/systemd issues. And we originally established this OBS setup, with associated rockstor_home repo, to provide more built targets so that we could support for example aarch64 OS variant.

An ultimate upstream exists but is similarly unmaintained bar a couple of fixes in 2025 (v2.21) with the release prior dating back to the end of 2016.

The particular nature of this package's function deems it extra sensitive re active maintenance both in upstream and OBS maintenance - which in the latter case we have assisted with for this duration within our limited means.

References:

- Workaround for our current Leap 16.0 target was to remove this rpm dependency: breaking its function within the Web-UI, but allowing us to progress in all other respects with supporting this new OS version.
-- https://github.com/rockstor/rockstor-rpmbuild/pull/98
-- "... shellinabox remarked out due to build issues in OBS."

---

Ideally we would replace this package with something providing similar capability; however we are in late RC status in our current testing phase and CLI login facility is easily established with client software (SSH clients).

phillxnet:
Remove all Shellinabox associated mechanisms and configs. Insufficient upstream maintenance.
- etc/nginx/nginx.conf re /shell/ http://127.0.0.1:4200 proxy.
- settings.py re user:group port and certs directory.
- settings.py normalise on sellinaboxd service file name: OBS package used sellinabox.service.
- scripts/prep_db.py Remove sole anomalous non-null default config belonging to the Shellinabox service.
- smart_manager/views removed: enabling and disabling view.
- system/shell.py Remove helper procedures.
- system supported_services whitelist entry removed.
- smart_manager/urls Remove API end-points.
- various storageadmin/static js & template dedicated references removed.
- Update system/tests-test_services.py.

Unrelated changes:
- Remove stray reference to previously removed NIS service:  views/services.js.

Fixes #3151 

---

Outstanding Shellinbox issues also closed by its removal in this branch:

- https://github.com/rockstor/rockstor-core/issues/2773
- https://github.com/rockstor/rockstor-core/issues/2522


phillxnet:
An RPM was build and installed on a Slowroll x86_64 host and successfully setup via the Web-UI.

- Services page worked as expected and we have no "Shell In A Box" listing on the main page or it's side-nav-bar, and within the Web-UI sub-header:

## Before removal example

preious-to-shellinabox-removal

## After removal (this branch on SR)

header-no-shellinabox

PR #3156, opened 2026-07-06
Replication: missing slash in snapshot path #3147

### LINKED ISSUE #3147: Replication: missing slash in snapshot path
Thanks to forum user [@riceru](https://forum.rockstor.com/u/riceru) for surfacing the issue during replication processing, where during one of the definitions of the snapshot path which is required for the replication processing, a missing `/` was found in the path.

Forum [link](https://forum.rockstor.com/t/replication-snapshot-does-not-exist-on-the-system-so-cannot-use-it/10943).

The coding in question, quoting from the forum [thread](https://forum.rockstor.com/t/replication-snapshot-does-not-exist-on-the-system-so-cannot-use-it/10943/5):

[t]here are two different place where the `snap_path` is defined. In one case it contains the slash:

https://github.com/rockstor/rockstor-core/blob/fd8f1bc53abdb3b6641f0f0351391ec134d0af2d/src/rockstor/smart_manager/replication/sender.py#L180

and re-defined in the other case (closer to where the error message is thrown that you see) where the slash is missing:

https://github.com/rockstor/rockstor-core/blob/fd8f1bc53abdb3b6641f0f0351391ec134d0af2d/src/rockstor/smart_manager/replication/sender.py#L207

In both cases it checks whether the `snap_path` is a subvolume - on the second place, if the missing slash causes it to always return an error message [...], then that might be the root cause.

At the very least, it will help to further eliminate another root cause for failing replications, if that is addressed.

phillxnet:
Fix bug introduced in last major re-work of this code. Created a failure to recover under certain circumstances that had previously been accounted for as an existing snapshot could not be identified as a result of this missing path delimiter. Resulting in reduced replication robustness.

Fixes #3147 

---

Thanks to @Hooverdan96 and forum member riceru for discovering and testing this fix. Much appreciated.

The associated bug reduced the robustness of the Replication system, where an existing snapshot, necessary for replication recovery, was not found due to a simply path string error.

### DIFF
--- a/src/rockstor/smart_manager/replication/sender.py
+++ b/src/rockstor/smart_manager/replication/sender.py
@@ -204,7 +204,7 @@ class Sender(ReplicationMixin, Process):
                     self.msg = f"{self.msg}. successful trail found for {self.rlatest_snap}".encode(
                         "utf-8"
                     )
-                    snap_path = f"{settings.MNT_PT}{self.replica.pool}.snapshots/{self.replica.share}/{self.rlatest_snap}"
+                    snap_path = f"{settings.MNT_PT}{self.replica.pool}/.snapshots/{self.replica.share}/{self.rlatest_snap}"
                     if is_subvol(snap_path):
                         self.msg = f"Snapshot({snap_path}) exists in the system and will be used as the parent".encode(
                             "utf-8"


phillxnet:
# Testing

We have already had feedback from forum member riceru on this modification and the git history indicates this was in-play prior to the last major changes in this area (Py2.7 to Py3.11 and f-strings etc).

As we have had a few minor changes in the testing branch today alone, an RPM was test built and installed on a Leap 16.0 x86_64 host and all tests were confirmed to pass.

Issue #2799, opened 2024-02-09
Unit test host sensitivity - rpmbuild - TW JeOS


phillxnet:
When running rpmbuild for the first time on a 'clean' TW 20240205 host, we have 3 remaining tests that are 'sensitive' to prior host conditions.
```
======================================================================
ERROR: test_delete_valid (rockstor.smart_manager.tests.test_task_scheduler.TaskSchedulerTests.test_delete_valid)
Test delete request with valid tdid
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib64/python3.11/shutil.py", line 825, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpmubg6ere' -> '/etc/cron.d/rockstortab'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/buildbot/rockstor-buildbot/worker/Poetry-Build-on-Tumbleweed-aarch64/rpmbuild/rockstor-core-5.0.7-2797/src/rockstor/smart_manager/tests/test_task_scheduler.py", line 235, in test_delete_valid
    response = self.client.delete("{}/{}".format(self.BASE_URL, tdid), data=data)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/buildbot/rockstor-buildbot/worker/Poetry-Build-on-Tumbleweed-aarch64/rpmbuild/rockstor-core-5.0.7-2797/.venv/lib/python3.11/site-packages/rest_framework/test.py", line 320, in delete
    response = super().delete(
               ^^^^^^^^^^^^^^^
  File "/home/buildbot/rockstor-buildbot/worker/Poetry-Build-on-Tumbleweed-aarch64/rpmbuild/rockstor-core-5.0.7-2797/.venv/lib/python3.11/site-packages/rest_framework/test.py", line 222, in delete
    return self.generic('DELETE', path, data, content_type, **extra)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3168, opened 2026-08-04
Address py3.13 syntax warnings #3136

### LINKED ISSUE #3136: Py3.13 SyntaxWarning in intended raw strings
Thanks to @FroggyFlox for a heads-up on this issue. As a follow-up to the same warnings in our test strings:

- https://github.com/rockstor/rockstor-core/pull/3123

We have a parallel issue in our regular code: via PyCharm using a debugger run (pydevd mode) of our in-code rock-tests config.
```
/opt/rockstor/.venv/bin/python -Wd -X pycache_prefix=/root/.cache/JetBrains/PyCharm2026.1/cpython-cache /root/.cache/JetBrains/RemoteDev/dist/293efcb3b6072_pycharm-2026.1.2/plugins/python-ce/helpers/pydev/pydevd.py --multiprocess --qt-support=auto --client 127.0.0.1 --port 39771 --file /root/.cache/JetBrains/RemoteDev/dist/293efcb3b6072_pycharm-2026.1.2/plugins/python-ce/helpers/pycharm/django_test_manage.py test rockstor /opt/rockstor/src/rockstor
Testing started at 5:22 PM ...
Connected to pydev debugger (build 261.24374.152)
/opt/rockstor/src/rockstor/fs/btrfs.py:590: SyntaxWarning: invalid escape sequence '\*'
  elif re.match("\t\*\*\* Some devices missing", l) is not None:
/opt/rockstor/src/rockstor/fs/btrfs.py:814: SyntaxWarning: invalid escape sequence '\.'
  if re.search("not mounted\.$", l) is not None:
/opt/rockstor/src/rockstor/fs/btrfs.py:1778: SyntaxWarning: invalid escape sequence '\W'
  fields = re.split("\W+", line)
/opt/rockstor/src/rockstor/system/osi.py:140: SyntaxWarning: invalid escape sequence '\d'
  base_dev_patterns = [("sd|vd", "\d+"), ("nvme|md|mmcblk", "p\d+")]
/opt/rockstor/src/rockstor/system/osi.py:140: SyntaxWarning: invalid escape sequence '\d'
  base_dev_patterns = [("sd|vd", "\d+"), ("nvme|md|mmcblk", "p\d+")]
/opt/rockstor/src/rockstor/system/osi.py:978: SyntaxWarning: invalid escape sequence '\d'
  end = re.search("\d+", disk).end()
/opt/rockstor/src/rockstor/storageadmin/models/validators.py:26: SyntaxWarning: invalid escape sequence '\*'
  "^(([a-zA-Z0-9\*]|[a-zA-Z0-9\*][a-zA-Z0-9\-\*]*"
/opt/rockstor/src/rockstor/storageadmin/models/validators.py:27: SyntaxWarning: invalid escape sequence '\*'
  "[a-zA-Z0-9\*])\.)*([A-Za-z0-9\*]|[A-Za-z0-9\*]"
/opt/rockstor/src/rockstor/storageadmin/models/validators.py:28: SyntaxWarning: invalid escape sequence '\-'
  "[A-Za-z0-9\-\*]*[A-Za-z0-9\*])$"
/opt/rockstor/src/rockstor/system/pkg_mgmt.py:134: SyntaxWarning: invalid escape sequence '\-'
  tags = "%{VERSION}\-%{RELEASE}\\\n%{BUILDTIME:day}"
/opt/rockstor/src/rockstor/system/samba.py:182: SyntaxWarning: invalid escape sequence '\['
  if re.match("\[global]", l) is not None:
/opt/rockstor/src/rockstor/system/samba.py:203: SyntaxWarning: invalid escape sequence '\['
  if global_section and re.match("\[", l) is not None:
/opt/rockstor/src/rockstor/system/directory_services.py:118: SyntaxWarning: invalid escape sequence '\['
  elif re.match("\[domain/%s]" % domain, line) is not None:
/opt/rockstor/src/rockstor/system/directory_services.py:173: SyntaxWarning: invalid escape sequence '\['
  elif re.match("\[sssd]", line) is not None:
/opt/rockstor/src/rockstor/system/directory_services.py:175: SyntaxWarning: invalid escape sequence '\['
  elif re.match("\[domain/{}]".format(server), line) is not None:
/opt/rockstor/src/rockstor/system/directory_services.py:215: SyntaxWarning: invalid escape sequence '\['
  elif re.match("\[sssd]", line) is not None:
/opt/rockstor/src/rockstor/system/directory_services.py:217: SyntaxWarning: invalid escape sequence '\['
  elif re.match("\[domain/{}]".format(server), line) is not None:
/opt/rockstor/src/rockstor/smart_manager/views/task_scheduler.py:133: SyntaxWarning: invalid escape sequence '\{'
  tab = "{} \{}\n".format(tab, td.crontabwindow)
/opt/rockstor/src/rockstor/system/smart.py:154: SyntaxWarning: invalid escape sequence '\s'
  if re.match(".*:\s+\(.*\)", o[j]) is not None:
/opt/rockstor/src/rockstor/system/smart.py:166: SyntaxWarning: invalid escape sequence '\s'
  elif re.match("\s", o[j]) is not None:
/opt/rockstor/src/rockstor/system/nut.py:197: SyntaxWarning: invalid escape sequence '\*'
  timer_start_pattern = "AT ONBATT \* START-TIMER early-shutdown"
/opt/rockstor/src/rockstor/system/nut.py:199: SyntaxWarning: invalid escape sequence '\*'
  timer_stop_pattern = "AT ONLINE \* CANCEL-TIMER early-shutdown"
/opt/rockstor/src/rockstor/storageadmin/views/rockon_helpers.py:539: SyntaxWarning: invalid escape sequence '\l'
  [DOCKER, "exec", c.name, "psql", "-U", "postgres", "-c", "\l"],
Found 292 test(s).
```

---
From the indicated partner pull request we have:

> A warning renamed in Py3.12:
> - Before: DeprecationWarning: invalid escape sequence
> - After: SyntaxWarning: invalid escape sequence
> 
> Move all reports to raw string definitions.

Hooverdan96:
Fixes #3136 
- Adjust expressions to raw strings
- Adjust one escaping instance ambiguity (instead of splitting string with raw strings indicator) in pgk_mgmt.py

### DIFF
--- a/src/rockstor/system/pkg_mgmt.py
+++ b/src/rockstor/system/pkg_mgmt.py
@@ -131,7 +131,7 @@ def current_version(get_build_date: bool = False) -> (str, str | None):
     # The following tags gives us a pre-formatted:
     # 'Version-Release' (first line)
     # 'Thu May 01 2025' (second line)
-    tags = "%{VERSION}\-%{RELEASE}\\\n%{BUILDTIME:day}"
+    tags = "%{VERSION}\\-%{RELEASE}\\\n%{BUILDTIME:day}"
     out, err, rc = run_command(
         [RPM, "-q", "--queryformat", tags, "rockstor"], throw=False
     )
--- a/src/rockstor/fs/btrfs.py
+++ b/src/rockstor/fs/btrfs.py
@@ -587,7 +587,7 @@ def get_pool_info(disk):
         elif re.match("\tTotal devices", l) is not None:
             fields = l.split()
             full_dev_count = int(fields[2])
-        elif re.match("\t\*\*\* Some devices missing", l) is not None:
+        elif re.match(r"\t\*\*\* Some devices missing", l) is not None:
             pool_info["hasMissingDev"] = True
     pool_info["fullDevCount"] = full_dev_count
     pool_info["missingDevCount"] = full_dev_count - attached_dev_count
@@ -811,7 +811,7 @@ def umount_root(root_pool_mnt):
         if ce.rc == 32:
             for l in ce.err:
                 l = l.strip()
-                if re.search("not mounted\.$", l) is not None:
+                if re.search(r"not mounted\.$", l) is not None:
                     return
             raise ce
     for i in range(20):
@@ -1775,7 +1775,7 @@ def pool_usage(mnt_pt):
 
     used = 0
     for line in out:
-        fields = re.split("\W+", line)
+        fields = re.split(r"\W+", line)
         if line.startswith("Data"):
             used += int(fields[5])
         elif re.search("Size", line):
--- a/src/rockstor/smart_manager/views/task_scheduler.py
+++ b/src/rockstor/smart_manager/views/task_scheduler.py
@@ -130,7 +130,7 @@ class TaskSchedulerMixin(object):
                     if td.crontabwindow is not None:
                         # add crontabwindow as 2nd arg to task script, new line
                         # moved here
-                        tab = "{} \{}\n".format(tab, td.crontabwindow)
+                        tab = fr"{tab} \{td.crontabwindow}" + "\n"
                     else:
                         logger.error("missing crontab window value")
                         continue
--- a/src/rockstor/storageadmin/models/validators.py
+++ b/src/rockstor/storageadmin/models/validators.py
@@ -23,9 +23,9 @@ from django.core.validators import validate_ipv46_address
 def validate_nfs_host_str(value):
     error_count = 0
     host_regex = (
-        "^(([a-zA-Z0-9\*]|[a-zA-Z0-9\*][a-zA-Z0-9\-\*]*"
-        "[a-zA-Z0-9\*])\.)*([A-Za-z0-9\*]|[A-Za-z0-9\*]"
-        "[A-Za-z0-9\-\*]*[A-Za-z0-9\*])$"
+        r"^(([a-zA-Z0-9\*]|[a-zA-Z0-9\*][a-zA-Z0-9\-\*]*"
+        r"[a-zA-Z0-9\*])\.)*([A-Za-z0-9\*]|[A-Za-z0-9\*]"
+        r"[A-Za-z0-9\-\*]*[A-Za-z0-9\*])$"
     )
     if re.match(host_regex, value) is None:
         error_count += 1
--- a/src/rockstor/storageadmin/views/rockon_helpers.py
+++ b/src/rockstor/storageadmin/views/rockon_helpers.py
@@ -536,7 +536,7 @@ def owncloud_install(rockon):
             cur_wait = 0
             while True:
                 o, e, rc = run_command(
-                    [DOCKER, "exec", c.name, "psql", "-U", "postgres", "-c", "\l"],
+                    [DOCKER, "exec", c.name, "psql", "-U", "postgres", "-c", r"\l"],
                     throw=False,
                 )
                 if rc == 0:
--- a/src/rockstor/system/directory_services.py
+++ b/src/rockstor/system/directory_services.py
@@ -115,7 +115,7 @@ def sssd_update_ad(domain, config):
                     # empty line or new section without empty line before it.
                     tfo.write(ol)
                     domain_section = False
-            elif re.match("\[domain/%s]" % domain, line) is not None:
+            elif re.match(r"\[domain/%s]" % domain, line) is not None:
                 domain_section = True
             tfo.write(line)
         if domain_section is True:
@@ -170,9 +170,9 @@ def sssd_add_ldap(ldap_params):
                 for k, v in opts.items():
                     if re.match(k, line) is None:
                         tfo.write("{} = {}\n".format(k, v))
-            elif re.match("\[sssd]", line) is not None:
+            elif re.match(r"\[sssd]", line) is not None:
                 sssd_section = True
-            elif re.match("\[domain/{}]".format(server), line) is not None:
+            elif re.match(r"\[domain/{}]".format(server), line) is not None:
                 domain_section = True
             tfo.write(line)
         if domain_section is False:
@@ -212,9 +212,9 @@ def sssd_remove_ldap(server):
                     continue
             elif domain_section is True:
                 continue
-            elif re.match("\[sssd]", line) is not None:
+            elif re.match(r"\[sssd]", line) is not None:
                 sssd_section = True
-            elif re.match("\[domain/{}]".format(server), line) is not None:
+            elif re.match(r"\[domain/{}]".format(server), line) is not None:
                 domain_section = True
                 continue
             elif len(line.strip()) == 0:
--- a/src/rockstor/system/nut.py
+++ b/src/rockstor/system/nut.py
@@ -194,9 +194,9 @@ def update_upssched_early_shutdown(seconds):
     """
     distro_id = distro.id()
     # setup out match patterns (with escaped '*') and associated lines
-    timer_start_pattern = "AT ONBATT \* START-TIMER early-shutdown"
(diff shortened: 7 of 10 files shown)


Hooverdan96:
Testing using test suite:

```
 ~# cd /opt/rockstor/src/rockstor
 ~# export DJANGO_SETTINGS_MODULE=settings
 ~# poetry run django-admin test -v 2
...
Ran 292 tests in 85.657s

OK
```
Single interactive test - creating a nfs share and put illegal characters into the client string to see whether the changed regex in the validator.py will catch it:

image

Successfully error is thrown, when alphanumeric client name is entered instead, the nfs export is saved.

src/rockstor/system/pkg_mgmt.py
```diff
@@ -131,7 +131,7 @@ def current_version(get_build_date: bool = False) -> (str, str | None):
     # The following tags gives us a pre-formatted:
     # 'Version-Release' (first line)
     # 'Thu May 01 2025' (second line)
-    tags = "%{VERSION}\-%{RELEASE}\\\n%{BUILDTIME:day}"
+    tags = "%{VERSION}\\-%{RELEASE}\\\n%{BUILDTIME:day}"
```
phillxnet, line 134: @Hooverdan96 Just checking this one:

```
rslowroll:~ # rpm -q --queryformat "%{VERSION}\-%{RELEASE}\\\n%{BUILDTIME:day}" rockstor
5.5.5-3168\nWed Aug 05 2026rslowroll:~ # 
rslowroll:~ # rpm -q --queryformat "%{VERSION}\\-%{RELEASE}\\\n%{BUILDTIME:day}" rockstor
5.5.5-3168\nWed Aug 05 2026rslowroll:~ #

rslowroll:~ # rpm -q --queryformat %{VERSION}\-%{RELEASE}\\\n%{BUILDTIME:day} rockstor
5.5.5-3168
Wed Aug 05 2026rslowroll:~ # rpm -q --queryformat %{VERSION}\\-%{RELEASE}\\\n%{BUILDTIME:day} rockstor
5.5.5-3168
Wed Aug 05 2026rslowroll:~ #
```

Which looks to be good.

phillxnet (changes requested):
@Hooverdan96 Thanks for seeing to these updates. It's a shame we keep having pending Python breakages when we are this far along in Python3. But hopefully these types of failures/pending-failures will stop cropping up so often.

As-is this all looks mostly good, and we have now gone over the pkg related change but the by-hand testing, in comments, looks to prove this to be a good replacement.

Unfortunately my prior reproducer of this, debug run of our tests in PyCharm, now looks to be quite intermittent re reproducing these prior warnings!

I think we have a little work remaining here on the schedule task string use.

If I create a scheduled Share snapshot task of say every 5 mins, limited to 3 "Maximum count" I get no snapshots created!!

And we have the following raw interpretation of our newline "\n" use repeated in the resulting crontab:

```
rslowroll:~ # cat /etc/cron.d/rockstortab 
SHELL=/bin/bash
PATH=/sbin:/bin:/usr/sbin:/usr/bin
MAILTO=root
# These entries are auto generated by Rockstor. Do not edit.
*/5 * * * * root /opt/rockstor/.venv/bin/st-snapshot 24 \*-*-*-*-*-*\n
```

We can likely just add two strings, the new raw r"" variant, and another using the "\n" to achieve what we had previously and appease the new Python/incoming interpreter at the same time.

I.e. we have from the following: https://docs.python.org/3/reference/lexical_analysis.html#string-prefixes

that we can combine both f-stings and raw. So in the case of the currently breaking change to tasks in these proposed changes resulting in a raw "\n" that voids the intended cron addition (above crontab entry), we could do instead:

```
                        tab = fr"{tab} \{td.crontabwindow}" + "\n"
```
Which also moves our older string.format use to the nicer readability wise f-sting; and also maintains our prior use of embedding a newline required by crontab: resulting in the the following crontab entry:

```
rslowroll:~ # cat /etc/cron.d/rockstortab
SHELL=/bin/bash
PATH=/sbin:/bin:/usr/sbin:/usr/bin
MAILTO=root
# These entries are auto generated by Rockstor. Do not edit.
*/5 * * * * root /opt/rockstor/.venv/bin/st-snapshot 25 \*-*-*-*-*-*
rslowroll:~ # 
```
And by my testing here a now working again scheduled Share snapshot:


ID | Status | Start Time | End Time
-- | -- | -- | --
1 | finished | August 5th 2026, 6:05:00 pm | August 5th 2026, 6:05:03 pm
2 | finished | August 5th 2026, 6:10:00 pm | August 5th 2026, 6:10:03 pm
3 | finished | August 5th 2026, 6:15:00 pm | August 5th 2026, 6:15:03 pm

Hooverdan96:
Ah ok. I did not really grasp the combination of prefixes to get the best of both worlds (honor curly brackets/variables, while also treating backslashes as literals. I will update that.

Hooverdan96:
Doesn't that also mean that the task_scheduler test does not accurately test a "valid" crontab schedule?

https://github.com/rockstor/rockstor-core/blob/testing/src/rockstor/smart_manager/tests/test_task_scheduler.py

Or does it just need another variant in there that checks the "false" syntax (like in your example above with the line feed)?

phillxnet:
@Hooverdan96 Re:

> Doesn't that also mean that the task_scheduler test does not accurately test a "valid" crontab schedule?

Yes, in a way. We likely need to extend the tests to look for what is actually written to disk. And for that we need employ a 'fake' test filesystem setup. From memory @FroggyFlox has done some tests that involve something like that I believe. And I think these were the first tests of this type in Rockstor. We have a lot of API type tests - i.e. does the subsystem respond appropriately type thing. I.e. rejecting errors and finishing apparently successfully if all was viable. But we also do disk-writes of config for example. Which is the case here in the tasks string - cronttab needs newlines or else fails silently as there is no line completely defined. So we can always do with more tests that prove what is written, by way of config, is actually what we intended to write. Hence the use of a faked filesystem that we can then check a file was written, and that its contents are correct for the given test data. We also have remaining some 'real' filesystem assumptions within our tests that aught not exist, detailed in the following issue:

- https://github.com/rockstor/rockstor-core/issues/2799

Curiously fromthat issue we have:
- "All faulty tests are from: rockstor.smart_manager.tests.test_task_scheduler"
-  "I was actually going to look at your more recent (and our first) proper file mocking in testing to fix this as it goes. Given we are now on new-enough Python. I think it was a more recent addition, from memory." from: https://github.com/rockstor/rockstor-core/issues/2799#issuecomment-1936176223

That last comment links to:
- Fix mocking insufficiencies in system.network.py #2717 #2718

phillxnet:
@Hooverdan96 Thanks for seeing to that last fix (hopefully). I've rebuild and installed another rpm built from this branch and the scheduling now works.

Could you squash and re-present this so we can then pop it in as it clears our way, at least by way of this deprecation to Syntax Python warning to error  situation.

# Test schedule result

Test system was a Slowroll instance on X86_64.

  | Name | Created on | Visibility to user | Writable | Total Size | Exclusive Size
-- | -- | -- | -- | -- | -- | --
  | sched-snap_202608061000 | August 6th 2026, 11:00:04 am | Visible | Yes | 16.00 KB | 16.00 KB
  | sched-snap_202608060955 | August 6th 2026, 10:55:03 am | Visible | Yes | 16.00 KB | 16.00 KB
  | sched-snap_202608060950 | August 6th 2026, 10:50:03 am | Visible | Yes | 16.00 KB | 16.00 KB

phillxnet (approved):
@Hooverdan96 Thanks again for seeing to these changes required for us to later upgrade our Python. Much appreciated.

Post your git squash, cheers, I've re-build a Slowroll rpm and installed it on the build host which was then able to setup and carry-out scheduled snapshots - our last change here.

I'll get this merged directly.

PR #3142, opened 2026-06-26
Django 6.0 point release + routine updates #3141

phillxnet:
Routine updates via `poetry update` to refresh poetry.lock, within existing pyproject.toml pinning bar "django" to latest.

```
poetry update
Updating dependencies
Resolving dependencies... (16.4s)

Package operations: 0 installs, 8 updates, 0 removals

  - Updating cryptography (48.0.0 -> 49.0.0)
  - Updating more-itertools (11.0.2 -> 11.1.0)
  - Updating idna (3.16 -> 3.18)
  - Updating certifi (2026.5.20 -> 2026.6.17)
  - Updating greenlet (3.5.1 -> 3.5.2)
  - Updating django (6.0.5 -> 6.0.6)
  - Updating jwcrypto (1.5.7 -> 1.5.8)
  - Updating zope-interface (8.4 -> 8.5)

Writing lock file
```

Fixes #3141

PR #3138, opened 2026-05-22
Upgrade to Django 6.0 #3093

phillxnet:
Pyproject.toml modified for Django and docutils; resulting in `poetry update`:

```
Package operations: 0 installs, 12 updates, 0 removals

  - Updating cryptography (47.0.0 -> 48.0.0)
  - Updating jaraco-functools (4.4.0 -> 4.5.0)
  - Updating idna (3.13 -> 3.16)
  - Updating certifi (2026.4.22 -> 2026.5.20)
  - Updating urllib3 (2.6.3 -> 2.7.0)
  - Updating django (5.2.13 -> 6.0.5)
  - Updating pytz (2026.1.post1 -> 2026.2)
  - Updating greenlet (3.4.0 -> 3.5.1)
  - Updating requests (2.33.1 -> 2.34.2)
  - Updating zope-event (6.1 -> 6.2)
  - Updating docutils (0.21.2 -> 0.22.4)
  - Updating psycopg (3.3.3 -> 3.3.4)

Writing lock file
```

Fixes #3093

PR #3177, opened 2026-08-17
Config restore JSON Decode Error #3175

### LINKED ISSUE #3175: Config restore JSON Decode Error
Using a current development build based on latest testing was unable to restore a config save file with JSON decode related logging as follows:
> from : /opt/rockstor/var/log/rockstor.log
```
[16/Aug/2026 16:42:21] DEBUG [storageadmin.views.config_backup:729] restore starting...
[16/Aug/2026 16:42:21] DEBUG [storageadmin.views.config_backup:731] restore submitted...
[16/Aug/2026 16:42:27] INFO [storageadmin.tasks:56] -STARTING Huey task: [restore_config], id: b7e72d9c-8cc6-48e0-a244-6b56fa7d1e15.
[16/Aug/2026 16:42:27] ERROR [storageadmin.tasks:108] Task [restore_config], id: b7e72d9c-8cc6-48e0-a244-6b56fa7d1e15 failed, Args: (9,), Kwargs: {}, ExceptionExpecting value: line 2 column 1 (char 1)
```
and via the related event detailed in Django admin site via the newely updated Huey we have more for the above front-trimmed message:

Time | Queue | Task | Signal | Duration | Error
-- | -- | -- | -- | -- | --
2026-08-16 16:42:27 | huey | storageadmin.views.config_backup.restore_config | error | 100 ms | JSONDecodeError('Expecting value: line 2 column 1 (char 1)')
2026-08-16 16:42:27 | huey | storageadmin.views.config_backup.restore_config | executing | - | -
2026-08-16 16:42:21 | huey | storageadmin.views.config_backup.restore_config | enqueued | - | -

---

The current reproducer system is not representative of released rpms, but noting here in case this is reproducible.

phillxnet:
Improve robustness re config backup file parsing, adding per line checks for the expected json array per line content. Move to using gzip.open() as a context manager and a more modern for-line-in-file_object approach.
Add code comments re config backup file format.
Variable name change refactoring for readability.

Fixes #3175

### DIFF
--- a/src/rockstor/storageadmin/views/config_backup.py
+++ b/src/rockstor/storageadmin/views/config_backup.py
@@ -14,6 +14,7 @@ General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
+
 import copy
 import gzip
 import json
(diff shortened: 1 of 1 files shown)


phillxnet:
This improvement, although generally nice-to-have, was primarily motivated by the referenced issue detailed behaviourly change in a Django admin tool `dumpdata` that we use to dump our two databases (one per line), serialized to JSON, to create our config backup file.

- Issue wise Django uses the Trac ticketing system:  https://code.djangoproject.com/ticket/36177
- Upstream Django change: https://github.com/django/django/pull/19232

From the above, this change was indicated in "miscellaneous" in the Django 6.0 release notes.
```
* The :ref:`JSON <serialization-formats-json>` serializer now writes a newline
+ at the end of the output, even without the ``indent`` option set.
```

Published here: https://docs.djangoproject.com/en/6.0/releases/6.0/#miscellaneous

phillxnet:
Given the context provided in https://github.com/rockstor/rockstor-core/pull/3177#issuecomment-5326261182 this inherited format change in our config backup file saves was brought about as from our adoption of Django 6.0+:

- https://github.com/rockstor/rockstor-core/pull/3138 (5.2.13 -> 6.0.5)
- https://github.com/rockstor/rockstor-core/pull/3142 (6.0.5 -> 6.0.6)

Both of which were initially part of **testing/edge channel RPM** releases as from **version 5.5.4-0** on Jun 26th 2026, and so will be part of the next Stable channel release of 5.6.*-* to be released in time.

- https://github.com/rockstor/rockstor-core/releases/tag/5.5.4-0
- Forum announcement: https://forum.rockstor.com/t/v5-5-testing-channel-changelog/10895/8

By accommodating these changes, rather than removing our now redundant LF ("\n") between our DB dumps, we honour config backup files taken as form 5.5.4-0 rpm (we are currently on 5.5.5-0 testing) and, as this PR text indicates,  we gain:

> Improve robustness re config backup file parsing

Also, if we were to remove our now redundant LF on config backup saves, added originally to split our JSON formatted DB dumps into their own lines, and upstream were to revert their change, we would encounter yet another breakage. Whilst also not addressing the observed incompatibility to restore config files created since 5.5.4-0 in our testing channel field testers. Ergo I propose we do not now also remove the redundant but now harmless as of the proposed changes here, clear line between our DB dumps in all config backup files saved as from v5.5.4-0 rpm releases.

phillxnet:
# Testing

Both a pre and post V5.5.4-0 formatted config save file were successfully restored to what is intended to be released in testing/edge repos as v5.5.6-0.

N.B. most editors will add a LF on save by default. Making our prior format a little strange given it did not end with "\n" character.

I.e. prior format reduced to just the json arrays with an empty object in each: not valid for different reasons but for demonstration purposes here:
## Old format
```
printf "[{}]\n[{}]" > old-format.json
cat -e old-format.json 
[{}]$
[{}]
```
## New format
```
printf "[{}]\n\n[{}]\n" > new-format.json
cat -e new-format.json 
[{}]$
$
[{}]$
```

An RPM package was successfully built and installed on a Leap 16.0 using this branch, and all existing tests passed.

Ideally we now introduce tests using our newly adopted pyfakefs to help against this same inadvertent failure from upstream changes again. I.e. save minimum config instantiated by fixtures, then confirm the file is of the expected format and content. We already have quite a few tests to cover the restore of the models themselves. But non that confirm the saved file format. However, due to our simple DB dump method, there are few layers actually involved and we were unlucky here in that one of them modified it's behaviour and consequently our format. But we do now have more robustness in this area.

Issue #3174, opened 2026-08-12
unable to destroy quota group: Device or resource busy


phillxnet:
Observed initially in Tumbleweed/Slowroll under the previously issued (referenced below), more complex scenarios, and now easily reproduced in our "Built on openSUSE Leap 16.0" instances via:

1. Create a Share
2. Delete said Share
3. Error show as below

**N.B. The Share is however deleted.**

N.B. also pertains to the deletion of a Share's snapshot.

I.e. Closing the Error dialog (text copied in below), and the  initial "Delete share-name ?" confirmation dialog that is still shown; as we have an apparent failure (all-be-it partial), and after a Web-UI refresh the Share has in fact been deleted.

In the below referenced issues, indicating the same failure but embedded in more complex scenarios, there are likely more failure artefacts; such as a failed snapshot-to-share promotion and consequent stuck replication.

## Dialog details

> Houston, we've had a problem.
> Failed to delete the share (test-share-2). Error from the OS: Error running a command. cmd = /usr/sbin/btrfs qgroup destroy 0/257 /mnt2/rock-pool. rc = 1. stdout = ['']. stderr = ['ERROR: unable to destroy quota group: Device or resource busy', ''] 

With the following details:

> Traceback (most recent call last):
>   File "/opt/rockstor/src/rockstor/storageadmin/views/share.py", line 391, in delete
>     remove_share(share.pool, share.subvol_name, share.pqgroup, force=force)
>     ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/opt/rockstor/src/rockstor/fs/btrfs.py", line 1196, in remove_share
>     qgroup_destroy(qgroup, root_pool_mnt)
>     ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
>   File "/opt/rockstor/src/rockstor/fs/btrfs.py", line 1503, in qgroup_destroy
>     return run_command([BTRFS, "qgroup", "destroy", qid, mnt_pt], log=True)
>   File "/opt/rockstor/src/rockstor/system/osi.py", line 266, in run_command
>     raise CommandException(cmd, out, err, rc)
[...]

PR #2892, opened 2024-08-21
Update Copyright notices #2883

phillxnet:
See the following GitHub issue for details:
- https://github.com/rockstor/rockstor-core/issues/2883

## Also Includes
- Normalise on 'Rockstor' capitalisation in non-code use: replacing 'RockStor'.
- Formatting changes/normalisation (per file type) in Copyright associated headers.

Fixes #2883

PR #3178, opened 2026-08-18
unable to destroy quota group: Device or resource busy #3174

phillxnet:
Adapt to changed behaviour re qgroups in more recent btrfs/kernel versions. Leap 16.0 onwards targeted. Add qgroup_remove(), step as a pre-cursor to existing qgroup_destroy() adapting to more recent btrfs quotas behaviour.
If a remaining share or snapshot qgroup is a child of a parent qgroup to be destroyed, we get "Device or resource busy", log but do not raise an unmanaged exception in this case. I.e. move qgroup_destroy() to be opportunistic.

- Refactor low-level remove_share() to remove_share_subvol() and minimise/rationalise arguments.
- Rock-ons service block for new 'Delete Pool management' behaviour if the associated Pool has Shares.
- Improve 'Delete Pool management' pop-up notices: link to config-backup Web-UI page and online docs, and explain we are DB deleting with consequences of reduced Web-UI to OS consistency.
- Remove NFS exports and groups in Pool management delete.
- Add some btrfs.py type hinting, some minor explicit "return None" clarifications and string.format use to f-strings.
- Minor test_pools.py updates.
- Create_repclone() minor type hinting.

## Improve UI consistency re Pool 'delete'

Previously on Pool delete we deleted all subvols, but did not delete the pool signatures. Full Pool delete we already offer at the device level via the eraser icon and the: "Wipe existing Filesystem" - with more severe warnings and a multi-step process.

Changing to Pool delete management as "unmount and DB delete" maintains our existing import option there after - only now (post PR) this option will allow import of the entire pool as-was, not just a sub-vols removed version, which is the pre PR behaviour. We also avoid potentially problematic delays inured when decommissioning a Pool via the prior method of subvol deletion. Instead folks would, post PR, delete the Pool's management, then wipe all that pools prior disk members: one at a time.

This new behaviour also enables system pool "Delete management".

We no longer auto-manage the system pool; but this was once default, and may have been inherited during updates. Ergo the removal of a block on system pool "Delete management" as this management can be a draw on system resources, enabling updated systems inheriting the old default to reconfigure to the new default of no system pool management.


phillxnet:
Getting there on this front. When doing a Pool management delete on this branch we now get:

```
[21/Aug/2026 18:26:15] INFO [storageadmin.views.pool:776] Deleting scheduled scrub task (Scrub-Task) for (rock-pool).
[21/Aug/2026 18:26:15] INFO [storageadmin.views.pool:802] Proceeding with unmount and database (DB) removal of:
[21/Aug/2026 18:26:15] INFO [storageadmin.views.pool:803] - Pool (rock-pool) mount point /mnt2/rock-pool.
[21/Aug/2026 18:26:15] INFO [storageadmin.views.pool:827] - Deleting scheduled snapshot task (snap-test) for (share1).
[21/Aug/2026 18:26:16] INFO [storageadmin.views.pool:833] -- Unmounting subvol (share1) mount point /mnt2/share1.
[21/Aug/2026 18:26:16] INFO [storageadmin.views.pool:827] - Deleting scheduled snapshot task (snap-test2) for (share2).
[21/Aug/2026 18:26:16] INFO [storageadmin.views.pool:833] -- Unmounting subvol (share2) mount point /mnt2/share2.
[21/Aug/2026 18:26:16] INFO [storageadmin.views.pool:833] -- Unmounting subvol (share3) mount point /mnt2/share3.
[21/Aug/2026 18:26:16] INFO [storageadmin.views.pool:837] - Unmounting Pool (rock-pool) mount point /mnt2/rock-pool.
[21/Aug/2026 18:26:16] INFO [storageadmin.views.pool:839] Removing Pool (rock-pool) management and associated configuration.
```

phillxnet:
Just got to work out why an as-yet uncommitted build-in post_delete signal from TaskDefinition.delete() is not being received, as that was the chosen path here for returning our crontab to consistency when associated TaskDefinition instance deletes occure. So nearly there. I'll likely package this under the Pool management delete now as that is the majority of the changes/work here - but it relates closely to the original quotas issue as we will no longer try to decommission, subvol by subvol, an entire Pool when we have had the "Pool management delete" idea waiting in the stage wings for some time now. Some doc changes will be required there-after to clarify the change in behaviour but having exercised this new behaviour quite a bit now it feels much more flexible than what we have had of late.

phillxnet:
I've ended up heeding all the Django doc warnings regarding signals, even the builtin ones, and going with a model.delete() override here to bring our custom crontab file back in line with the Web-UI (DB) when doing a model delete of task definitions during Pool management delete. Taking care to only write to our crontab when a change has been made. I've also included the appropriate warnings regarding this override being skipped on mass deletions.

This way we stick to a light-weight approach during Pool management delete and avoid consequent API calls back-in that would rather complicate things while we were mid Pool management delete DB transaction.atomic.

phillxnet:
Outstanding issues:
- ~~NFS config tidy on Pool delete~~ ~~not actually functional currently~~ - planning to schedule this operation, and consequent service restart to be out from under our atomic transaction.
- Add last sub-system (SFTP) config clean-up on Pool management delete.

Hooverdan96:
@phillxnet, it will probably be a pain in the neck, but since you're touching such a large amount of files, do you also want to include the copyright headers with the current year? I know, I keep forgetting that all the time.

phillxnet:
@Hooverdan96 Agreed, I have updated some Copyright headers years, but yes I should at least see to updating all the files I end up changing here in that regard.

As this PR will be difficult to review as-is, I suggest we spin-off an all-project sed based copyight year update now we have normalise ish our copyright headers under:

- https://github.com/rockstor/rockstor-core/pull/2892

